Issue #9, opened 2015-08-25
Support configuration options


justin808:
Location of server webpack generated file, and default for prerender.

justin808:
LGTM -- is this ready for merge?

PR #15, opened 2015-08-31
Add React on Rails configuration

samnang:
https://github.com/shakacode/react_on_rails/issues/9

Pair with Mario(@mapreal19)

### DIFF
--- a/spec/dummy/config/initializers/react_on_rails.rb
+++ b/spec/dummy/config/initializers/react_on_rails.rb
@@ -1,0 +1,3 @@
+ReactOnRails.configure do |config|
+  config.bundle_js_file = "app/assets/javascripts/generated/server.js"
+end
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -26,8 +26,8 @@ module ReactOnRailsHelper
     # server has already rendered the HTML.
     @react_component_index ||= 0
 
-    prerender = options[:prerender].nil? ? true : options[:prerender]
-    trace = options[:trace].nil? ? false : options[:trace]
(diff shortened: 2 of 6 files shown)


spec/dummy/config/initializers/react_on_rails.rb
```diff
@@ -0,0 +1,3 @@
+ReactOnRails.configure do |config|
+  config.bundle_js_file = "app/assets/javascripts/generated/server.js"
```
alex35mil, line 2: What if there will be more than 1 js app? I think it should be defined on `react_component` level.

justin808, line 2: @alexfedoseev this is a good point. I think we can do the following:
1. `bundle_js_file` is the default
2. the bundle file can be overridden like the prerender option
3. we offer a setup for multiple bundles after v.1.0 which would have the ability to setup a common file, and then the calls to react_component would specify the option of the `bundle_js_file`, and this would be combined with the common

### HOWEVER

If we're caching the server.js file on the Rails server between requests, why not just have **ONE** server.js file that can render any required component.

I'm not seeing what could be the advantage of separate bundles for the server side.

I can see an advantage for the client side in terms of not sending extra JS. What's the advantage on the server side?

CC: @samnang

samnang, line 2: @justin808 good point, I totally agree if we want to cache on server, then there is no any benefit for splitting the file.

samnang:
We could merge this one for now because this pull request only for introducing configurations. We could deal with multiple bundle files if necessary in separate pull request.

justin808:
LGTM!

PR #19, opened 2015-09-10
Allow rendering of React components without using a generator function (just direct React Component)

josiasds:
When using Redux, the main component exposed to the API is a function that returns a React component. This PR allows rendering of simple React components. The more advanced functionality of using a generator function is essential for tools like redux and react-router.

In the dummy app, along with App component (that uses Redux), `ServerApp` and `ClientApp` now also export a simple React component, HelloWorld. The component can be rendered by the client or using server side rendering, as demonstrated in the index page.

To test, run the dummy app and export some components with and without Redux, in both client an server apps.

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -28,18 +28,18 @@ module ReactOnRailsHelper
     prerender = options.fetch(:prerender) { ReactOnRails.configuration.prerender }
     trace = options.fetch(:trace, false)
 
-    dataVariable = "__#{component_name.camelize(:lower)}Data#{@react_component_index}__"
-    reactComponent = component_name.camelize
-    domId = "#{component_name}-react-component-#{@react_component_index}"
+    data_variable = "__#{component_name.camelize(:lower)}Data#{@react_component_index}__"
+    react_component_name = component_name.camelize
+    dom_id = "#{component_name}-react-component-#{@react_component_index}"
     @react_component_index += 1
 
     turbolinks_loaded = Object.const_defined?(:Turbolinks)
-    install_render_events = turbolinks_loaded ? turbolinks_bootstrap(domId) : non_turbolinks_bootstrap
+    install_render_events = turbolinks_loaded ? turbolinks_bootstrap(dom_id) : non_turbolinks_bootstrap
 
     page_loaded_js = <<-JS
       (function() {
-        window.#{dataVariable} = #{props.to_json};
-        #{define_render_if_dom_node_present(reactComponent, dataVariable, domId, trace)}
+        window.#{data_variable} = #{props.to_json};
+        #{define_render_if_dom_node_present(react_component_name, data_variable, dom_id, trace)}
         #{install_render_events}
       })();
     JS
@@ -49,7 +49,7 @@ module ReactOnRailsHelper
     # Create the HTML rendering part
     if prerender
       render_js_expression = <<-JS
-        this.React.renderToString(#{render_js_react_element(reactComponent, props.to_json)});
+        this.React.renderToString(#{render_js_react_element(react_component_name, props.to_json)});
       JS
       server_rendered_react_component_html = render_js(render_js_expression)
     else
@@ -58,7 +58,7 @@ module ReactOnRailsHelper
 
     rendered_output = content_tag(:div,
                                   server_rendered_react_component_html,
-                                  id: domId)
+                                  id: dom_id)
 
     <<-HTML.strip_heredoc.html_safe
       #{data_from_server_script_tag}
@@ -76,27 +76,28 @@ module ReactOnRailsHelper
 
   private
 
-  def debug_js(react_component, data_variable, dom_id, trace)
+  def debug_js(react_component_name, data_variable, dom_id, trace)
     if trace
       <<-JS.strip_heredoc
-        console.log("CLIENT SIDE RENDERED #{react_component} with dataVariable #{data_variable} to dom node with id: #{dom_id}");
+        console.log("CLIENT SIDE RENDERED #{react_component_name} with data_variable #{data_variable} to dom node with id: #{dom_id}");
       JS
     else
       ""
     end
   end
 
-  def render_js_react_element(react_component, props)
-    ReactOnRails::ReactRenderer.new.render_js_react_element(react_component, props).html_safe
+  def render_js_react_element(react_component_name, props_name)
+    ReactOnRails::ReactRenderer.render_js_react_element(react_component_name, props_name)
   end
 
-  def define_render_if_dom_node_present(react_component, data_variable, dom_id, trace)
+  def define_render_if_dom_node_present(react_component_name, data_variable, dom_id, trace)
     <<-JS.strip_heredoc
       var renderIfDomNodePresent = function() {
         var domNode = document.getElementById('#{dom_id}');
         if (domNode) {
-          #{debug_js(react_component, data_variable, dom_id, trace)}
-          React.render(#{render_js_react_element(react_component, data_variable)}, domNode);
+          #{debug_js(react_component_name, data_variable, dom_id, trace)}
+          var reactElement = #{render_js_react_element(react_component_name, data_variable)};
+          React.render(reactElement, domNode);
         }
       }
     JS
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -49,7 +49,7 @@ module ReactOnRailsHelper
     # Create the HTML rendering part
     if prerender
       render_js_expression = <<-JS
-          renderReactComponent(this.#{reactComponent}, #{props.to_json})
+        renderReactComponent(this.#{reactComponent}, #{props.to_json})
       JS
       server_rendered_react_component_html = render_js(render_js_expression)
     else
@@ -90,10 +90,18 @@ module ReactOnRailsHelper
     <<-JS.strip_heredoc
       var renderIfDomNodePresent = function() {
         var domNode = document.getElementById('#{dom_id}');
+
         if (domNode) {
           #{debug_js(react_component, data_variable, dom_id, trace)}
-          var reactComponent = #{react_component}(#{data_variable});
-          React.render(reactComponent, domNode);
+          var element;
+
+          if (Object.getPrototypeOf(#{react_component}) === React.Component) {
+            element = React.createElement(#{react_component}, #{data_variable});
+          } else {
+            // when using Redux, we need to pull the component off the wrapper function
+            element = #{react_component}(#{data_variable});
+          }
+          React.render(element, domNode);
         }
       }
     JS
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -49,7 +49,7 @@ module ReactOnRailsHelper
     # Create the HTML rendering part
     if prerender
       render_js_expression = <<-JS
-        renderReactComponent(this.#{reactComponent}, #{props.to_json})
+        this.React.renderToString(#{render_js_react_element(reactComponent, props.to_json)});
       JS
       server_rendered_react_component_html = render_js(render_js_expression)
     else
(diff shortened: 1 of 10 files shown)


app/helpers/react_on_rails_helper.rb
```diff
@@ -90,10 +90,18 @@ def define_render_if_dom_node_present(react_component, data_variable, dom_id, tr
...
       var renderIfDomNodePresent = function() {
         var domNode = document.getElementById('#{dom_id}');
+
         if (domNode) {
           #{debug_js(react_component, data_variable, dom_id, trace)}
-          var reactComponent = #{react_component}(#{data_variable});
-          React.render(reactComponent, domNode);
+          var element;
+
+          if (Object.getPrototypeOf(#{react_component}) === React.Component) {
```
justin808: :+1:  Looks excellent. CC: @alexfedoseev @samnang. This looks like a good way to determine if the component is a React.Component. We need to make sure this works for createClass as well as the ES6 class definition. Another option could be to check if the object has a property called `render` that's a function. We also need to consider if the React API might change in ways that would break this.

app/helpers/react_on_rails_helper.rb
```diff
@@ -90,10 +90,18 @@ def define_render_if_dom_node_present(react_component, data_variable, dom_id, tr
     <<-JS.strip_heredoc
```
justin808, line 94: Let's move this method over to lib/react_on_rails/react_renderer.rb.

Then we can reuse the common lines from 96 to 103.

The below code has the var names hard coded as `reactComponent` and `props`. The upper code uses the rails variables.

``` javascript
        var element;

        if (Object.getPrototypeOf(reactComponent) === this.React.Component) {
          element = this.React.createElement(reactComponent, props);
        } else {
          // when using Redux, we need to pull the component off the wrapper function
          element = reactComponent(props);
        }
```

spec/dummy/client/app/startup/serverGlobals.jsx
```diff
@@ -3,14 +3,17 @@
...
 
+// Example of server rendering without using Redux
+import { HelloWorld } from './ServerApp';
+
 // Example of server rendering with no React
 import HelloString from '../non_react/HelloString';
 
 // We can use the node global object for exposing.
 // NodeJs: https://nodejs.org/api/globals.html#globals_global
-// Uncomment next 4 lines to use global
```
justin808, line 11: good job cleaning up out of date docs!

lib/react_on_rails/react_renderer.rb
```diff
@@ -2,13 +2,19 @@ module ReactOnRails
...
-    # calling the calling context which is the "this" in the execJs environment.
+    # calling context which is the "this" in the execJs environment.
     def render_js_react_component
       <<-JS.strip_heredoc
-      function renderReactComponent(componentClass, props) {
-        return this.React.renderToString(
-         componentClass(props)
-        );
+      function renderReactComponent(reactComponent, props) {
+        var element;
```
justin808: see note above on DRY'ing this up.

spec/dummy/app/views/pages/index.html.erb
```diff
@@ -39,16 +22,50 @@
...
+  <% puts "server rendering react components" %>
   <% puts "=" * 80 %>
   <!-- Default for prerender is true for the app, set in config/react_on_rails.rb -->
   <%= react_component("App", @app_props_server_render, trace: true) %>
+  <hr/>
+
+  <h1>Server Rendered/Cached React Component Without Redux</h1>
+  <code>
+    <%% cache @app_props_server_render do %><br/>
+    <%%    = react_component("HelloWorld", @app_props_server_render, trace: true) %><br/>
```
justin808, line 34: Maybe it's confusing that we call it HelloWorld in one case, and HelloWorldComponent in another case. Why is this done? Maybe I'm missing something. Be sure to note that ClientApp.jsx is ONLY loaded client side in the browser, and ServerApp.jsx is ONLY loaded server side by the rails server, so there would be no naming conflict.

app/helpers/react_on_rails_helper.rb
samnang: Yes, it's a good point. We need to make sure it work with `React.createClass` as well.

samnang: I could confirm that it doesn't work with `createClass`. http://jsbin.com/relijopuqi/edit?js,console,output

spec/dummy/app/views/pages/index.html.erb
josiasds, line 34: @justin808 [HelloWorldComponent](https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/client/app/startup/ClientHelloWorldComponent.jsx) is a function that returns the component `<HelloWorld />`.
Here, the HelloWorld component is directly called without the wrapper function, and loaded server side (after exported by `ServerApp`), for demonstration purposes.
Below, the same HelloWorld component is loaded by the client.

justin808, line 34: We should make the component generator "function" name more explicit. Like `HelloWorldComponentGenerator` or `HelloWorldApp`. @alexfedoseev What do you think? I think we might be using the suffix `App` to mean these generators.

app/helpers/react_on_rails_helper.rb
justin808: @petehunt What's the best, most canonical, way to decide if we have a real React Component? (class or generated)? Something that will work as React evolves...

spec/dummy/app/views/pages/index.html.erb
alex35mil, line 34: :+1: for `App` suffix

app/helpers/react_on_rails_helper.rb
petehunt: Probably just stick with: https://github.com/facebook/react/blob/master/src/isomorphic/classic/element/ReactElementValidator.js#L259

justin808:
@josiasds Let's get this wrapped up. I want to push this to rubygems ASAP! Lots of people ready to use this! :+1: :-)

josiasds:
@justin808 Ready!

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,17 @@
...
-    # calling the calling context which is the "this" in the execJs environment.
-    def render_js_react_component
-      <<-JS.strip_heredoc
-      function renderReactComponent(componentClass, props) {
-        return this.React.renderToString(
-         componentClass(props)
-        );
-      }
+    # Returns a React element, unwrapping it if the component is a generator function
+    def render_js_react_element(react_component, props)
```
justin808: static?
`def self.render_js...`

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,17 @@
...
-    # "this" does not need a closure as it refers to the "this" defined by the
-    # calling the calling context which is the "this" in the execJs environment.
-    def render_js_react_component
-      <<-JS.strip_heredoc
-      function renderReactComponent(componentClass, props) {
-        return this.React.renderToString(
-         componentClass(props)
-        );
-      }
+    # Returns a React element, unwrapping it if the component is a generator function
```
justin808: Definitely worth expanding this comment.

app/helpers/react_on_rails_helper.rb
```diff
@@ -86,14 +86,17 @@ def debug_js(react_component, data_variable, dom_id, trace)
     end
   end
 
+  def render_js_react_element(react_component, props)
+    ReactOnRails::ReactRenderer.new.render_js_react_element(react_component, props).html_safe
```
justin808: 1. Why html safe? Shouldn't be required. The output of the entire helper needs to be marked as html_safe.
2. Why not make the ReactRenderer.render_js_react_element static?

app/helpers/react_on_rails_helper.rb
```diff
@@ -86,14 +86,17 @@ def debug_js(react_component, data_variable, dom_id, trace)
...
+
   def define_render_if_dom_node_present(react_component, data_variable, dom_id, trace)
     <<-JS.strip_heredoc
       var renderIfDomNodePresent = function() {
         var domNode = document.getElementById('#{dom_id}');
         if (domNode) {
           #{debug_js(react_component, data_variable, dom_id, trace)}
-          var reactComponent = #{react_component}(#{data_variable});
-          React.render(reactComponent, domNode);
+          React.render(#{render_js_react_element(react_component, data_variable)}, domNode);
```
justin808: The JS code will be more readable if you breakout the reactComponent into a var.

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,17 @@
...
+    def render_js_react_element(react_component, props)
+      <<-JS
```
justin808: `<<-JS.strip_heredoc`
or maybe you want the spaces on the left?

lib/react_on_rails/react_renderer.rb
justin808: Let's change the signature so it's clear that you're not passing the real props.

```
def render_js_react_element(react_component_name, props_name)
```

app/helpers/react_on_rails_helper.rb
```diff
@@ -49,7 +49,7 @@ def react_component(component_name, props = {}, options = {})
     # Create the HTML rendering part
     if prerender
       render_js_expression = <<-JS
-          renderReactComponent(this.#{reactComponent}, #{props.to_json})
+        this.React.renderToString(#{render_js_react_element(reactComponent, props.to_json)});
```
justin808: Let's fix the lines above changing:

```
     dataVariable = "__#{component_name.camelize(:lower)}Data#{@react_component_index}__"
     reactComponent = component_name.camelize
```

to

```
     data_variable = "__#{component_name.camelize(:lower)}Data#{@react_component_index}__"
     react_component_name = component_name.camelize
```

That way, it's clear which vars are ruby and which are javascript. and change react_component to react_component_name.

Let's break this into a variable: 

```
var reactComponent = #{render_js_react_element(react_component_name, props.to_json)};
```

justin808: Sorry, the above should be:

``` javascript
var reactElement = #{render_js_react_element(react_component_name, props.to_json)};
```

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,17 @@
...
+        #{react_component}.generator ?
+          #{react_component}(#{props}) :
```
justin808: I wish there was something like this for determining a component (rather than an element)
cursor_and_top-level_api___react

@petehunt We could not figure out a way to distinguish between our own generator function to create a react element, versus the function created by calling React.createClass.

CC: @alexfedoseev

app/helpers/react_on_rails_helper.rb
```diff
@@ -49,7 +49,7 @@ def react_component(component_name, props = {}, options = {})
     # Create the HTML rendering part
     if prerender
       render_js_expression = <<-JS
-          renderReactComponent(this.#{reactComponent}, #{props.to_json})
+        this.React.renderToString(#{render_js_react_element(react_component_name, props.to_json)});
```
justin808, line 52: add comment above this line:

```
    # create the server generated html of the react component with props
```

app/helpers/react_on_rails_helper.rb
```diff
@@ -76,24 +76,28 @@ def render_js(js_expression)
...
-        console.log("CLIENT SIDE RENDERED #{react_component} with dataVariable #{data_variable} to dom node with id: #{dom_id}");
+        console.log("CLIENT SIDE RENDERED #{react_component_name} with data_variable #{data_variable} to dom node with id: #{dom_id}");
       JS
     else
       ""
     end
   end
 
-  def define_render_if_dom_node_present(react_component, data_variable, dom_id, trace)
+  def render_js_react_element(react_component_name, props_name)
```
justin808, line 89: Let's change `props_name` to `props_string` b/c it's either a name or the full json version of a hash.

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,21 @@
...
 
-    # "this" does not need a closure as it refers to the "this" defined by the
-    # calling the calling context which is the "this" in the execJs environment.
-    def render_js_react_component
+    # Returns the JavaScript code to generate a React element.
+    # The parameter react_component_name can be a React component or a generator function
+    # that returns a React component. To be invoked as a function, react_component_name
+    # must have the property "generator" set to true and be a function that
+    # takes one parameter, props, that is used to construct the React component.
+    def self.render_js_react_element(react_component_name, props_name)
```
justin808, line 9: change props_name to props_string.

Let's put this comment:

```
   # props_string is either the variable name used to hold the props (client side) or the
   # stringified hash of props from the Ruby server side.
```

lib/react_on_rails/react_renderer.rb
```diff
@@ -1,20 +1,21 @@
 module ReactOnRails
   class ReactRenderer
 
-    # "this" does not need a closure as it refers to the "this" defined by the
```
justin808, line 4: Let's put back this comment.

justin808:
See #20

PR #20, opened 2015-09-13
Support direct react components and server console messages and linting

justin808:
* Added console & error logging
* Catch exceptions and log them on the server side
* Issue with HelloWorld component is not loade properly.
  Visit /focused to see the issue
* Created /focused so that a specific part can be individually tested.
* Lots of cleanup and refactoring.
* Consistency with using suffix App for generator functions
* Added default for generator_function

Replace #19


justin808:
@alexfedoseev @josiasds Please pull this code. I need some help. This is really close, but if you go to http://localhost:3000/focused then you'll see the error which is that the HelloWorld component is not found. This is doing simple client rendering. Same issue server rendering. Thanks.

Also, we added the option for the helper of `generator_function`. We couldn't figure out a better way to identify the function created by React.createClass vs. our own generator function, and thus, we'll just specify this in the helper.

spec/dummy/client/app/components/HelloWorld.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 // Super simple example of the simplest possible React component
-export default class HelloWorld extends React.Component {
```
alex35mil, line 4: Why do we need to split definition and export if we don't do anything with exported class?

spec/dummy/client/app/startup/ClientApp.jsx
```diff
@@ -17,7 +17,7 @@ import HelloWorldContainer from '../components/HelloWorldContainer';
  *  React will see that the state is the same and not do anything.
  *
  */
-window.App = props => {
+let App = props => {
```
alex35mil, line 20: Why `let`, not `const`?

app/helpers/react_on_rails_helper.rb
```diff
@@ -47,53 +56,97 @@ def react_component(component_name, props = {}, options = {})
...
+    server_rendered_html =
+      server_rendered_react_component_html(options, props, react_component_name)
+
+    rendered_output = content_tag(:div,
+                                  server_rendered_html,
+                                  id: dom_id)
+
+    <<-HTML.strip_heredoc.html_safe
+#{data_from_server_script_tag}
+    #{rendered_output}
```
samnang: shouldn't it be aligned?

app/helpers/react_on_rails_helper.rb
```diff
@@ -47,53 +56,97 @@ def react_component(component_name, props = {}, options = {})
...
+    HTML
+  end
+
+  def next_react_component_index
+    @react_component_index ||= -1
+    @react_component_index += 1
```
samnang, line 74: It so strange to keep instance variables in module, but we could tackle it in another PR. I could see we will be able to clean it up. This file is quite long already.

justin808, line 74: @samnang the instance variables work very well as Rails includes the helpers in the view object. I don't see where else we would put this. Note, there's no threading issues here. The `@` variables are like other class instance variables.

spec/dummy/client/app/components/HelloWorld.jsx
justin808, line 4: @alexfedoseev We don't need to split. We were experimenting. Voodoo coding.

spec/dummy/client/app/startup/ClientApp.jsx
justin808, line 20: We were trying to add a property to a function object. For some reason, we can't go 

``` javascript
App.generator = true
```

The generator part is lost when the modules are bundled. Any ideas on why?

README.md
```diff
@@ -88,10 +109,11 @@ Contributions and pull requests welcome!
 
 ## Installation
 
-Add this line to your application's Gemfile:
+Add these linesto your application's Gemfile, sustituting your preferable JavaScript engine.
```
mapreal19: typo: "these lines to"

app/helpers/react_on_rails_helper.rb
```diff
@@ -2,132 +2,196 @@
...
+  #          MyReactComponentApp.generator = true;
+  #    Exposing the react_component_name is necessary to both a plain ReactComponent as well as
+  #      a generator:
+  #    For client rendering, expose the react_component_name on window:
+  #      window.MyReactComponentApp = MyReactComponentApp;
+  #    For server rendering, export the react_component_name on global:
+  #      global.MyReactComponentApp = MyReactComponentApp;
+  #    See spec/dummy/client/app/startup/serverGlobals.jsx and
+  #      spec/dummy/client/app/startup/ClientApp.jsx for examples of this
   # props: Ruby Hash which contains the properties to pass to the react object
```
mapreal19, line 20: missing space here.

justin808, line 20: where is 'here'?

app/helpers/react_on_rails_helper.rb
```diff
@@ -2,132 +2,196 @@
...
-  #    to get the data.
-  #    App(__appData__)
-  #    That way it gets the data you pass from the Rails helper.
-  #
   #  options:
-  #    prerender: <true/false> defaults to true
-  #    trace: <true/false> defaults to false, set to true to print additional info
+  #    prerender: <true/false> (configurable default), set to false when debugging!
+  #    trace: <true/false>  (configurable default), set to true to print additional info
   def react_component(component_name, props = {}, options = {})
```
mapreal19, line 30: I would try to move away of passing a hash for options. https://github.com/bbatsov/ruby-style-guide#no-optional-hash-params

You need comments or read through the body of the method to know what those options are. A better approach in my opinion would be: 

```
react_component(component_name, props: {}, prerender: ReactOnRails.configuration.prerender, trace: ReactOnRails.configuration.trace)
```

That way we wouldn't need to fetch from options, and we could remove those 2 helper methods

app/helpers/react_on_rails_helper.rb
```diff
@@ -2,132 +2,196 @@
...
-    @react_component_index += 1
-
+    # We use this react_component_index in case we have the same component multiple times on the page.
+    react_component_index = next_react_component_index
+    react_component_name = component_name.camelize # Not sure if we should be doing this (JG)
+    dom_id = "#{component_name}-react-component-#{react_component_index}"
+
+    # Setup the page_loaded_js, which is the same regardless of prerendering or not!
+    # The reason is that React is smart about not doing extra work if the server rendering did its job.
+    data_variable_name = "__#{component_name.camelize(:lower)}Data#{react_component_index}__"
```
mapreal19, line 44: any reason of why here is `component_name.camelize(:lower)` instead of `react_component_name`?

app/helpers/react_on_rails_helper.rb
justin808, line 30: Ah, interesting! I hardly ever have method calls for the options. I can see how this could work. @josiasds Please give this a try.

app/helpers/react_on_rails_helper.rb
justin808, line 44: @mapreal19 Just a styling thing so that the name always looks consistent. I think originally, the coder had to know of the the name, but that's no longer true. Still, for `MyBigComponent`, the name becomes `__myBigComponentData__` for the data variable. If there's a good reason not to change the name, we could do that.

app/helpers/react_on_rails_helper.rb
```diff
@@ -2,132 +2,196 @@
...
-
-    rendered_output = content_tag(:div,
-                                  server_rendered_react_component_html,
-                                  id: domId)
-
-    <<-HTML.strip_heredoc.html_safe
-      #{data_from_server_script_tag}
-      #{rendered_output}
-    HTML
+    server_rendered_react_component_html
```
mapreal19, line 92: no need for a temp here. 

```
if ...
   return render_js(...)
else 
   return ""
```

justin808, line 92: that is correct, and if the return is the last statement, then we don't need the return either.

mapreal19, line 92: :+1:

app/helpers/react_on_rails_helper.rb
mapreal19, line 20: line 23: 

```
# props:  (Original)
#  props: 
```

app/helpers/react_on_rails_helper.rb
mapreal19, line 30: @josiasds A good post explaining this: https://robots.thoughtbot.com/ruby-2-keyword-arguments

app/helpers/react_on_rails_helper.rb
mapreal19, line 44: I see. For me it's fine like this

app/helpers/react_on_rails_helper.rb
josiasds, line 30: @justin808 @mapreal19 I'll try that.

mapreal19, line 30: @justin808 @josiasds Also what about introducing an object that outputs the html? There are lots of temp variables, arguments and comments in just one method that makes it hard to follow. An object will make things clearer

http://www.refactoring.com/catalog/extractClass.html :heart:

samnang, line 30: @mapreal19 keyword arguments are good for few arguments, but they are hard to read if we have more than a few with their default values, and also they only support for ruby 2+.

Refactoring may could multiple steps and result may depend on decisions on those small steps.
You may start with [Method Object](https://sourcemaking.com/refactoring/replace-method-with-method-object), then extract different responsibility into individual classes.

justin808, line 30: For the gem, it's think it's OK that we support ruby 2+. On the keyword arguments, I like the pattern shown by @mapreal19. However, does it work for Ruby 2.0 and 2.1?

PR #27, opened 2015-09-16
Linting and rake default rspec task

dylangrafmyre:
Create Rake rspec task the runs rspec for gem top level and rspec for spec/dummy app.
Remove unused test directory within spec/dummy
Setup linting from docker container. Add dockerfile and docker-compose.yml config to build and run lint container.
Add all linting config files.
Edit README for Development setup for test and linting.


justin808:
LGTM

Rakefile
```diff
@@ -1,6 +1,89 @@
...
+end
+
+task :default => :run_rspec
+
+namespace :lint do
+
+  desc "Run Rubocop as shell"
+  task :rubocop do
+    cmd = "rubocop ."
+    sh cmd
```
samnang: Should we just inline `sh "rubocop ."`? I feel it even easier to read in inline.

samnang:
Look good to me. :+1: 

@dylangrafmyre I keep heard a good thing about `Docker`, but I haven't tried it yet. Could you share me what the benefit in this context? Will we use it more in our projects?

dylangrafmyre:
@samnang The benefit in this context is keeping all the linting gems out of .gemspec. Once the container is installed locally, it runs super fast with the linting. Code-Climate does the same thing with their local setup. The have a different container for each linter with a custom .yml file that runs all the linters.

We could set up our linting container how we like, and not have to install linting gems in each project. Only add custom linting configs to override the container config if needed for special reasons per project. Eliminate the need to keep configuring linting for every project. Just pull down the custom hosted container from Docker hub for each project, and linting is automatically ready to go.

Rakefile
justin808: @dylangrafmyre, @samnang is right on this one. No reason to have separate lines.

dylangrafmyre: Fixed with recommendation. Thanks @samnang!

justin808:
@dylangrafmyre Once you are comfortable with this, please merge, and I'll merge into my working branch.

PR #29, opened 2015-09-20
Format js code better examples no server file required

justin808:


spec/dummy/client/app/startup/ClientReduxApp.jsx
```diff
@@ -17,7 +17,7 @@ import HelloWorldContainer from '../components/HelloWorldContainer';
  *  React will see that the state is the same and not do anything.
  *
  */
-let App = props => {
+let ReduxApp = props => {
```
alex35mil: It should be `const`.

spec/dummy/client/app/startup/ServerReduxApp.jsx
```diff
@@ -15,7 +15,7 @@ import reducers from '../reducers/reducersIndex';
 
 import HelloWorldContainer from '../components/HelloWorldContainer';
 
-let App = props => {
+let ReduxApp = props => {
```
alex35mil: Also should be `const`. And to save some typing and show what's exactly we're exporting here, we can skip variable declaration and just export the function:

``` js
export default props => {
  // ...
}
```

alexfedoseev:
LGTM! :+1:

PR #30, opened 2015-09-20
Add capybara and headless poltergeist gems for feature specs and write test

Johnnycus:
Capybara and selenium gems for testing. Tests that enters text in React's inputs and ensures that text is displayed.

### DIFF
--- a/spec/dummy/vendor/assets/javascripts/es5-shim.js
+++ b/spec/dummy/vendor/assets/javascripts/es5-shim.js
@@ -1,0 +1,1596 @@
+/*!
+ * https://github.com/es-shims/es5-shim
+ * @license es5-shim Copyright 2009-2015 by contributors, MIT License
+ * see https://github.com/es-shims/es5-shim/blob/master/LICENSE
+ */
+
+// vim: ts=4 sts=4 sw=4 expandtab
+
+// Add semicolon to prevent IIFE from being passed as argument to concatenated code.
+;
+
+// UMD (Universal Module Definition)
+// see https://github.com/umdjs/umd/blob/master/returnExports.js
+(function (root, factory) {
+    'use strict';
+
+    /* global define, exports, module */
+    if (typeof define === 'function' && define.amd) {
+        // AMD. Register as an anonymous module.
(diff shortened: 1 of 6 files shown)


justin808:
@Johnnycus Please use as examples the tests in https://github.com/shakacode/react-webpack-rails-tutorial/tree/master/spec/features and https://github.com/reactjs/react-rails/tree/master/test.

However, we're using rspec and not minitest.

justin808:
@Johnnycus Pay careful attention to: https://github.com/jnicklas/capybara#using-capybara-with-rspec for the correct API for integration tests. @samnang if you can give Albert a hand, he might write a bunch of tests for us. There's some good tips here: http://betterspecs.org/.

samnang:
@Johnnycus Should we try lightweight headless browser like phanthomjs with https://github.com/jnicklas/capybara#poltergeist?

We could have different test levels here. Unit level to test isolation just for ruby side, and integration tests that should live under dummy app. We suppose to have integration tests with different scenarios, we shouldn't it renders on the example page of dummy app. I'm not sure yet, but I suppose we have different   stub files that we could pull and use it in this, then we will call the rails helper to render the component inside that file, then we expect the output.

Johnnycus:
@samnang Yes, thanks for an advice, I'm going to add poltergeist.

Johnnycus:
@justin808 Ok, thanks!

Johnnycus:
@samnang Now we have headless browser for testing.

spec/dummy/vendor/assets/javascripts/es5-shim.js
```diff
@@ -0,0 +1,1596 @@
+/*!
```
justin808, line 1: @Johnnycus We add dependencies via NPM. Please remove this file and add via npm. See https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/package.json

justin808:
Closing for now. @Johnnycus Feel free to pull the latest and submit some more tests. Tests are **always helpful**. Let us know if we can help you. Thanks for helping!

PR #33, opened 2015-09-21
Update README.md

taylorbrooks:
* Make it obvious that `react_component` is a view helper.
* View helper types


justin808:
Thanks! Really appreciated!

PR #38, opened 2015-09-25
Fix bug in error messages printing

### LINKED ISSUE #36: In the error message handling, we print SERVER_SIDE instead of CLIENT_SIDE
This code in: https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/react_renderer.rb#L103

needs to take an option so what we use on the server and client side is different.

We're printing server side for client side messages!
```
console.error('SERVER SIDE: Exception in server side rendering!');
      if (e.fileName) {
        console.error('SERVER SIDE: location: ' + e.fileName + ':' + e.lineNumber);
      }
      console.error('SERVER SIDE: message: ' + e.message);
      console.error('SERVER SIDE: stack: ' + e.stack);
      msg += 'SERVER SIDE Exception in rendering!\\n' +
        (e.fileName ? '\\nlocation: ' + e.fileName + ':' + e.lineNumber : '') +
        '\\nMessage: ' + e.message + '\\n\\n' + e.stack;
```

josiasds:
This should fix #36.

@justin808 Can you review it?

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -60,8 +60,9 @@ module ReactOnRailsHelper
     data_from_server_script_tag = javascript_tag(page_loaded_js)
 
     # Create the HTML rendering part
-    server_rendered_html =
-      server_rendered_react_component_html(options, props, react_component_name)
+    server_rendered_html, console_script =
+      server_rendered_react_component_html(options, props, react_component_name,
+                                           data_variable_name, dom_id)
 
     rendered_output = content_tag(:div,
                                   server_rendered_html,
@@ -71,6 +72,7 @@ module ReactOnRailsHelper
     <<-HTML.html_safe
 #{data_from_server_script_tag}
 #{rendered_output}
+#{console_script}
     HTML
   end
 
@@ -79,10 +81,12 @@ module ReactOnRailsHelper
     @react_component_index += 1
   end
 
-  def server_rendered_react_component_html(options, props, react_component_name)
+  # Returns Array [0]: html, [1]: script to console log
+  def server_rendered_react_component_html(options, props, react_component_name, data_variable, dom_id)
     if prerender(options)
       render_js_expression = <<-JS
 (function(React) {
+        #{debug_js(react_component_name, data_variable, dom_id, trace(options))}
         var reactElement = #{render_js_react_element(react_component_name, props.to_json, generator_function(options))}
         return React.renderToString(reactElement);
       })(this.React);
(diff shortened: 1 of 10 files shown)


app/helpers/react_on_rails_helper.rb
```diff
@@ -89,12 +89,11 @@ def server_rendered_react_component_html(options, props, react_component_name)
...
       options[:react_component_name] = react_component_name
-      server_rendered_react_component_html =
-        render_js(render_js_expression, options)
+      options[:prerender] = true
+      render_js(render_js_expression, options)
     else
-      server_rendered_react_component_html = ""
+      ''
     end
-    server_rendered_react_component_html
```
justin808, line 97: :+1:

lib/react_on_rails/react_renderer.rb
```diff
@@ -70,7 +71,7 @@ def render_js(js_code, options = {})
       end
     end
 
-    def self.wrap_code_with_exception_handler(js_code, component_name)
+    def self.wrap_code_with_exception_handler(js_code, component_name, prerender)
```
justin808: Let's not use the variable name `prerender` as that's confusing because what's really important is whether or not the code is running server side.

So how about `server_side`? True means running on server. False means running in client.

lib/react_on_rails/react_renderer.rb
```diff
@@ -100,24 +101,31 @@ def self.wrap_code_with_exception_handler(js_code, component_name)
...
 
-      console.error('SERVER SIDE: Exception in server side rendering!');
+      #{render_error_messages(prerender)}
+    }
+      JS
+    end
+
+    private
+
+    def self.render_error_messages(prerender)
```
justin808: `server_side` rather than `prerender`

lib/react_on_rails/react_renderer.rb
```diff
@@ -100,24 +101,31 @@ def self.wrap_code_with_exception_handler(js_code, component_name)
...
+      side = prerender ? 'SERVER' : 'CLIENT'
```
justin808: `server_or_client` rather than `side`

lib/react_on_rails/react_renderer.rb
```diff
@@ -100,24 +101,31 @@ def self.wrap_code_with_exception_handler(js_code, component_name)
...
         (e.fileName ? '\\nlocation: ' + e.fileName + ':' + e.lineNumber : '') +
         '\\nMessage: ' + e.message + '\\n\\n' + e.stack;
 
       var reactElement = React.createElement('pre', null, msg);
       result = React.renderToString(reactElement);
-    }
       JS
     end
 
-    private
```
justin808, line 119: why remove private?

josiasds, line 119: it wasn't removed, it's in the line 109

josiasds:
@justin808 Done!

justin808:
I changed the strategy so that the place we put in the location of the messages is inside of the place where we polyfill on the server.

PR #40, opened 2015-09-27
Pass options to content tag

justin808:
* Options passed through to content tag
* Now you can have replay_console set to false and logging_server_set to
  true and server messages go only to the server. replay_console only
  affects the browser console display.


justin808:
@alexfedoseev provided review. Choosing to put html options in with other options to keep consistent with api at https://github.com/reactjs/react-rails/.

PR #42, opened 2015-09-27
Fix no console message on server crash

justin808:
* Fixes crash using render_js and no console msgs
* Fixes missing html_safe when calling content_tag


alexfedoseev:
Looks good! :+1:

PR #44, opened 2015-09-28
[WIP] Refactor to pull js code into its own js file

justin808:


justin808:
Here's the result! Way cleaner client side generated code and way cleaner ruby!

justin808:
Paired all the way with @samnang! Only thing that got us stuck was the execJS not wanting a semicolon at the end of the string!!!

PR #49, opened 2015-10-05
Adds Integration Tests

mapreal19:


.travis.yml
```diff
@@ -3,10 +3,20 @@ rvm:
...
+env:
+ - export RAILS_ENV=test
+before_script:
+- cd ../../
+- export DISPLAY=:99.0
+- sh -e /etc/init.d/xvfb start
 script:
 - rake run_rspec:gem
-- rake run_rspec:dummy
+- DRIVER=selenium_firefox rake run_rspec:dummy
```
justin808, line 19: why not chrome?

justin808:
- [ ] Lots of options to cover, many of which are covered by the new ERB pages I added. Main ones are: `generator_function`, `prerender`, `id`, `class`.
- [ ] Especially important to cover server rendering and check the page source contains the div elements.
- [ ] See if we can easily verify the console output. If not easy, don't worry about this one.
- [ ] How's the coverage?

Thanks! looking super!

.travis.yml
dylangrafmyre, line 19: @justin808 because this is for Travis CI, and that is what they use. Pretty much all CI's use firefox browser for the selenium web driver.

justin808:
@mapreal19 How about we merge this and open a new PR for the next ones?

mapreal19:
@justin808 yeah makes sense!

mapreal19:
@justin808 I'd merge so we can benefit from those tests and slowly improve coverage. 

It's good too if someone finds a bug, he would be able to write a spec on top of these changes

justin808:
Thanks @mapreal19! Let's get the coverage up!

PR #52, opened 2015-10-09
supports 0.13 and 0.14 with one gem

rstudner:
Tested with the react-webpack sample project using both React 0.13 and 0.14.

There is an outstanding branch on the react-webpack project:
https://github.com/shakacode/react-webpack-rails-tutorial/pull/115

That handles all the bits and pieces to make the sample project work with 0.14

### DIFF
--- a/app/assets/javascripts/react_on_rails.js
+++ b/app/assets/javascripts/react_on_rails.js
@@ -18,7 +18,7 @@
         if (domNode) {
           var reactElement = createReactElement(componentName, propsVarName, props,
             domId, trace, generatorFunction);
-          React.render(reactElement, domNode);
+          ReactOnRails.provideClientReact().render(reactElement, domNode);
         }
       }
       catch (e) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/react_on_rails.js
```diff
@@ -150,4 +150,19 @@
 
     return consoleReplay;
   }
+
+  ReactOnRails.provideClientReact = function() {
```
justin808: Should we expose this globally?

I suggest we do it the same we declared:

https://github.com/rstudner/react_on_rails/blob/react-0.14-support/app/assets/javascripts/react_on_rails.js#L78

``` javascript
  function createReactElement(componentName, propsVarName, props, domId, trace, generatorFunction) {
```

justin808:
Looks good with one suggestion.

@rstudner @dylangrafmyre @mapreal19, I wonder if there is some way we can specify an env variable to configure which version of React. Maybe there's a way to configure the CI to do this.

app/assets/javascripts/react_on_rails.js
```diff
@@ -78,7 +78,7 @@
   function createReactElement(componentName, propsVarName, props, domId, trace, generatorFunction) {
     if (trace) {
       console.log('RENDERED ' + componentName + ' with data_variable ' +
-        propsVarName + ' to dom node with id: ' + domId);
+          propsVarName + ' to dom node with id: ' + domId);
```
justin808, line 81: why the white space diff?

PR #53, opened 2015-10-09
Add Support generator to build a template webpack setup

aaronvb:
```
> rails generate react_on_rails:install
```

Created generators for:

- React and Redux folders
- Assets folder
- Symlinks for images and fonts of they exist
- Blank global jsx files
- Update gitignore to ignore node modules
- Insert generated and react_on_rails includes into application.js
- Copy react_on_rails.rb config to app/initializers
- Nodejs server files for the hot reload
- Webpack config files
- npm package

I moved rails engine to its own file to be more explicit.

Regarding the npm package.json, I copied it from the tutorial repo. I'm not sure if all of those packages should be included in the gem, or if you want to slim it down to the barebones react libraries.


justin808:
Checking this out now! @rstudner can you give this a try? or any others on the team. I'll take a quick look as well.

lib/generators/react_on_rails/directory_generator.rb
```diff
@@ -0,0 +1,37 @@
+require "rails/generators"
+
+module ReactOnRails
+  module Generators
+    class DirectoryGenerator < Rails::Generators::Base
+      source_root File.expand_path("../templates", __FILE__)
+
+      def create_react_and_redux_dir
```
justin808: Could we, should we, have the flux system (redux) to be optional?

lib/generators/react_on_rails/directory_generator.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      def create_assets_dir
+        empty_directory "client/assets/stylesheets"
+      end
+
+      def create_assets_symlinks
+        create_link "client/assets/fonts", "../../app/assets/fonts" if Dir.exist?("app/assets/fonts")
+
+        create_link "client/assets/images", "../../app/assets/images" if
+        Dir.exist?("app/assets/images")
```
justin808: white space... @dylangrafmyre, the linter should catch this, right?

lib/generators/react_on_rails/setup_files_generator.rb
```diff
@@ -0,0 +1,26 @@
+require "rails/generators"
+
+module ReactOnRails
+  module Generators
+    class SetupFilesGenerator < Rails::Generators::Base
+      source_root File.expand_path("../templates", __FILE__)
+
+      def update_git_ignore
+        append_to_file ".gitignore" do
+          "# React on Rails\nnpm-debug.log\nnode_modules"
```
justin808: we need the directory of generated files in the rails gitignore:

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/.gitignore#L30

```
# Generated js bundles
/app/assets/javascripts/generated/*
```

justin808:
For file: lib/generators/react_on_rails/templates/client/app/startup/clientGlobals.jsx and some of the other files, let's put in a few useful doc links on what is supposed to go in these files. We can provide:

1. URL to example file in the tutorial
2. A few sample lines

lib/generators/react_on_rails/setup_files_generator.rb
```diff
@@ -0,0 +1,26 @@
...
+        end
+      end
+
+      def update_application_js
+        insert_into_file "app/assets/javascripts/application.js",
+                         before: "//= require_tree ." do
```
justin808: There's no guarantee that `require_tree .` will be in `application.js`. I've never used that.

This is what we have in the tutorial:

```
// Need to be on top to allow Poltergeist test to work with React.
//= require es5-shim/es5-shim
//= require react_on_rails

// It is important that generated/vendor-bundle must be before bootstrap since it is exposing jQuery and jQuery-ujs
//= require generated/vendor-bundle
//= require generated/app-bundle

//= require bootstrap-sprockets
//= require turbolinks
```

lib/generators/react_on_rails/directory_generator.rb
rstudner: I think it is okay if the project is 'opinionated' on Redux (and willing to switch to the most valuable Flux implementation in the future if it switches).  Have an option, to 'opt-out' of having redux as the default.  But it is okay if the generator provides 'react_on_rails' opinion on the best seed/starter setup.

lib/generators/react_on_rails/setup_files_generator.rb
aaronvb: Good point, I was basing this off a fresh rails project. I'll have to look through the Thor docs a little more to see what I can do to get it before bootstrap and before tree if tree exists.

rstudner: +1  I think deleting require_tree in application.js is a first two-minutes operation for many projects :)

lib/generators/react_on_rails/templates/client/index.jade
```diff
@@ -0,0 +1,14 @@
+doctype html
```
justin808, line 2: For all the generated files, let's leave a comment of the source github URL:

For example, for this file: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/index.jade

The reasons are:
1. To see where to get more documentation
2. Inevitably, the gem will get out of sync with the tutorial

lib/generators/react_on_rails/templates/client/package.json
```diff
@@ -0,0 +1,81 @@
+{
```
justin808, line 1: Note, we can't put comments in the .json files (at least not conveniently).

Maybe we can place a nice README.md in the created /client directory that explains some things to customize.

In the below file, those would be:
1. node version
2. dependencies (documentation on how to update them here: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/README.md)
3. difference between devDependencies and dependencies (anything needed for heroku deployment needs to go in "dependencies". Anything not needed for heroku deployment should go in "devDependencies".

lib/generators/react_on_rails/templates/client/server.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-console, func-names, no-var */
+var bodyParser = require('body-parser');
+var webpack = require('webpack');
+var WebpackDevServer = require('webpack-dev-server');
+var jade = require('jade');
+var sleep = require('sleep');
+var config = require('./webpack.client.hot.config');
+
+var comments = [
```
justin808: It might make sense to have an option, defaulting to true, to copy the sample JSX files so that the node server can start.

At the min, let's have a comment referring how to setup basic REST API handling, with a reference to: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/server.js

lib/generators/react_on_rails/templates/client/package.json
aaronvb, line 1: We can also output text from a file, ie: README, at the end of the generation.

http://guides.rubyonrails.org/generators.html#readme

lib/generators/react_on_rails/templates/client/webpack.client.hot.config.js
```diff
@@ -0,0 +1,63 @@
+// Run like this:
+// cd client && node server.js
```
justin808: I think this should be `npm start`. @alexfedoseev can you confirm?

lib/generators/react_on_rails/templates/client/server.js
```diff
@@ -0,0 +1,42 @@
...
+server.app.use('/', function(req, res) {
+  var locals = {
+    props: JSON.stringify(comments),
+  };
+  var layout = process.cwd() + '/index.jade';
+  var html = jade.compileFile(layout, { pretty: true })(locals);
+  res.send(html);
+});
+
+server.listen(3000, 'localhost', function(err) {
```
justin808: FYI, we might change this default: https://github.com/shakacode/react-webpack-rails-tutorial/issues/116

lib/generators/react_on_rails/templates/client/webpack.client.rails.config.js
```diff
@@ -0,0 +1,45 @@
...
+config.entry.vendor.unshift(
+  'es5-shim/es5-shim',
+  'es5-shim/es5-sham'
+);
+config.entry.app.push('./app/startup/clientGlobals');
+
+// See webpack.common.config for adding modules common to both the webpack dev server and rails
+
+config.module.loaders.push(
+  {test: /\.jsx?$/, loader: 'babel-loader', exclude: /node_modules/}
```
justin808, line 28: We're missing this file: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/.babelrc

justin808:
Wow! Great start! Let's consider getting this in v1.0.0, or very shortly thereafter. @aaronvb Don't hesitate to ask us if you have any questions. If you need any help with the suggestions, please let us know. This is very useful.

lib/generators/react_on_rails/templates/client/package.json
justin808, line 1: Let's do both!

lib/generators/react_on_rails/directory_generator.rb
justin808: I think @rstudner is correct, with an option to opt-out. :+1:

aaronvb:
@justin808 Thanks for the feedback. I'll be going through this over the next few days and will join you guys in Slack for any questions. A lot of what I have in here was extracted from the webpack tutorial, so I expected lots of changes and fixes.

lib/generators/react_on_rails/setup_files_generator.rb
aaronvb: @justin808 How are the js files being generated on a production environment?

justin808: @aaronvb the JS files are generated in production using this: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/lib/tasks/assets.rake

I think we should copy brakeman, ci, and linters as well from https://github.com/shakacode/react-webpack-rails-tutorial/tree/master/lib/tasks

justin808: @aaronvb Since we'll be generating the files, we'll assume the directory matches the defaults that we generate. In other words, we'll be fixing our recommended defaults.

aaronvb: @justin808 Got it, thanks!

aaronvb: @justin808 if I add the linters task, rake assets:precompile bombs out with 'LoadError: cannot load such file -- scss_lint/rake_task'. I've traced it to https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/Gemfile#L73

Should we require that gem to be added or should I not add the linters task?

justin808: Good questions I think we should add the linting gems for now. It's easy enough to comment out. We can add flags later to configure this.

aaronvb: @justin808 What do you think about adding an exception in the linter task. With rescue LoadError nil, the task doesn't get loaded if the gem doesn't exist and the precompile doesn't throw an error. Tested with rake -T and rake assets:precompile, the task gets completely ignored.

If you want to go with your plan, should we let the user add the gem or have the generator add it.

justin808: Good idea on the catching of the exception.

I'd say we go with adding the gems from the generator. Or at least commented out.

One thing is that we'd have to be careful not to create duplicates.

So maybe let's add the rake task with the lines to copy into the gemfile?

lib/generators/react_on_rails/templates/client/webpack.client.hot.config.js
alex35mil: Confirm, we should use `npm scripts` to run client's operations.

aaronvb:
I've pushed updates to all the files noted above and added:

* Made changes to the generators to separate react and redux.
* Added the JS linter files so ```npm run lint``` works now
* Readme's

Install stays the same, except it doesn't create the redux folders or include the redux packages in package.json
```
rails generate react_on_rails:install
```

Install with redux will do the same as the regular install, but create the redux directories and include redux in the package.json.
```
rails generate react_on_rails:install_with_redux
```

Please review the copy in both README files.

rstudner:
Do we want to add a command line parameter for indicating 0.13 vs 0.14 setup (exposing ReactDOM and ReactDOMServer as appropriate)

lib/generators/react_on_rails/setup_files_generator.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      def update_git_ignore
+        append_to_file ".gitignore" do
+          "# React on Rails\nnpm-debug.log\nnode_modules\n\n# Generated js bundles\n/app/assets/javascripts/generated/*"
+        end
+      end
+
+      def update_application_js
+        append_to_file "app/assets/javascripts/application.js" do
```
derencius: I have one app that uses application.js.coffee and the generator breaks here. Also, one suggestion is to use prepend instead of append, because probably those requires should be on top.

bellow an attempt to add support for application.js.coffee and to show a message if no application.js/js.coffee file is found.

``` ruby
      def update_application_js
        data = "\n// It is important that generated/vendor-bundle must be before bootstrap since it is exposing jQuery and jQuery-ujs\n//= require generated/vendor-bundle\n//= require generated/app-bundle\n//= require react_on_rails\n\n"
        if File.exists?("app/assets/javascripts/application.js")
          prepend_to_file "app/assets/javascripts/application.js", data
        elsif File.exists?("app/assets/javascripts/application.js.coffee")
          prepend_to_file "app/assets/javascripts/application.js.coffee", data
        else
          puts "** app/assets/javascripts/application.js was not found. please add the following content to your main javascript file:\n\n#{data}\n\n"
        end
      end


```

aaronvb: @derencius Awesome, thanks for the feedback. I'll will test this out now and update the method.

lib/generators/react_on_rails/templates/lib/tasks/linters.rake
```diff
@@ -0,0 +1,71 @@
+if %w(development test).include? Rails.env
+  namespace :lint do
+    # require "rubocop/rake_task"
+    # require "slim_lint/rake_task"
+    require "scss_lint/rake_task"
```
derencius: if the gem scss_lint is not in the Gemfile, running "rake -T" is going to fail here. 

I could adjust it to be flexible about the scss_lint being installed with this code:

``` ruby

  desc "See docs for task 'scss_lint'"
    task :scss do
      begin
        require 'scss_lint/rake_task'
        SCSSLint::RakeTask.new do |t|
          t.files = ["app/assets/stylesheets/", "client/assets/stylesheets/"]
        end
        Rake::Task[:scss_lint].invoke
      rescue LoadError
        puts "** add gem 'scss_lint' to your Gemfile for scss linting."
      end
    end
```

aaronvb: @derencius This was discussed earlier, and @justin808 decided to forgo the exception and just provide a note to the user at the end of the install to add the gem. I'm still open to further discuss this

justin808: @aaronvb @derencius I'm OK to adding the linter gems together.

I like @derencius suggested catch block.

justin808:
@rstudner @aaronvb Regarding any command line param for the old react, there's *probably* no need if we're doing a generator. As far as I can tell, React 0.14 gives warnings for the deprecations, so those are relatively easy to fix.

That being said, there could be cross dependencies that might not yet like React 0.14.

So if it's easy to add a param, then sure, but this is not a huge priority, at least not until at least one person really wants it.

rstudner:
Correct -- is definitely not necessary.  Just suggested it since some people get all upset when their chrome console has warnings in it they don't understand ;)

justin808:
@rstudner You mean terminal, right?

justin808:
@aaronvb would you say this is ready to go? I'll take a look at this tomorrow, and give it a spin.

aaronvb:
@justin808 In my opinion, yes. If all looks good to you, I will update the README.md in the gem with instructions on using the generate install.

rstudner:
@justin808 Sorry I missed this in the email.  yeah, I meant terminal sorry.  The use of the word 'console' was an extremely poor choice on my part :)

rstudner:
So my first run.. npm 2.14.4 and node v0.12.7 'failed'

         run  cd client && webpack --config webpack.client.rails.config.js from "."
sh: webpack: command not found
         run  cd client && webpack --config webpack.server.rails.config.js from "."
sh: webpack: command not found

Basically webpack didn't get installed etc.  Checking/looking into why now

liorbrauer:
@aaronvb I tested this PR on my project, using Rails 3.2.22/Ruby 2.2.2, it works very nice. Saving A LOT of boilerplate. Thanks for this :+1: 

I will report if I find anything useful.

rstudner:
Does the Procfile.dev etc need to get created (is that part of the boilerplate)

I do realize there are other ways to startup the rails & node servers in parallel etc.

But since the react-webpack-rails-tutorial project has the 'opinion' to do Procfile.dev should this?

justin808:
Procfile.dev is very handy and well worth adding.

justin808:
I'm going to take this over for now. Thanks @aaronvb!

dylangrafmyre:
From CI build. Rubocop errors.

build__230_-_shakacode_react_on_rails_-_travis_ci

gugl:
I recommend setting the .babelrc stage to 2 or 1 as stage 0 is not considered recommended for production code.

alexfedoseev:
@gugl We're on the stage 0 mainly because of static properties. Moreover base libs (like `redux` & `react-router`) are compiled with stage 0, so anyway we're using it in production.

justin808:
Next up:

1. Add linter gems to Gemfile
2. Add hello world react component -- let's use the same simple setup we use for the spec/dummy app
3. Don't run webpack in the fresh generator install, as it doesn't pick the version just installed. P

Per running steps:

From directory of `react_on_rails`, with a test app named "react_on_rails_gen"
```
cd ..
rails new react_on_rails_gen
cd react_on_rails_gen
git init
git add .
git commit -m "Rails new plus react_on_rails"
```

Edit the Gemfile, adding these 2 lines:
```ruby
gem 'react_on_rails', path: '../react_on_rails'
gem 'therubyracer'
```
Note the relative path to the react_on_rails gem.

```bash
bundle
git commit -am "added react_on_rails gem"
rails generate react_on_rails:install
```

Then you can see the changes that the generator did, ready for a commit.

Then run:

```bash
foreman start -f Procfile.dev
```

Then visit ports 3000 and 4000.

That's it!

justin808:
For extra credit:

1. Add bootstrap sass options
2 Add some tests.

However, we can do those in a separate PR.

justin808:
@aaronvb Let's run some basic sanity check that nvm, npm, node are installed and abort the generator with a link to get those installed. I think this is going to be the basic stumbling block for most of the ruby community when they first see this.

aaronvb:
Added linter gems to Gemfile.

Added a simple react hello world component that demonstrates client rendering, server rendering, and hot reload.

You can generate the hello world files by using ```rails generate react_on_rails:hello_world``` After generating the initial install.

Removed running webpack after initial install.

Added a check for npm and node before installing, will abort with a message containing a url to the libraries.

aaronvb:
Fixing CI errors now.

justin808:
@aaronvb 
Just pulled latest. Looking *AWESOME*

1. ci.rake file is empty
2. clientGlobals.jsx should have the hello world setup
3. putting an empty images directory. That should probably be a symlink: https://github.com/shakacode/react-webpack-rails-tutorial/tree/master/client/assets Same for fonts.
4. Need to put in current versions of node and npm: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/package.json#L7
5. Nice job on client/README.md
6. Let's have server globals be there
7. Let's have the hello world example do server rendering.

Overall, very awesome!

lib/generators/react_on_rails/templates/lib/tasks/assets.rake
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  desc "Compile assets with webpack"
+  task :webpack do
+    sh "cd client && npm run build:client"
+    sh "cd client && npm run build:server"
+  end
+
+  task :clobber do
+    rm_rf "#{RailsReactTutorial::Application.config.root}/app/assets/javascripts/generated/client-bundle.js"
```
necronet: Where is `RailsReactTutorial` defined?

aaronvb: @necronet This needs to be updated, thanks for catching this. The assets.rake was pulled from https://github.com/shakacode/react-webpack-rails-tutorial - sorry for the confusion.

justin808: @necronet Good point. @aaronvb Please use `Rails.application.config.root`.

justin808: @aaronvb We should update the tutorial as well, in case others copy it. @robwise Can you do a search and replace on that one? Thanks!

aaronvb:
@justin808 

1. *ci.rake file is empty*
    Fixed

2. *clientGlobals.jsx should have the hello world setup*
    The clientGlobals.jsx gets updated with the hello world setup when using the ```rails generate react_on_rails:hello_world``` generator. I can have the hello world example code created with the main install generator.

3. *putting an empty images directory. That should probably be a symlink: https://github.com/shakacode/react-webpack-rails-tutorial/tree/master/client/assets Same for fonts.*
   Images and font directories are symlinks and can only be symlinked if they exist. In a new Rails app, the font directory does not exist so a symlink is not created. See https://github.com/aaronvb/react_on_rails/blob/generators/lib/generators/react_on_rails/setup_files_generator.rb#L8

4. *Need to put in current versions of node and npm: https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/package.json#L7*
    Fixed

5.    

6. *Let's have server globals be there*
    See 2

7. *Let's have the hello world example do server rendering.*
    https://github.com/aaronvb/react_on_rails/blob/generators/lib/generators/react_on_rails/templates/hello_world/app/views/pages/index.html.erb#L4

lib/generators/react_on_rails/templates/lib/tasks/assets.rake
robwise: @justin808 @aaronvb Changes made in tutorial on https://github.com/shakacode/react-webpack-rails-tutorial/pull/129

lib/generators/react_on_rails/templates/client/server.js
```diff
@@ -0,0 +1,36 @@
...
+  var locals = {
+  };
+  var layout = process.cwd() + '/index.jade';
+  var html = jade.compileFile(layout, { pretty: true })(locals);
+  res.send(html);
+});
+
+server.listen(4000, 'localhost', function(err) {
+  if (err) console.log(err);
+  console.log('Listening at localhost:3000...');
```
necronet: Shouldnt this be 'Listening at localhost:4000...' ?

alex35mil: Should be:

``` js
console.log('Listening at localhost:' + this.address().port + '...');
```

robwise:
The work for this PR branch has been merged to a new branch on the original repo: https://github.com/shakacode/react_on_rails/tree/add-generators

robwise:
@justin808 I propose that we close this so as to ensure all future progress is made on #70, right?

PR #55, opened 2015-10-12
Support React 0.14

justin808:
* Also fixed linting errors

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -54,39 +54,33 @@ desc "Runs all linters. Run `rake -D lint` to see all available lint options"
 task lint: ["lint:lint"]
 
 namespace :docker do
-  def docker_sh(cmd)
-    docker_init = "eval \"$(docker-machine env default)\";"
-    sh "#{docker_init} #{cmd}"
-  end
-
   desc "Run Rubocop linter from docker"
   task :rubocop do
-    docker_sh "#{docker_init} docker-compose run lint rake lint:rubocop"
+    sh "docker-compose run lint rake lint:rubocop"
   end
 
   desc "Run ruby-lint linter from docker"
   task :ruby do
-    docker_sh "docker-compose run lint rake lint:ruby"
+    sh "docker-compose run lint rake lint:ruby"
   end
 
   desc "Run scss-lint linter from docker"
   task :scss do
-    docker_sh "docker-compose run lint rake lint:scss"
+    sh "docker-compose run lint rake lint:scss"
   end
 
   desc "Run eslint linter from docker"
   task :eslint do
-    docker_sh "docker-compose run lint rake lint:eslint"
+    sh "docker-compose run lint rake lint:eslint"
   end
 
   desc "Run jscs linter from docker"
   task :jscs do
-    docker_sh "docker-compose run lint rake lint:jscs"
+    sh "docker-compose run lint rake lint:jscs"
   end
-
   desc "Run all linting from docker"
   task :lint do
-    docker_sh "docker-compose run lint rake lint"
+    sh "docker-compose run lint rake lint"
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -54,33 +54,39 @@ desc "Runs all linters. Run `rake -D lint` to see all available lint options"
 task lint: ["lint:lint"]
 
 namespace :docker do
+  def docker_sh(cmd)
+    docker_init = "eval \"$(docker-machine env default)\";"
+    sh "#{docker_init} #{cmd}"
+  end
+
   desc "Run Rubocop linter from docker"
   task :rubocop do
-    sh "docker-compose run lint rake lint:rubocop"
+    docker_sh "#{docker_init} docker-compose run lint rake lint:rubocop"
   end
 
   desc "Run ruby-lint linter from docker"
   task :ruby do
-    sh "docker-compose run lint rake lint:ruby"
+    docker_sh "docker-compose run lint rake lint:ruby"
   end
 
   desc "Run scss-lint linter from docker"
   task :scss do
-    sh "docker-compose run lint rake lint:scss"
+    docker_sh "docker-compose run lint rake lint:scss"
   end
 
   desc "Run eslint linter from docker"
   task :eslint do
-    sh "docker-compose run lint rake lint:eslint"
+    docker_sh "docker-compose run lint rake lint:eslint"
   end
 
   desc "Run jscs linter from docker"
   task :jscs do
-    sh "docker-compose run lint rake lint:jscs"
+    docker_sh "docker-compose run lint rake lint:jscs"
   end
+
   desc "Run all linting from docker"
   task :lint do
-    sh "docker-compose run lint rake lint"
+    docker_sh "docker-compose run lint rake lint"
   end
 end
 
--- a/app/assets/javascripts/react_on_rails.js
+++ b/app/assets/javascripts/react_on_rails.js
@@ -22,7 +22,8 @@
         }
       }
       catch (e) {
-        handleError(e, componentName, provideClientReact());
+        ReactOnRails.handleError({e: e, componentName: componentName,
+          serverSide: false});
       }
     };
 
@@ -68,28 +69,21 @@
       htmlResult = provideServerReact().renderToString(reactElement);
     }
     catch (e) {
-      htmlResult = handleError(e, componentName, provideServerReact());
+      htmlResult = ReactOnRails.handleError({e: e, componentName: componentName,
+        serverSide: true});
     }
 
     consoleReplay = ReactOnRails.buildConsoleReplay();
     return JSON.stringify([htmlResult, consoleReplay]);
   };
 
-  function createReactElement(componentName, propsVarName, props, domId, trace, generatorFunction) {
-    if (trace) {
-      console.log('RENDERED ' + componentName + ' with data_variable ' +
-        propsVarName + ' to dom node with id: ' + domId);
-    }
-
-    if (generatorFunction) {
-      return this[componentName](props);
-    } else {
-      return React.createElement(this[componentName], props);
-    }
-  }
-
   // Passing either componentName or jsCode
-  function handleError(e, componentName, jsCode, properReact) {
+  ReactOnRails.handleError = function(options) {
+    var e = options.e;
+    var componentName = options.componentName;
+    var jsCode = options.jsCode;
+    var serverSide = options.serverSide;
+
     var lineOne =
       'ERROR: You specified the option generator_function (could be in your defaults) to be\n';
     var lastLine =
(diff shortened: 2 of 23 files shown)


Rakefile
```diff
@@ -54,33 +54,39 @@ desc "Runs all linters. Run `rake -D lint` to see all available lint options"
...
 namespace :docker do
+  def docker_sh(cmd)
+    docker_init = "eval \"$(docker-machine env default)\";"
+    sh "#{docker_init} #{cmd}"
+  end
+
   desc "Run Rubocop linter from docker"
   task :rubocop do
-    sh "docker-compose run lint rake lint:rubocop"
+    docker_sh "#{docker_init} docker-compose run lint rake lint:rubocop"
```
justin808: @dylangrafmyre Does this look right?

app/assets/javascripts/react_on_rails.js
```diff
@@ -68,28 +69,21 @@
       htmlResult = provideServerReact().renderToString(reactElement);
     }
     catch (e) {
-      htmlResult = handleError(e, componentName, provideServerReact());
+      htmlResult = ReactOnRails.handleError({e: e, componentName: componentName,
```
justin808, line 72: @rstudner, please review the changes here.

app/assets/javascripts/react_on_rails.js
```diff
@@ -127,14 +121,14 @@
...
     console.error('stack: ' + e.stack);
-    msg += 'Exception in rendering!\n' +
-      (e.fileName ? '\nlocation: ' + e.fileName + ':' + e.lineNumber : '') +
-      '\nMessage: ' + e.message + '\n\n' + e.stack;
-
-    var reactElement = React.createElement('pre', null, msg);
-
-    return properReact.renderToString(reactElement);
-  }
+    if (serverSide) {
```
justin808, line 124: Using a boolean for serverSide as we don't need the rendering part unless we're on the server.

lib/react_on_rails/server_rendering_pool.rb
```diff
@@ -1,62 +1,20 @@
...
 module ReactOnRails
   class ServerRenderingPool
     def self.reset_pool
       options = { size: ReactOnRails.configuration.server_renderer_pool_size,
                   timeout: ReactOnRails.configuration.server_renderer_pool_size }
-      @@js_context_pool = ConnectionPool.new(options) { create_js_context }
+      @js_context_pool = ConnectionPool.new(options) { create_js_context }
     end
 
-    def self.eval_js(js_code)
```
justin808, line 12: I moved the private methods below.

lib/react_on_rails/server_rendering_pool.rb
```diff
@@ -89,13 +41,67 @@ def self.server_render_js_with_console_logging(js_code)
...
-              Rails.logger.info { "[react_on_rails] #{match[:msg]}" }
-            end
+            Rails.logger.info { "[react_on_rails] #{match[:msg]}" } if match
           end
         end
       end
       result
     end
+
+    class << self
```
justin808, line 51: This should make the module methods private. At least it does it on a class. @samnang correct?

lib/react_on_rails/server_rendering_pool.rb
```diff
@@ -1,62 +1,20 @@
...
+# Based on the react-rails gem.
+# None of these methods should be called directly.
+# See app/helpers/react_on_rails_helper.rb
 module ReactOnRails
   class ServerRenderingPool
     def self.reset_pool
       options = { size: ReactOnRails.configuration.server_renderer_pool_size,
                   timeout: ReactOnRails.configuration.server_renderer_pool_size }
-      @@js_context_pool = ConnectionPool.new(options) { create_js_context }
+      @js_context_pool = ConnectionPool.new(options) { create_js_context }
```
justin808, line 11: Changed to using the class instance variable. I think this would only make a difference if there was subclassing, but this is the linters preferred way. @samnang, @rstudner, @alexfedoseev Agree?

spec/dummy/app/views/pages/client_side_log_throw.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= render "header" %>
```
justin808, line 1: Added a client only ERB file.

spec/dummy/client/app/components/HelloWorld.jsx
```diff
@@ -10,12 +11,12 @@ class HelloWorld extends React.Component {
   }
 
   _handleChange() {
-    const name = React.findDOMNode(this.refs.name).value;
+    const name = Common.reactFindDOMNode(this.refs.name).value;
```
justin808, line 14: This has changed for React 0.14, so abstracting this function.

spec/dummy/client/app/components/HelloWorldWithLogAndThrow.jsx
```diff
@@ -7,10 +7,10 @@ class HelloWorldWithLogAndThrow extends React.Component {
   }
 
   render() {
-    console.log("console.log in HelloWorld")
-    console.warn("console.warn in HelloWorld")
-    console.error("console.error in HelloWorld");
-    throw "throw in HelloWorldContainer";
+    console.log('console.log in HelloWorld');
```
justin808, line 10: Fixed lots of linting errors.

spec/dummy/client/app/startup/ClientReduxApp.jsx
```diff
@@ -21,11 +22,20 @@ const ReduxApp = props => {
   const combinedReducer = combineReducers(reducers);
   const store = applyMiddleware(middleware)(createStore)(combinedReducer, props);
 
-  const reactComponent = (
-    <Provider store={store}>
+  let reactComponent;
+  if (Common.react013()) {
```
justin808, line 26: Need 2 different ways of doing this to avoid warnings.

Rakefile
dylangrafmyre: @justin808 I think we will have problems with this. 

 Certain people might name their docker-machine instance something different than `default`. Like `dev` for example. Hardcoding the machine name here will cause an error.

Linux users will not be using docker-machine or boot2docker as a  lightweight linux vm to run docker in since it runs native on linux. This is only needed for mac and windows users.  `docker_init` will cause `rake` to abort for linux users.

I think it is better for users to place `eval "$(docker-machine env default)"` in their .zshrc or .bashrc files. or manually call `eval "$(docker-machine env default)"`if they need to execute.

justin808: OK

spec/dummy/client/app/startup/ClientReduxApp.jsx
alex35mil, line 26: I'm a bit confused that it's exposed to user land, and actually I don't understand why in 0.14 version we pass component to Redux's `Provider` instead of function, that returns component. Maybe I missed something.

app/helpers/react_on_rails_helper.rb
```diff
@@ -89,10 +89,18 @@ def react_component(component_name, props = {}, options = {})
     <<-HTML.html_safe
 #{data_from_server_script_tag}
 #{rendered_output}
-#{replay_console(options) ? console_script : ""}
+#{replay_console(options) ? console_script : ''}
     HTML
   end
 
+  def sanitized_props_string(props)
```
alex35mil, line 96: :+1:

spec/dummy/client/app/startup/ClientReduxApp.jsx
justin808, line 26: This changes in React 0.14.

you get a nice error when you do this

app/helpers/react_on_rails_helper.rb
justin808, line 96: Next PR.

@samnang 's idea didn't work -- need to find the right thing to escape.

spec/dummy/client/app/components/HelloWorld.jsx
alex35mil, line 14: Is this our internals or user should use this abstracted API?

justin808, line 14: I was trying to reuse the same code for testing both React 0.13 and 0.14.

justin808, line 14: So I'm abstracting the couple places there are differences in the apps. For the react_on_rails Gem, that's all hidden inside the gem.

alex35mil, line 14: I see, maybe worth to add comment why it used here this way.

justin808, line 14: I think this is just for testing, so it's not critical. The https://github.com/shakacode/react-webpack-rails-tutorial/ is geared toward a complete example.

spec/dummy/client/app/startup/ClientReduxApp.jsx
alex35mil, line 26: Yep, here it is: https://github.com/rackt/redux/issues/867

spec/dummy/client/app/components/HelloWorld.jsx
alex35mil, line 14: :+1:

app/assets/javascripts/react_on_rails.js
rstudner, line 72: made it private again -- is fine.

app/assets/javascripts/react_on_rails.js
rstudner, line 124: Makes sense.  The top part is common code, the bottom is server side only.  (y)

spec/dummy/client/app/startup/ClientReduxApp.jsx
rstudner, line 26: Yep.  I ran into this the first day I changed it over to 0.14.  Someone elses PR against the sample project made the same sort of corrections against that project :)

spec/dummy/client/app/utils/Common.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+const Common = {
+  react013() {
+    return React.version.match(/^0\.13/);
```
rstudner, line 6: Smart -- since React always exists.

spec/dummy/client/webpack.client.js
```diff
@@ -3,6 +3,7 @@ const path = require('path');
 module.exports = {
   entry: [
     'startup/clientGlobals',
+    'react-dom'
```
rstudner, line 6: Interesting.  What's this for?

spec/dummy/client/webpack.server.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   context: __dirname,
   entry: [
     'startup/serverGlobals',
+    'react-dom/server'
```
rstudner, line 7: same here.

justin808:
@dylangrafmyre Any word on getting these to pass?

justin808:
See #56. Contains code for this PR plus more test code.

PR #56, opened 2015-10-12
Update dependencies tests react013

justin808:
See https://github.com/shakacode/react_on_rails/pull/55 for the changes without the clutter of the extra test files.


justin808:
@dylangrafmyre This branch has tests for 0.13 added after updating the gems.

Let's see if we can get this to pass CI!

dylangrafmyre:
@justin808 module fsevents1.0.2 is breaking the builds because that version does not support OS linux.

image

justin808:
@dylangrafmyre @mapreal19 Any word on getting these to pass?

justin808:
What should we do about fsevents1.0.2? @dylangrafmyre

dylangrafmyre:
@justin808 dependency `chokidar@1.2.0` has a optional dependency on `fsevents^1.0.0`. Is there a way to exclude this from production since no watching is necessary?

alexfedoseev:
@dylangrafmyre @justin808 This is strange stuff: it's coming from `webpack`, but I've updated everything today in `fng` and it works fine, checkout PR #129. `fsevents@1.0.2` is there and travis is green.

dylangrafmyre:
@alexfedoseev FnG uses circleci and react_on_rails uses travisci. I wonder if there is something different with the linux builds or versions?

alexfedoseev:
> FnG uses circleci and react_on_rails uses travisci

I need to get some sleep :sweat_smile:

dylangrafmyre:
@justin808 @alexfedoseev @mapreal19 Got past the `npm install` issue with `fsevents` by upgrading to `node 4.2.0` and `npm 3.3.6` on the CI builds. Found they fixed the issue with npm release 3.2.0

https://github.com/npm/npm/issues/8921

Now the failing test need to be addressed.

justin808:
@rstudner, @dylangrafmyre, @geoffevason, @mapreal19, @samnang, @alexfedoseev @josiasds this one is **READY**.

Any suggestions?

While we could put in some lower level unit tests, I think it's sufficient to ship.

@dylangrafmyre Coveralls does not seem to be using the integration tests inside of spec/dummy. How can we fix that?

Issue #62, opened 2015-10-18
Tune up Coveralls


justin808:
@dylangrafmyre @robwise Let's tune up Coveralls https://coveralls.io/github/shakacode/react_on_rails

1. I can't find react_on_rails_helper.rb listed in the coverage.
2. The coverage of some of the other files is a bit dubious. It seems that initialization of the gem is not getting flagged as covered.

![2015-10-17_21-17-58](https://cloud.githubusercontent.com/assets/1118459/10562528/988c38be-7514-11e5-865d-38684f6fe91d.png)

![2015-10-17_21-18-54](https://cloud.githubusercontent.com/assets/1118459/10562530/b8f1dd8e-7514-11e5-837d-fd60265d0f53.png)

![2015-10-17_21-20-07](https://cloud.githubusercontent.com/assets/1118459/10562536/e5c4495a-7514-11e5-9dc3-53d22d90479b.png)

![2015-10-17_21-21-04](https://cloud.githubusercontent.com/assets/1118459/10562538/05cc6c46-7515-11e5-8625-7dba4c95d0be.png)

robwise:
@justin808 Are we good on this after #64 ?

PR #64, opened 2015-10-19
Tune up Coveralls and SimpleCov

robwise:
Addresses #62

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,1 +1,0 @@
-SimpleCov.start if ENV["COVERAGE"]
--- a/Rakefile
+++ b/Rakefile
@@ -11,14 +11,14 @@ namespace :run_rspec do
   task :dummy do
     # TEST_ENV_NUMBER is used to make SimpleCov.command_name unique in order to
     # prevent a name collision
-    sh %( cd spec/dummy && DRIVER=selenium_firefox COVERAGE=true TEST_ENV_NUMBER=1 rspec )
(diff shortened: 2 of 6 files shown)


robwise:
I'm not sure why or how the `Gemfile.lock` got changed

spec/dummy/Gemfile.lock
```diff
@@ -4,7 +4,7 @@ PATH
     react_on_rails (1.0.0.pre)
       connection_pool
       execjs (~> 2.5)
-      rails (>= 4.0)
+      rails (>= 3.2)
```
justin808: @robwise This shouldn't show up in your diff. It's a good practice to rebase your change on top of master, so long as nobody else is branching off of your branch.

robwise: @justin808 I'm pretty sure it was a glitch somehow, since I didn't want to change it and master didn't change it. As you can see in that link, my computer was constantly changing it back to that. I restarted and it stopped doing it so that I could set it straight and force push a new commit that doesn't include that change (so the log will be cleaner than having a "Undo change Gemfile.lock" commit).

robwise:
Force pushed new commit [52306dd](https://github.com/robwise/react_on_rails/commit/52306dd6e0248d2a74f46e125c0a3cd654db23ef)

justin808:
Thanks Rob! Looks great!

PR #43, opened 2015-09-28
Added some basic test coverage for the helpers

mapreal19:
- Resolves #5

PR #66, opened 2015-10-20
Add unit tests to helper

robwise:
A reconciled version of PR #43

### DIFF
--- a/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
+++ b/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
@@ -58,11 +58,11 @@ describe ReactOnRailsHelper, type: :helper do
     end
   end
 
-  describe "#render_js" do
-    subject { render_js("this.HelloString.world()") }
+  describe "#server_render_js" do
+    subject { server_render_js("this.HelloString.world()") }
 
     let(:hello_world) do
-      "Hello WORLD! Will this work?? YES! Time to visit Maui\n"
+      "    Hello WORLD! Will this work?? YES! Time to visit Maui\n    \n"
     end
 
     it { expect(self).to respond_to :react_component }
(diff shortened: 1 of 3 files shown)


spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
```diff
@@ -34,4 +34,40 @@
...
+      subject { react_component("App", {}, id: id) }
+
+      let(:id) { "shaka_div" }
+
+      it { is_expected.to include id }
+      it { is_expected.not_to include react_component_div }
+    end
+  end
+
+  describe "#server_render_js" do
```
robwise, line 61: These tests were originally calling just `#render_js` and got method missing, so I assumed they meant `#server_render_js`.

spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
```diff
@@ -34,4 +34,40 @@
...
+    subject { server_render_js("this.HelloString.world()") }
+
+    let(:hello_world) do
+      "    Hello WORLD! Will this work?? YES! Time to visit Maui\n    \n"
```
robwise: I had to change the expected message to include the whitespace at the beginning and end, as well as the 2nd `/n`. Is this desired behavior?

justin808: @robwise Good question. I'm not sure why there's extra white space. Might be worth a quick look. Also, the test could trim the subject, so as not to require the extra whitespace.

justin808: Let's definitely see where this whitespace is coming from.

robwise:
@justin808 Okay, I think this should be good to go. Whitespace issue fixed in commit 8c04ed2. All of Mario's tests now pass!

justin808:
LGTM. Thanks! Great job @mapreal19 and @robwise!

mapreal19:
Thanks @robwise! :clap:

robwise:
@mapreal19 you bet! nice job!

PR #75, opened 2015-11-03
Server reload bundle fix

rstudner:
@robwise let me know if I have this right.  Was pretty straight forward, just had to make sure not to include the 'extra stuff'.

### DIFF
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -11,19 +11,19 @@ module ReactOnRails
       logging_on_server: true,
       generator_function: false,
       trace: Rails.env.development?,
-      reload_server_js_every_request: Rails.env.development?,
+      development_mode: Rails.env.development?,
       server_renderer_pool_size: 1,
       server_renderer_timeout: 20)
   end
 
   class Configuration
     attr_accessor :server_bundle_js_file, :prerender, :replay_console,
-                  :generator_function, :trace, :reload_server_js_every_request,
+                  :generator_function, :trace, :development_mode,
                   :logging_on_server, :server_renderer_pool_size,
                   :server_renderer_timeout
 
     def initialize(server_bundle_js_file: nil, prerender: nil, replay_console: nil,
-                   generator_function: nil, trace: nil, reload_server_js_every_request: nil,
+                   generator_function: nil, trace: nil, development_mode: nil,
                    logging_on_server: nil, server_renderer_pool_size: nil,
                    server_renderer_timeout: nil)
       if File.exist?(server_bundle_js_file)
@@ -36,10 +36,10 @@ module ReactOnRails
       self.replay_console = replay_console
       self.logging_on_server = logging_on_server
       self.generator_function = generator_function
-      if reload_server_js_every_request.nil?
-        self.reload_server_js_every_request = Rails.env.development?
+      if development_mode.nil?
+        self.development_mode = Rails.env.development?
       else
-        self.reload_server_js_every_request = reload_server_js_every_request
+        self.development_mode = development_mode
       end
       self.trace = trace.nil? ? Rails.env.development? : trace
 
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -11,17 +11,19 @@ module ReactOnRails
       logging_on_server: true,
       generator_function: false,
       trace: Rails.env.development?,
+      reload_server_js_every_request: Rails.env.development?,
       server_renderer_pool_size: 1,
       server_renderer_timeout: 20)
   end
 
   class Configuration
     attr_accessor :server_bundle_js_file, :prerender, :replay_console, :generator_function, :trace,
-                  :logging_on_server, :server_renderer_pool_size, :server_renderer_timeout
+                  :reload_server_js_every_request, :logging_on_server, :server_renderer_pool_size, :server_renderer_timeout
 
     def initialize(server_bundle_js_file: nil, prerender: nil, replay_console: nil,
-                   generator_function: nil, trace: nil, logging_on_server: nil,
-                   server_renderer_pool_size: nil, server_renderer_timeout: nil)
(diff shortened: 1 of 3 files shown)


robwise:
@rstudner Awesome! This all looks good to me besides there being still one linter thingy:

```
Offenses:
lib/react_on_rails/configuration.rb:39:45: C: Avoid multi-line ?: (the ternary operator); use if/unless instead.
      self.reload_server_js_every_request = reload_server_js_every_request.nil? ?
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rstudner:
Haha on the last failure.  "line too long."  "don't break up line".  haha

robwise:
Yeah lol, with the ternary operator it's too long for the line length, and then they don't want you to use multi-line ternary operators, so we should go with an if-else

lib/react_on_rails/configuration.rb
```diff
@@ -11,17 +11,21 @@ def self.configuration
       logging_on_server: true,
       generator_function: false,
       trace: Rails.env.development?,
+      reload_server_js_every_request: Rails.env.development?,
```
justin808: this is correct, being the default

lib/react_on_rails/server_rendering_pool.rb
```diff
@@ -11,6 +11,15 @@ def self.reset_pool
       @js_context_pool = ConnectionPool.new(options) { create_js_context }
     end
 
+    def self.reset_pool_if_server_bundle_was_modified
+      return unless ReactOnRails.configuration.reload_server_js_every_request
+      file_mtime = File.mtime(ReactOnRails.configuration.server_bundle_js_file)
+      @server_bundle_timestamp ||= file_mtime
```
justin808: so this line above says "don't reset pool if first time?"

I'd rather remove this line and have

```
return if @server_bundle_timestamp == file_mtime || @server_bundle_timestamp.nil?
```

because I read that as return if the timestamp did not change or we have not yet set the server_bundle_timestamp

lib/react_on_rails/configuration.rb
robwise: @justin808 although, it's technically not what's going on, it's reloading server js only if the server-bundle has been modified.

lib/react_on_rails/server_rendering_pool.rb
```diff
@@ -11,6 +11,15 @@ def self.reset_pool
...
+      return if @server_bundle_timestamp == file_mtime
+      ReactOnRails::ServerRenderingPool.reset_pool
+      @server_bundle_timestamp = file_mtime
```
justin808, line 20: This line should be where we [actually read the file, right](https://github.com/shakacode/react_on_rails/pull/75/files#diff-10c7b74a600ef3ecde9dca86ed42c16eR90)?

So long as we don't have multiple threads, there's no timing issues. Consider if there could be race conditions.

lib/react_on_rails/configuration.rb
```diff
@@ -32,6 +36,11 @@ def initialize(server_bundle_js_file: nil, prerender: nil, replay_console: nil,
       self.replay_console = replay_console
       self.logging_on_server = logging_on_server
       self.generator_function = generator_function
+      if reload_server_js_every_request.nil?
+        self.reload_server_js_every_request = Rails.env.development?
+      else
+        self.reload_server_js_every_request = reload_server_js_every_request
+      end
```
justin808, line 43:
```
self.server_renderer_pool_size = self.reload_server_js_every_request ? 1 : server_renderer_pool_size
```

https://github.com/shakacode/react_on_rails/pull/75/files#diff-aec9762d9d3a2eca38fcc1363cffca27R47

lib/react_on_rails/server_rendering_pool.rb
robwise, line 20: @justin808 that's where we had it, but if you put it there you're going to be in a new instance every request, so in order to keep the value of the timestamp variable, you'd need to make it a class variable, which is why you had us move it to server_rendering_pool.

lib/react_on_rails/configuration.rb
robwise: @justin808 what about:

```
self.server_renderer_pool_size = self.reload_server_js_every_request || server_renderer_pool_size`

File: lib/react_on_rails/server_rendering_pool.rb (hunk 2 above)
Comment (justin808) [inline lib/react_on_rails/server_rendering_pool.rb]: Let's rename the @server_bundle_timestamp to @read_server_bundle_timestamp

File: lib/react_on_rails/server_rendering_pool.rb (hunk 3 above)
Comment (justin808) [inline lib/react_on_rails/server_rendering_pool.rb:20]: OK -- I agree. This makes sense.

File: lib/react_on_rails/server_rendering_pool.rb (hunk 2 above)
Comment (justin808) [inline lib/react_on_rails/server_rendering_pool.rb]:
```
first_time = @server_bundle_timestamp.nil?
return if first_time || @server_bundle_timestamp == file_mtime 
```

File: lib/react_on_rails/configuration.rb (hunk 1 above)
Comment (justin808) [inline lib/react_on_rails/configuration.rb]: It should be named:

```
development_mode
```

I think @robwise has a good point. We're not accurate and this is asking the user of the gem to think too much.

Comment (rstudner) [inline lib/react_on_rails/configuration.rb]: Question - I change this on line 14.

what about here:
```
      if development_mode.nil?
        self.reload_server_js_every_request = Rails.env.development?
      else
        self.reload_server_js_every_request = development_mode
      end
```

Do we want to rename out internal variable development_mode also?

lib/react_on_rails/server_rendering_pool.rb
rstudner: I'm not following these two comments/am not sure what to change.

justin808:
@rstudner are we going to put those tests in with this fix? or afterwards?
cc: @robwise

robwise:
@justin808 It's on his fork so maybe just merge it and I will make another PR for its tests?

PR #80, opened 2015-11-05
fixes the variable name change

rstudner:
this was missed due to the timing of making the variable name change w/ the generators branch in flight etc etc.


dylangrafmyre:
@rstudner LGTM :+1:

PR #82, opened 2015-11-05
Linters are not the default

justin808:
Now you have to specify --linters or -L to add linters

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ In the following screenshot you can see the actual HTML rendered for a side-by-s
   rails generate react_on_rails:install
   ```
 
-3. NPM install:
+3. NPM install. Make sure you are on a recent version of node, preferably using nvm.
 
   ```bash
   npm install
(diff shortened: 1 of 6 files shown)


robwise:
Side note, having to make the changes in both `package.json`s kind of highlights how un-DRY it is to have one for non-redux and one for redux. Maybe I should just use the same ERB conditionals technique for those packages that differ between the two and that way we can have just a single `client/package.json` template.

justin808:
@robwise I just rebased down to one commit...

justin808:
@samnang @alexfedoseev @mapreal19 @geoffevason @dylangrafmyre @robwise We just lost a bunch of comments after a rebase and push force.

This is making me think that http://reviewable.io might be reasonable to try out for our public and internal projects. 

Here's a list: https://www.quora.com/What-are-the-best-code-review-tools

rupurt:
@justin808 you may already know this but GitHub now allows you to configure [protected branches](https://github.com/blog/2051-protected-branches-and-required-status-checks) to disable things like force push.

justin808:
@rupurt Yes, saw that a month or two ago. I'll set that up. force-push should be disabled for master.

justin808:
@robwise Yes, I suggest 100% that we use a ERB partial for parts like this!

justin808:
Switching gears.

Turning linters back on for JS (stuff in /client) and having option `--ruby-linters` for turning on ruby linters and the default for that is false. 

CC: @robwise

mapreal19:
:+1: for reviewable!

PR #83, opened 2015-11-05
Clobber all files in generated directory

rupurt:


lib/generators/react_on_rails/templates/base/base/lib/tasks/assets.rake
```diff
@@ -19,8 +19,6 @@ namespace :assets do
   end
 
   task :clobber do
-    rm_rf "#{Rails.application.config.root}/app/assets/javascripts/generated/vendor-bundle.js"
-    rm_rf "#{Rails.application.config.root}/app/assets/javascripts/generated/client-bundle.js"
-    rm_rf "#{Rails.application.config.root}/app/assets/javascripts/generated/server-bundle.js"
+    rm_rf Dir.glob(Rails.root.join("app/assets/javascripts/generated/*"))
```
justin808: :+1:

justin808: @rupurt while we're at it, let's change to `rm_r` if that's available for Rake.

no need for rm -rf, right?

rupurt: :+1: can't see why not. Do you want me to make it 2nd commit or amend the previous?

justin808:
@rupurt Please rebase on top of master. This will go out in 1.0.2.

rupurt:
@justin808 done

PR #87, opened 2015-11-06
Improve Linter Generators

robwise:
Several commits that incrementally improve the linter generation functionalities. Primarily, this change splits the linters generator into two new generators, one for linters written in JavaScript and one for linters written in Ruby. The JavaScript linter generator is now on by default, but the ruby linter is off by default.

Syntax:

`--skip-js-linters` (alias `-j`)
`--ruby-linters` (alias `-L`)

For the sake of passing linters, I have removed the `sleep` and `bodyParser` packages from client/package.json. I believe that these were only used by client/server.js for express middleware when implementing async functionality. Since the HelloWorld app does not have any async functionality, they were not needed. I have merely commented them out, however, along with the example commented-out async code.

Lastly, the missing Ruby linter configs have been added. These are verbatim what was currently in the tutorial app, and will probably need a quick review from @justin808 when he has the chance.


lib/generators/react_on_rails/install_generator.rb
```diff
@@ -20,19 +7,25 @@ class InstallGenerator < Rails::Generators::Base
...
       class_option :server_rendering,
                    type: :boolean,
                    default: false,
-                   desc: "Configure for server-side rendering of webpack JavaScript",
+                   desc: "Add necessary files and configurations for server-side rendering",
                    aliases: "-S"
-      # --linters
-      class_option :linters,
+      # --skip-js-linters
+      class_option :skip_js_linters,
```
justin808, line 19: js_linters

default: true

no alias needed since default is true, right?

lib/generators/react_on_rails/install_generator.rb
```diff
@@ -41,7 +34,8 @@ def run_generators
         invoke "react_on_rails:base"
         invoke "react_on_rails:react_no_redux" unless options.redux?
         invoke "react_on_rails:react_with_redux" if options.redux?
-        invoke "react_on_rails:linters" if options.linters?
+        invoke "react_on_rails:js_linters" unless options.skip_js_linters?
```
justin808, line 37:
``` ruby
invoke "react_on_rails:js_linters" if options.js_linters?
```

lib/generators/react_on_rails/templates/base/base/client/server.js
```diff
@@ -21,6 +19,8 @@ var server = new WebpackDevServer(webpack(config), {
   },
 });
 
+// var bodyParser = require('body-parser');
```
justin808, line 28: Nice

comment that you would need to:

``` bash
npm i --save body-parser sleep
```

lib/generators/react_on_rails/templates/redux/base/client/app/lib/middlewares/loggerMiddleware.js
```diff
@@ -12,7 +12,7 @@ export default function logger({ getState }) {
 
     // We can't console.log immutable objects out-of-the-box.
     const immutableState = getState();
-    const readableState = _.reduce(immutableState, (result, immutable, key) => {
+    const readableState = _.reduce(immutableState, (object, immutable, key) => {
       result[key] = immutable.toJS();
```
justin808: shouldn't this now be 

```
object[key] = immutable.toJS();
```

lib/generators/react_on_rails/install_generator.rb
robwise, line 19: We have now discussed in Slack and decided to use the Rails `new` generator API as our example to emulate since it will be familiar for users.

https://github.com/rails/rails/blob/master/railties/lib/rails/generators/app_base.rb

lib/generators/react_on_rails/templates/redux/base/client/app/lib/middlewares/loggerMiddleware.js
robwise: @justin808 This is something I actually wanted to talk to @alexfedoseev about on Skype or something. It's a really weird situation going on with variable shadowing and scope that's actually interesting in just understanding how JavaScript handles this scenario. Even when what is now `object` was also called `result`, despite my intuition, `result[key]` was referring to the object _outside_ the function and I believe this is the intended behavior. In other words, in both versions, what is now `object` was never being used.

justin808: Lodash docs: https://lodash.com/docs#reduce
The iteratee is bound to thisArg and invoked with four arguments: (accumulator, value, index|key, collection). 

Initial value is on line 17: `{}`

When you used `result` as your accumulator, it hid the parent one so it worked just fine. You could have called it pretty much anything in the outer closure.

result[key] was 100% not supposed to refer to the result of the outer scope.

lib/generators/react_on_rails/install_generator.rb
robwise, line 37: see above

lib/generators/react_on_rails/templates/base/base/client/server.js
robwise, line 28: added

lib/generators/react_on_rails/templates/redux/base/client/app/lib/middlewares/loggerMiddleware.js
robwise: @justin808 Yep, you're right. I think the fact that even with the variable names sorted out, the code was just wrong so it was confusing me, it's fixed now.

justin808:
LGTM -- please rebase down to one commit

we should probably merge these without the --no-ff option from now on when squashed down to one commit

that requires doing this on the command line

robwise:
@justin808 Just an idea before I rebase: I've done this thing before in my own repo where I do the `--no-ff` thing, but don't rebase. This way it will be obvious in the visual log that these commits are all bundled under a single overarching idea which is summarized by the merge commit. If you want to just stick to the main commits, you only read those ones in the leftmost column:

image

PR #88, opened 2015-11-06
Make generator recognize application.scss

### LINKED ISSUE #86: Generator Should Recognize application.scss File in Addition to application.css.scss
Currently the generator is looking to append Sass imports into the `app/assets/stylesheets/application.css` file (which it renames to `application.css.scss`) or an `application/assets/stylesheets/application.css.scss`, but it doesn't look for simply `application.scss` and therefore throws an error.

This is extracted from: https://github.com/shakacode/react_on_rails/issues/78#issuecomment-154257465

robwise:
Makes the generator recognize `application.scss` files instead of only looking for `application.css.scss` by just forcing all variations to be `application.scss`.

Fixes #86

We're just forcing any version of application.css or
application.css.scss to be application.scss

This avoids a lot of headache when trying to reference the file
in multiple places, and ensures that users could have either variation without the generator failing.

### DIFF
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -32,6 +32,7 @@ module ReactOnRails
            client/bootstrap-sass.config.js).each { |file| copy_file(base_path + file, file) }
       end
 
+      # rename to application.scss from application.css or application.css.scss
       def force_application_scss_naming_if_necessary
         base_path = "app/assets/stylesheets/"
         application_css = "#{base_path}application.css"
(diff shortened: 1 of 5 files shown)


justin808:
@robwise what if there's no file at all named application.xxxxx

justin808:
LGTM

robwise:
@justin808 In that case, the generator outputs a message to the user indicating that it was unable to find the file and continues with the build. Do you think we should handle this differently, such as just creating our own application.scss file?

justin808:
@robwise I'd say just create a new file. User can them move code to another file.

justin808:
@robwise I just did a merge commit via the UI on github.

robwise:
@justin808 shouldn't we rebase master, not merge it? It's just noise, right?

justin808:
@robwise go ahead and merge, or if you have time, rebase this to one commit on top of master.

PR #95, opened 2015-11-12
Add unicorn

robwise:
Adds unicorn files necessary for Heroku Deployment

### DIFF
--- a/docs/additional_reading/heroku_deployment.md
+++ b/docs/additional_reading/heroku_deployment.md
@@ -3,7 +3,9 @@ The generator has created the necessary files and gems for deployment to Heroku.
 
 + `Procfile`: used by Heroku and Foreman to start the server
 + `.buildpacks`: used to install Ruby and Node environments
++ `config/unicorn.rb`: Unicorn config file
 + `12factor` gem: required by Heroku
++ `unicorn` gem: app server used by Heroku
 + `lib/tasks/assets.rake`: rake task that generates your JavaScript bundles for production.
 
 ## How to Deploy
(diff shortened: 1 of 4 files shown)


justin808:
Since we missed the 1.0.2 bus, let's put this in with:

`--heroku` option

and have all the heroku bits go with it.

robwise:
@justin808  roger that no problem, should we default to not having it enabled then?

justin808:
Not sure on that one! Maybe check the project if it has 12factor already and let that be the factor! SMART!

docs/additional_reading/heroku_deployment.md
```diff
@@ -3,7 +3,9 @@ The generator has created the necessary files and gems for deployment to Heroku.
 
 + `Procfile`: used by Heroku and Foreman to start the server
 + `.buildpacks`: used to install Ruby and Node environments
++ `config/unicorn.rb`: Unicorn config file
```
justin808, line 6: @robwise why unicorn?

puma is the current heroku default

dylangrafmyre, line 6: @justin808 I think @robwise has been reading to many articles about dying unicorns posted in our slack channel and he really meant `config/puma.rb` :)

robwise, line 6: @justin808 @dylangrafmyre I was told to model the generators off of the tutorial, so that's what I did. I can switch it to Puma, no problem (I switched from Unicorn to Puma on my own personal Rails app a long time ago).

robwise:
Opted to go for Puma in #100

PR #100, opened 2015-11-15
Make Heroku Generator Optional

robwise:
Add Puma to Heroku Generator, enable the option to turn it on or off, and default it to off.


justin808:
@robwise This looks great, but this seems to be for Heroku and not puma.

PR #102, opened 2015-11-17
Use data attrs when rendering react_components instead of script tags

ryanaip:
Client side rendering scripts do not get called when using the forward/back buttons with turbolinks, because turbolinks does not re-evaluate script tags when navigating through the history.

Instead of creating script tags throughout the page, use html elements with data attributes specifying each react component to be rendered, and iterate through all of those elements whenever the page loads.

E.g.,
```html
<div class="react-component"
     data-component-name="App"
     data-props="{&quot;name&quot;:&quot;My Test Name&quot;}"
     data-trace="false"
     data-generator-function="false"
     data-expect-turbolinks="true"
     data-dom-id="El-react-component-0"></div>
<div id="El-react-component-0"></div>
```

This has the added benefit of removing global data variables from being set on the window object and simplifying the rendering code.

(Note that it will be possible to support turbolinks navigation with the still-on-master `data-turbolinks-eval=always` feature of turbolinks.  But this solution doesn't depend on that upgrade and is a bit cleaner anyway.)

### DIFF
--- a/app/assets/javascripts/react_on_rails.js
+++ b/app/assets/javascripts/react_on_rails.js
@@ -1,66 +1,10 @@
 (function() {
   this.ReactOnRails = {};
-
-  ReactOnRails.clientRenderReactComponent = function(options) {
-    var componentName = options.componentName;
-    var domId = options.domId;
-    var propsVarName = options.propsVarName;
-    var props = options.props;
-    var trace = options.trace;
-    var generatorFunction = options.generatorFunction;
-    var expectTurboLinks = options.expectTurboLinks;
-
-    this[propsVarName] = props;
-
-    var renderIfDomNodePresent = function() {
-      try {
-        var domNode = document.getElementById(domId);
-        if (domNode) {
-          var reactElement = createReactElement(componentName, propsVarName, props,
-            domId, trace, generatorFunction);
-          provideClientReact().render(reactElement, domNode);
-        }
-      }
-      catch (e) {
-        ReactOnRails.handleError({
-          e: e,
-          componentName: componentName,
-          serverSide: false,
-        });
-      }
-    };
-
-    var turbolinksInstalled = (typeof Turbolinks !== 'undefined');
-    if (!expectTurboLinks || (!turbolinksInstalled && expectTurboLinks)) {
-      if (expectTurboLinks) {
-        console.warn('WARNING: NO TurboLinks detected in JS, but it is in your Gemfile');
-      }
-
-      document.addEventListener('DOMContentLoaded', function(event) {
-        renderIfDomNodePresent();
-      });
-    } else {
-      function onPageChange(event) {
-        var removePageChangeListener = function() {
-          document.removeEventListener('page:change', onPageChange);
-          document.removeEventListener('page:before-unload', removePageChangeListener);
-          var domNode = document.getElementById(domId);
-          provideClientReact().unmountComponentAtNode(domNode);
-        };
-
-        document.addEventListener('page:before-unload', removePageChangeListener);
-
-        renderIfDomNodePresent();
-      }
-
-      document.addEventListener('page:change', onPageChange);
-    }
-  };
+  var turbolinksInstalled = (typeof Turbolinks !== 'undefined');
 
   ReactOnRails.serverRenderReactComponent = function(options) {
     var componentName = options.componentName;
     var domId = options.domId;
-    var propsVarName = options.propsVarName;
     var props = options.props;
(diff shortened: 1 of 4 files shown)


justin808:
@ryanaip Looks great! Let's get in touch and go through this ASAP. Please send me an email to justin@shakacode.com.

Questions: 

1. if there's a lot of props, how much of a performance hit is it to put that in as JSON in a data attribute?
2. Did you test this with server rendering?

I totally agree on the need to solve the back/forward buttons with turbolinks.

@alexfedoseev @samnang @robwise @rstudner Take a look.

app/assets/javascripts/react_on_rails.js
```diff
@@ -156,10 +100,59 @@
     return consoleReplay;
   };
 
-  function createReactElement(componentName, propsVarName, props, domId, trace, generatorFunction) {
+  function forEachComponent(fn) {
+    var els = document.getElementsByClassName('react-component');
```
justin808, line 104: I'd suggest a longer name to avoid any chance of class conflicts:

`'js-react-on-rails-react-component'`

robwise, line 104: @justin808 aren't we good since we're inside of an IIFE?

justin808:
@ryanaip what do you think of the alternative change to make the script tags to use the `data-turbolinks-eval=always` tag (if turbolinks is installed):



https://github.com/rails/turbolinks

All things being equal, I'd prefer to see the more direct JS script tag, but I'm totally open to going over the pros and cons.

app/assets/javascripts/react_on_rails.js
justin808, line 104: IIFE applies to the global namespace for JavaScript not CSS.

robwise, line 104: @justin808 oh sorry I thought you were talking about the function name, now I got ya!

alexfedoseev:
@ryanaip, thanks for PR! I'm not familiar with turbolinks api, but big amount of data on DOM nodes can cause performance issues during reflows (correct me if I'm wrong). So my vote for leaving props inside JS scripts and wait until `data-turbolinks-eval` will be released.

/cc: @justin808

ryanaip:
@alexfedoseev, @justin808 — You're right about the new `data-turoblinks-eval` functionality.  I went back and forth on whether to just use that, but it seemed cleaner to me to separate the javascript from the list of components to be rendered (specifically, react_on_rails.js feels easier to understand with my changes).  But, per @justin808's comment, that decision is a stylistic one.  Up to you all on which you think is best, although I personally think it's cleaner not to have script tags in the document body if you can help it.

Regarding @alexfedoseev's comment about storing data, it looks like the official [react-rails](https://github.com/reactjs/react-rails#rendering--mounting) gem uses data attributes as well, see also their [javascript](https://github.com/reactjs/react-rails/blob/master/lib/assets/javascripts/react_ujs.js.erb). There is something to be said for consistency across projects to help people who may be moving between the two — We could even use the same attribute naming convention and start to build a bit of a standard.

I'd guess that data performance becomes an issue either if you put very large data into the components or have a large number of components on the page.  I don't know how much that will happen in practice.

justin808:
Merged in #113

jopdeklein:
Is it feasible to make this a configurable choice between the script and the data attributes methods? I understand the motivation for switching to the data attributes approach, but in our case where we pre-render a lot of HTML on the server this means the raw data precedes the rendered HTML. The means the initial render is postponed until all meta data is downloaded - not a good concept for performance.

justin808:
@jopdeklein I'd be happy to include a PR to offer the alternative. However, I could not seem to ensure that the non-data-attributes way worked OK with Turbolinks.

jopdeklein:
@justin808 Thanks that would be very useful, we don't use turbolinks at the moment so that's not an issue.

justin808:
@jopdeklein We need to be sure that any changes don't break TurboLinks.

justin808:
@jopdeklein Can you see if putting the store definition at the bottom of the page (new feature of v3) solves your problem. This way, the data attributes are not connected to the server rendering element.

jopdeklein:
@justin808 From what I can see in the changelogs it should work. I won't have time this week but will test it ASAP. Thanks for the help.

justin808:
@jopdeklein Sweet! That wasn't an intentional use case, but sounds good. Can you please report back? Maybe I will update the doc to recommend putting the store definition at the very bottom. 

CC: @alexfedoseev

PR #104, opened 2015-11-17
Add data-turbolinks-eval always to script tags

justin808:
Forward and back buttons don't re-evaluate scripts for turbolinks unless
data-turbolinks-eval is set to 'always'. However, this is a turbolinks
v3 feature which is not yet released.

The branch contains a link to turbolinks at master, and this setting
works for react_on_rails.

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -52,6 +52,8 @@ module ReactOnRailsHelper
     # The reason is that React is smart about not doing extra work if the server rendering did its job.
     data_variable_name = "__#{component_name.camelize(:lower)}Data#{react_component_index}__"
     turbolinks_loaded = Object.const_defined?(:Turbolinks)
+    script_options = {}
+    script_options['data-turbolinks-eval'] = "always" if turbolinks_loaded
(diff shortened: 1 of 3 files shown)


justin808:
I prefer the inline scripts, but this depends on the unreleased version of turbolinks.

CC: @ryanaip, @robwise, @alexfedoseev

justin808:
I just checked Turbolinks: https://github.com/rails/turbolinks/issues/628

Looks like this is shipping anytime now, so I'm OK saying that react_on_rails will require an update to turbolinks.

justin808:
@alexfedoseev @ryanaip, This solution may also be missing this:

https://github.com/shakacode/react_on_rails/pull/102/files#diff-11f0a5146e6819b284aaa17cb1e40f9dR186

Basically, should we be sure that we call something like this:

```
function unmount(el) {
    var domId = el.getAttribute('data-dom-id');
    var domNode = document.getElementById(domId);
    provideClientReact().unmountComponentAtNode(domNode);
  }
```
 I've added #105 for an overall discussion of the best approach.

justin808:
It's unclear when and if TurboLinks will support the `data-turbolinks-eval="always"` option.

Thus, going with data attributes for now.

PR #108, opened 2015-11-20
Have generator create application.js file if not found

### LINKED ISSUE #78: Generator raises error when application.js is not found
We currently have our application split up into multiple smaller applications so don't use the main `application.js` file. It would nice if the generator allowed you to specify the name of the manifest file e.g.

```bash
rails generate react_on_rails:install -m my_cool_app
```

robwise:
Fixes #78
Generator used to put a setup error message when it did not find
application.js. Now, it will simply just create one with the
necessary data in it.

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -71,14 +71,15 @@ module ReactOnRails
 
           // bootstrap-sprockets depends on generated/vendor-bundle for jQuery.
           //= require bootstrap-sprockets
+
         DATA
 
-        application_js_path = "app/assets/javascripts/application.js"
-        application_js = dest_file_exists?(application_js_path) || dest_file_exists?(application_js_path + ".coffee")
-        if application_js
-          prepend_to_file(application_js, data)
+        app_js_path = "app/assets/javascripts/application.js"
+        found_app_js = dest_file_exists?(app_js_path) || dest_file_exists?(app_js_path + ".coffee")
+        if found_app_js
+          prepend_to_file(found_app_js, data)
         else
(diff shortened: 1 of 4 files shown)


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -73,12 +73,12 @@ def update_application_js
...
-        application_js = dest_file_exists?(application_js_path) || dest_file_exists?(application_js_path + ".coffee")
-        if application_js
-          prepend_to_file(application_js, data)
+        app_js_path = "app/assets/javascripts/application.js"
+        found_app_js = dest_file_exists?(app_js_path) || dest_file_exists?(app_js_path + ".coffee")
+        if found_app_js
+          prepend_to_file(found_app_js, data)
         else
-          puts_setup_file_error("#{application_js} or #{application_js}.coffee", data)
+          create_file(app_js_path, data << "\n\n//= require turbolinks\n\n")
```
justin808: Why add turbolinks? It's not a requirement of react_on_rails. We support it, but don't necessarily encourage it.

I'd say remove this.

robwise: @justin808 roger that, makes sense

robwise:
@justin808 this should be good to go pending a final review

justin808:
@robwise This looks fine so long as it's well tested.

robwise:
@justin808 yep, the important part of the associated test is here: https://github.com/shakacode/react_on_rails/pull/108/files#diff-dd34ce0d9a02be7e957b081e5d11cb7cR146 since it doesn't create an existing application.js like it usually does, and yet the file still shows up as expected as running the generator.

PR #113, opened 2015-11-22
Use data attrs when rendering react_components instead of script tags

justin808:
See discussion at #102.

### DIFF
--- a/app/assets/javascripts/react_on_rails.js
+++ b/app/assets/javascripts/react_on_rails.js
@@ -101,7 +101,7 @@
   };
 
   function forEachComponent(fn) {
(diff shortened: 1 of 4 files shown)


justin808:
@ryanaip @alexfedoseev I'm going to go with Ryan's changes for now. There's too many concerns with the next version of TurboLinks to depend on TurboLinks eval'ing the script tag. I tested out the Rails 5 version and it's not there right now, and v3 will probably never be released.

I'm going to make the changes for react-router on this branch as well.

justin808:
@ryanaip Big thanks for this fix.

ryanaip:
Looks good — Glad to help out!

Issue #96, opened 2015-11-12
Bootstrap as an option


yorzi:
My existing Rails application is using Foundation instead of Bootstrap. I am wondering if Bootstrap can be an option in generator, if so, that'd be great.

justin808:
@yorzi We'd like to do that! Any chance you can help us with that?

yorzi:
@justin808 sure, I'd like to contribute something to this gem, but I am not sure where to start.. This might be a good one. :)

justin808:
@robwise and @jbhatab can give you some tips!

yorzi:
@justin808 ok, I will ping them in slack room if I encounter problems.

justin808:
@yorzi, actually @aaronvb would be a great guy to pair with on this one! @aaronvb did the amazing work of getting the generator code going.

robwise:
@yorzi the way I wrote the generators, everything related to Bootstrap is already isolated and put inside its own generator called `bootstrap_generator`. Everything is initiated from the `install_generator`, so follow the lead from the other generator options in that file and you should be able to just not called the bootstrap_generator with a conditional that checks the options, just like the other ones (you'll see if you look at the file).

yorzi:
@robwise let me try to work on that.

PR #119, opened 2015-11-25
Bootstrap as an option for generator

yorzi:
Really busy in whole last week, only start to have some time for #96  

It's not fully ready yet. But I hope to have this PR submitted for people to review and give me feedback, so that I am able to make it better. I still need to adjust/add some rspec testings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Please see [Getting Started](#getting-started) for how to set up your Rails proj
         + [Bootstrap via Rails Server](#bootstrap-via-rails-server)
         + [Bootstrap via Webpack HMR Dev Server](#bootstrap-via-webpack-hmr-dev-server)
         + [Keeping Custom Bootstrap Configurations Synced](#keeping-custom-bootstrap-configurations-synced)
+        + [Skip Bootstrap Integration](#skip-bootstrap-integration)
     - [Linters](#linters)
         + [JavaScript Linters](#javascript-linters)
         + [Ruby Linters](#ruby-linters)
@@ -213,6 +214,7 @@ Options:
   -j, [--skip-js-linters], [--no-skip-js-linters]      # Skip installing JavaScript linting files
   -L, [--ruby-linters], [--no-ruby-linters]            # Install ruby linting files, tasks, and configs
   -H, [--heroku-deployment], [--no-heroku-deployment]  # Install files necessary for deploying to Heroku
+  -b, [--skip-bootstrap], [--no-skip-bootstrap]        # Skip installing files for bootstrap support
 
 Runtime options:
   -f, [--force]                    # Overwrite files that already exist
@@ -262,6 +264,9 @@ When using the webpack dev server, which does not go through Rails, bootstrap is
 
 #### Keeping Custom Bootstrap Configurations Synced
 Because the webpack dev server and Rails each load Bootstrap via a different file (explained in the two sections immediately above), any changes to the way components are loaded in one file must also be made to the other file in order to keep styling consistent between the two. For example, if an import is excluded in `_bootstrap-custom.scss`, the same import should be excluded in `bootstrap-sass-config.js` so that styling in the Rails server and the webpack dev server will be the same.
+
+#### Skip Bootstrap Integration
+Bootstrap integration is enabled by default, but can be disabled by passing the `--skip-bootstrap` flag (alias `-b`). When you don't need Bootstrap in your existing project, just skip it as needed.
 
 ### Linters
 The React on Rails generator can add linters and their recommended accompanying configurations to your project. There are two classes of linters: ruby linters and JavaScript linters.
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -34,6 +34,13 @@ module ReactOnRails
                    desc: "Install ruby linting files, tasks, and configs",
                    aliases: "-L"
 
+      # --skip-bootstrap
+      class_option :skip_bootstrap,
+                   type: :boolean,
+                   default: false,
+                   desc: "Skip integrating Bootstrap and don't initialize files and regarding configs",
+                   aliases: "-b"
+
       def add_hello_world_route
         route "get 'hello_world', to: 'hello_world#index'"
       end
@@ -70,9 +77,6 @@ module ReactOnRails
           //= require generated/vendor-bundle
           //= require generated/app-bundle
 
-          // bootstrap-sprockets depends on generated/vendor-bundle for jQuery.
-          //= require bootstrap-sprockets
-
         DATA
 
         app_js_path = "app/assets/javascripts/application.js"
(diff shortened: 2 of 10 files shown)


justin808:
@robwise Please take a look at this and see if this should go into v1.1.0.

justin808:
@robwise @yorzi We're getting some build failures.

lib/generators/react_on_rails/install_generator.rb
```diff
@@ -34,6 +34,13 @@ class InstallGenerator < Rails::Generators::Base
                    desc: "Install files necessary for deploying to Heroku",
                    aliases: "-H"
 
+      # --bootstrap
+      class_option :bootstrap,
+                   type: :boolean,
+                   default: false,
```
robwise, line 40: I talked to @justin808 and he said he'd prefer if this would default to true

yorzi, line 40: @robwise let me make it default to `true`
@justin808 I will let this PR pass build today.

justin808:
@yorzi, let @robwise and I know when this one is ready.

yorzi:
@justin808 I think it's ready with correct functions and testings. The only concern for now is to document acordingly, but I am not sure where and how to do that correctly. Please take that document thing in mind before merging to your master.

justin808:
@robwise I'll let you take a first review at this, then I'll review.

justin808:
@robwise Let's try to get this wrapped up tomorrow. I'd like to push a version of the generator diffs (prs) to take a look at going basic -> basic + bootstrap.

README.md
```diff
@@ -128,7 +128,7 @@ In most cases, you should use the `prerender: false` (default behavior) with the
 
 Now the server will interpret your JavaScript using [ExecJS](https://github.com/rails/execjs) and pass the resulting HTML to the client. We recommend using [therubyracer](https://github.com/cowboyd/therubyracer) as ExecJS's runtime. The generator will automatically add it to your Gemfile for you.
 
-Note that **server-rendering requires globally exposing your components by setting them to `global`, not `window`** (as is the case with client-rendering). If using the generator, you can pass the `--server-rendering` option to configure your application for server-side rendering. 
+Note that **server-rendering requires globally exposing your components by setting them to `global`, not `window`** (as is the case with client-rendering). If using the generator, you can pass the `--server-rendering` option to configure your application for server-side rendering.
```
robwise, line 132: @justin808 @yorzi  this may be a little unclear, as it makes it sound as though we don't need to expose components to window when server rendering. We still need to do that, just _in addition_ we need to expose to global.

README.md
```diff
@@ -213,6 +213,7 @@ Options:
   -j, [--skip-js-linters], [--no-skip-js-linters]      # Skip installing JavaScript linting files
   -L, [--ruby-linters], [--no-ruby-linters]            # Install ruby linting files, tasks, and configs
   -H, [--heroku-deployment], [--no-heroku-deployment]  # Install files necessary for deploying to Heroku
+  -B, [--bootstrap], [--no-bootstrap]  # Install files necessary for bootstrap support.
```
robwise: @justin808 @yorzi to be consistent with what we've got so far, if the default is that the bootstrap code is generated, and the user would need to specifically pass an option to turn it off, then we should use the `--skip-bootstrap` option not just `bootstrap`, just like how we're using `--skip-js-linters` since by default they are on

lib/generators/react_on_rails/base_generator.rb
```diff
@@ -33,6 +33,13 @@ class BaseGenerator < Rails::Generators::Base # rubocop:disable Metrics/ClassLen
                    desc: "Install ruby linting files, tasks, and configs",
                    aliases: "-L"
 
+      # --bootstrap
+      class_option :bootstrap,
```
robwise: see comment above about `--skip-bootstrap`

robwise:
Just a couple of comments, the biggest thing being using `--skip-bootstrap` as the option instead of `--bootstrap` if by default the bootstrap generator will be run. This is a Rails convention that we were copying as it hints to the user which options are on by default and which are not. 

@yorzi All in all, a very good job—I'm delighted to see that you were able to read the code and change things exactly how they should have been, and without any help! I also liked how you refactored the bootstrap examples tests.

README.md
yorzi, line 132: @robwise I think this doc is not supposed to be updated in this branch. It might because my editor removes an space here.. anyway, can you or @justin808 do this doc updates accordingly?

README.md
yorzi: @robwise this is a good suggestion. Let me update this part. Thanks :)

README.md
justin808, line 132: @yorzi or @robwise we should do something to get rid of trailing spaces.

justin808:
@robwise when you're ready for me to do a final review **after** you have generated the PR of basic to basic + bootstrap, let me know.

robwise:
@justin808 I'm just confused as to what we are recommending the user should be doing when using stylesheets without bootstrap. Can they just use them the normal way you would for a Rails app where they go in `app/assets/stylesheets`, or alternatively, the normal way you would use webpack to require the stylesheets? Both? Are there any special caveats that they will need to watch out for or address?

yorzi:
@robwise good points! I've updated README to include some more explanation for bootstrap option. You or @justin808 can help to reword it with more details if needed.

justin808:
@yorzi Great job!

- [ ] rebase down to a nice commit with a nice message
- [ ] work with Rob to get yours here: https://github.com/shakacode/react_on_rails-generator-results Rob has a script that creates these. Once yours is up, I'll review the diffs.

yorzi:
@justin808 what do you mean by "rebase down to a nice commit with a nice message" here?

@robwise can you help with the generator results or would you let me know how to do that with your script?

justin808:
@yorzi

*  http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
* https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history
* https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

robwise:
@yorzi As for the generators, all you need to do is go to `rakelib/examples.yml` and follow the pattern there to generate an example app that passes `--skip-bootstrap`. If it turns out to fail the tests not because it isn't working but because something about the tests expects bootstrap, you can modify the `dev_tests` generator to do something special (such as put a different type of test in the `spec` folder) in the case that `--skip-bootstrap` is passed.

yorzi:
@justin808 oh, I do know the way to rebase and combine commits/logs, so you want me to combine these commits with better git logs, right?

justin808:
Please combine to one commit with nice description that I can use in the change log.

yorzi:
@robwise your guide is to test dummy app, but @justin808 asked for a generator results for diff from here: https://github.com/shakacode/react_on_rails-generator-results . Do I understand something incorrectly?

robwise:
@yorzi You're absolutely right, I misunderstood. I went ahead and took the liberty of modifying my little generator script to generate an app that uses your new option so that we can look at the diffs.

Check it out here:
https://github.com/shakacode/react_on_rails-generator-results-testing/pulls

robwise:
Also I configured rake to generate an example app for you in a branch based off of your changes: https://github.com/shakacode/react_on_rails/blob/bootstrap-as-an-option/rakelib/examples_config.yml#L18

It would probably be easiest if you just manually made this change in your own branch

yorzi:
@robwise :+1: and I added that configuration for sample app .
cc: @justin808

PR #122, opened 2015-11-25
Render server side every time in development environment

### LINKED ISSUE #114: Server Bundle Caching Problem (Again)
Justin and I noticed when pairing that the server bundle was not being reloaded when modified. He said he has noticed it at several other times as well. This was supposed to have been fixed, but apparently not.

spadin:
Is this sufficient to fix #114? It looks like the `ConnectionPool` is being set on startup at: https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/engine.rb#L4 so I don't think we have to account for the initial pool creation here.

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -134,7 +134,7 @@ module ReactOnRailsHelper
     return ["", ""] unless prerender(options)
 
     # Make sure that we use up-to-date server-bundle
-    ReactOnRails::ServerRenderingPool.reset_pool_if_server_bundle_was_modified
+    ReactOnRails::ServerRenderingPool.reset_pool if Rails.env.development?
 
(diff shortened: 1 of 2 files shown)


robwise:
@spadin We had added this code to address the issue that the server-bundle.js file was not being reloaded when webpack generated a new version in development. However, we are still experiencing this problem, and may just have it reset the pool every single request when in development without even bothering to check the file (which is what React-Rails is doing).

justin808:
@spadin Thanks! We're going to get a fix in later today. I'm going to work on this so we get v1.1.0 out later today or tomorrow.

spadin:
No problem. Just wanted to give this a shot. This change works for me in development, but I didn't confirm it works in production. Feel free to close this PR. You guys probably have a better solution since your more familiar with the code.

robwise:
See https://github.com/shakacode/react_on_rails/pull/125

PR #129, opened 2015-11-27
rewrite NODE_ENV as nodeEnv so the two mentions of the variable match up

marcellosachs:
when trying to run `foreman start -f Procfile.dev`, I was getting the error - 
`ReferenceError: NODE_ENV is not defined`

Then when I checked in webpack.server.rails.config.js I noticed that the variable `NODE_ENV` was indeed not defined, though a variable `nodeEnv` was. My guess is that `NODE_ENV` and `nodeEnv` are intended to mean the same thing, so I rewrote `NODE_ENV` to `nodeEnv` so they match up. After doing this, I no longer received the `ReferenceError` when running `foreman`.

### DIFF
--- a/lib/generators/react_on_rails/templates/base/base/client/webpack.client.base.config.js.tt
+++ b/lib/generators/react_on_rails/templates/base/base/client/webpack.client.base.config.js.tt
@@ -37,7 +37,7 @@ module.exports = {
 
     new webpack.DefinePlugin({
       'process.env': {
-        NODE_ENV: JSON.stringify(NODE_ENV),
+        NODE_ENV: JSON.stringify(nodeEnv),
       },
     }),
 
--- a/lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
+++ b/lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
@@ -8,7 +8,6 @@ const webpack = require('webpack');
 const config = require('./webpack.client.base.config');
 const devBuild = process.env.NODE_ENV !== 'production';
 
-
 config.output = {
   filename: '[name]-bundle.js',
   path: '../app/assets/javascripts/generated',
--- a/lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
+++ b/lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
@@ -6,6 +6,8 @@
 
 const webpack = require('webpack');
 const config = require('./webpack.client.base.config');
+const devBuild = process.env.NODE_ENV !== 'production';
+
 
 config.output = {
   filename: '[name]-bundle.js',


justin808:
https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/webpack.client.base.config.js#L6

```
const nodeEnv = devBuild ? 'development' : 'production';
```

@robwise we should generate the PRs for 1.1.0 to compare.

@marcellosachs the real bug is that we're missing a line in the config file.

Please see the ones for the https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/ for comparison.

Thanks!

lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
```diff
@@ -6,6 +6,8 @@
 
 const webpack = require('webpack');
 const config = require('./webpack.client.base.config');
+const devBuild = process.env.NODE_ENV !== 'production';
+
```
justin808, line 10:
``` js
const nodeEnv = devBuild ? 'development' : 'production';
```

justin808, line 10: one more line needed

marcellosachs:
@justin808 ok gotcha. Noticing now that my second commit also addresses something (missing definition of `devBuild`) that is not an issue in the react-webpack-rails-tutorial repo.

lib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js
marcellosachs, line 10: @justin808 ok I can add that in. It doesn't seem like `nodeEnv` is used in 
`ib/generators/react_on_rails/templates/base/base/client/webpack.client.rails.config.js`
although it is used in 
`lib/generators/react_on_rails/templates/base/base/client/webpack.client.base.config.js.tt`
(subject of my commit before this one)
Do you still want me to include that line?
Realizing now that perhaps I should have split these two things into two different PRs (?)

lib/generators/react_on_rails/templates/base/base/client/webpack.client.base.config.js.tt
```diff
@@ -37,7 +37,7 @@ module.exports = {
 
     new webpack.DefinePlugin({
       'process.env': {
-        NODE_ENV: JSON.stringify(NODE_ENV),
+        NODE_ENV: JSON.stringify(nodeEnv),
```
justin808, line 40: :+1:

justin808:
@marcellosachs @robwise Let's merge once this passes CI! @robwise let's get that auto testing of the generated apps in soon!

justin808:
Released 1.1.1 of gem for this fix. Thanks @marcellosachs.

justin808:
@marcellosachs I can send you an invite to our slack room if you send me an email.

marcellosachs:
@justin808 no problem! I just got on the slack channel yesterday, so good to go. Thanks for checking this out so quickly!

PR #137, opened 2015-12-01
Create examples tasks to test generated example apps

### LINKED ISSUE #130: Update generated package.json dependencies
The generator still uses outdated node dependencies

robwise:
**BIG THANK YOU to @dylangrafmyre for helping me debug Travis CI!!!**

- removes the templating of npm-shrinkwrap as this was causing issues with the generated apps due to outdated dependencies
- refactored the other rake tasks into their own `.rake` files inside of `rakelib`
- scripts for pushing to the generator-results repo are not implemented yet, but we shouldn't wait to merge this just for that
- after a long session with Dylan, we realized that the only way to get Travis to load all of the necessary gem dependencies was to have them in a single gemfile. It doesn't seem to want to use more than one for an entire build, so I moved all necessary dependencies to the gem's Gemfile. I'm assuming that these dependencies will not be required when users require our gem, right?


rakelib/examples.rake
```diff
@@ -0,0 +1,107 @@
...
+require_relative "task_helpers"
+
+include ReactOnRails::TaskHelpers
+
+namespace :examples do
+  # Loads data from examples_config.yml and instantiates corresponding ExampleType objects
+  examples_config_file = File.expand_path("../examples_config.yml", __FILE__)
+  examples_config = YAML.load(File.read(examples_config_file))
+  examples_config["example_type_data"].each do |example_type_data|
+    example_type_data.keys.each { |key| example_type_data[key.to_sym] = example_type_data.delete(key) }
```
robwise: this is an ugly way to coerce string keys into symbols which makes it easy to pass as a single arg

rakelib/examples.rake
```diff
@@ -0,0 +1,107 @@
...
+    ExampleType.new(example_type_data)
+  end
```
robwise: this whole block is ugly, really

rakelib/examples.rake
```diff
@@ -0,0 +1,107 @@
...
+
+  # Define tasks for each example type
+  ExampleType.all.each do |example_type|
+    # GENERATED FILES
+    example_type.generated_files.each do |f|
+      file f => example_type.source_files do
```
robwise, line 21: this is an example of a `file` task in Rake. We are using a FileList of all generated files for the given example type, and declaring them dependent on a FileList of all of the source files (in lib/generators/react_on_rails). This dependency means that if the dependee has a newer modification date than then dependent, then the block will be invoked. In this case, we invoke a task to re-generate the app. Note that rake keeps track of tasks being invoked, and therefore will not invoke the same task more than twice.

rakelib/examples_config.yml
```diff
@@ -0,0 +1,16 @@
+example_type_data:
```
robwise, line 1: I figured a yml file would be easiest for people to work with. If we want to add another example type, such as running the app with a new generator option, you just add the name and generator options and you're done. That's it.

rakelib/examples.rake
justin808: @robwise how about using http://api.rubyonrails.org/classes/Hash.html#method-i-deep_symbolize_keys

rakelib/example_type.rb
```diff
@@ -0,0 +1,160 @@
+require "rake"
```
justin808, line 1: Why is the file called example_type?

rakelib/examples_config.yml
```diff
@@ -0,0 +1,16 @@
...
+  -
+    name: basic
```
justin808, line 3: I hope we can use the generator_options in the pr description

justin808, line 3: do we want to configure the git remote in this yml file?

rakelib/run_rspec.rake
```diff
@@ -0,0 +1,65 @@
+require "coveralls/rake/task"
+require "pathname"
+
+require_relative "task_helpers"
+require_relative "example_type"
+
+include ReactOnRails::TaskHelpers
+
+namespace :run_rspec do
+  desc "Run RSpec for top level only"
```
justin808, line 10: Can we also lint the code, only the JS part? would that make sense? My concern is that we generate some JS that does not lint.

justin808:
Cursory review done. Looks amazing. @samnang please take a peek at this.

rakelib/examples.rake
robwise: I could do that I guess since we've got rails in the gem's gemfile anyway

rakelib/examples_config.yml
robwise, line 3: @justin808 I was thinking of keeping this file separate and then just making another yml file for the apps we generate for the diffs repo. These apps are different than those because they are going to have noise from the dev_tests generator and are within the gem's git repo.

rakelib/example_type.rb
robwise, line 1: @justin808 because it defines a class called `ExampleType`, which are objects that hold data about the types of examples we want to generate

rakelib/run_rspec.rake
robwise, line 10: @justin808 yes! Great idea!

rakelib/examples.rake
robwise: well actually this fails in the linter, and I think it would be weird to randomly have activesupport in our linters docker image. I think I'm just going to keep it as is.

rakelib/run_rspec.rake
robwise, line 10: @justin808 actually, now that I'm thinking about, I think this happens automatically since I never bothered to add the generated apps to the linters' ignore/exclude lists

robwise:
@yorzi I'm using a new syntax called "required keyword arguments" which is only in Ruby 2.1+

PR #142, opened 2015-12-02
Check for config/initializers/assets.rb file

buren:
First of all thanks for an awesome gem :star: 

react_on_rails:install fails when there is no existing assets initializer file:

```
$ rails generate react_on_rails:install
# ...
# => No such file or directory @ rb_sysopen /Users/xxx../config/initializers/assets.rb
```

I added a small patch that followed the same pattern used in [`BootstrapGenerator#prepend_to_application_scss `](https://github.com/shakacode/react_on_rails/blob/213fd8049840387f32bdb626e20fbde29a5acea5/lib/generators/react_on_rails/bootstrap_generator.rb#L82).

### DIFF
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -20,7 +20,12 @@ module ReactOnRails
 
           Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
         DATA
-        append_to_file("config/initializers/assets.rb", data)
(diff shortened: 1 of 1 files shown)


justin808:
@buren You closed the issue. Is it not an issue?

buren:
@justin808 i jumped the gun a little bit, I'll be back :)

buren:
@justin808 tried to re-open this PR, opened a new one instead. Apparently github doesn't allow you re-open PRs if there has been a force-push in between. See #143

PR #143, opened 2015-12-02
Check for config/initializers/assets.rb file

buren:
First of all thanks for an awesome gem :star: 

react_on_rails:install fails when there is no existing assets initializer file:

```
$ rails generate react_on_rails:install
# ...
# => No such file or directory @ rb_sysopen /Users/xxx../config/initializers/assets.rb
```

I added a small patch that followed the same pattern used in [`BootstrapGenerator#prepend_to_application_scss `](https://github.com/shakacode/react_on_rails/blob/213fd8049840387f32bdb626e20fbde29a5acea5/lib/generators/react_on_rails/bootstrap_generator.rb#L82).

### DIFF
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -24,7 +24,7 @@ module ReactOnRails
         if File.exist?(assets_intializer)
           append_to_file(assets_intializer, data)
         else
-          puts_setup_file_error(assets_intializer, data)
+          create_file(assets_intializer, data)
         end
       end
 
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -20,7 +20,12 @@ module ReactOnRails
 
           Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
         DATA
-        append_to_file("config/initializers/assets.rb", data)
+        assets_intializer = File.join(destination_root, "config/initializers/assets.rb")
+        if File.exist?(assets_intializer)
+          append_to_file(assets_intializer, data)
+        else
+          puts_setup_file_error(assets_intializer, data)
+        end
       end
 
       def copy_bootstrap_files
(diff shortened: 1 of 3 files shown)


robwise:
@buren Would it be possible to also get a test for this? It would basically be just like the application.scss file test here: https://github.com/shakacode/react_on_rails/pull/137

lib/generators/react_on_rails/bootstrap_generator.rb
```diff
@@ -20,7 +20,12 @@ def append_to_assets_initializer
 
           Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
         DATA
-        append_to_file("config/initializers/assets.rb", data)
+        assets_intializer = File.join(destination_root, "config/initializers/assets.rb")
+        if File.exist?(assets_intializer)
+          append_to_file(assets_intializer, data)
+        else
+          puts_setup_file_error(assets_intializer, data)
```
justin808: we should create the file if it doesn't exist, not error out

robwise: @justin808 actually that's a good point, @buren isn't that what it does for the appliaction.js file?

buren: :+1: I'll update this to use `create_file(assets_intializer, data)` and I'll also add some test.

buren: Thanks @robwise and @justin808 for the quick review! :rocket:

buren:
@robwise, @justin808 I replaced `puts_setup_file_error` with `create_file` for `config/initializers/assets.rb` and added test :)

Please let me know if there is anything I can do to improve this PR :+1: 

Thanks!

buren:
I also tested/verified this on an app that doesn't have `config/initializers/assets.rb` file.

justin808:
This looks excellent. Once @robwise has a chance to look at it, we'll merge!

justin808:
@buren can you please rebase your commits into one commit on top of master, and then I'll merge. Thanks!

buren:
@justin808 rebased and I also added another test for `with assets.rb file`, to follow the same pattern you use with `application.js`..

justin808:
Thanks @buren!

buren:
Thank you @justin808, @robwise and others for your hard work with this project! :+1:

Issue #140, opened 2015-12-02
running generator overwrites application.js incorrectly.


yorzi:
In my existing application.js, I have something like `//= require_tree ./lh/utility/`, when running generator, it seems all `//= require_tree .` will be removed and messy up all my existing required js, which should not be the original purpose.

justin808:
@yorzi I totally agree! thanks for finding this. We should

1. ONLY comment out, not remove any lines that we need to remove.
2. not touch require_tree for below the top level, as your example suggests.

yorzi:
@justin808 I will fix this later.

PR #144, opened 2015-12-03
Fix application.js incorrect overwritten issue

yorzi:
Just comment out incompatible requires. Refers to #140

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -87,7 +87,7 @@ module ReactOnRails
         application_js = File.join(destination_root, "app/assets/javascripts/application.js")
         gsub_file(application_js, "//= require jquery_ujs", "// require jquery_ujs")
         gsub_file(application_js, %r{//= require jquery$}, "// require jquery")
-        gsub_file(application_js, %r{//= require_tree .$}, "// require_tree .")
+        gsub_file(application_js, %r{//= require_tree \.$}, "// require_tree .")
       end
 
       def strip_application_js_of_double_blank_lines
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -85,9 +85,9 @@ module ReactOnRails
 
       def strip_application_js_of_incompatible_sprockets_statements
         application_js = File.join(destination_root, "app/assets/javascripts/application.js")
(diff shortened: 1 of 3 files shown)


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -85,9 +85,9 @@ def update_application_js
 
       def strip_application_js_of_incompatible_sprockets_statements
         application_js = File.join(destination_root, "app/assets/javascripts/application.js")
-        gsub_file(application_js, "//= require jquery_ujs", "")
-        gsub_file(application_js, %r{//= require jquery$}, "")
-        gsub_file(application_js, "//= require_tree .", "")
+        gsub_file(application_js, "//= require jquery_ujs", "// require jquery_ujs")
+        gsub_file(application_js, %r{//= require jquery$}, "// require jquery")
+        gsub_file(application_js, %r{//= require_tree .$}, "// require_tree .")
```
robwise: I believe the period needs to be escaped, right?

spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -143,7 +143,7 @@
   end
 
   context "without existing application.js or application.js.coffee file" do
-    before(:all) { run_generator_test_with_args([], application_js: false) }
+    before(:all) { run_generator_test_with_args([], application_js: true) }
```
robwise: we still want to test for it creating an application_js file if one is missing, so we don't want to set this to true for this test, right?

spec/react_on_rails/support/shared_examples/base_generator_examples.rb
```diff
@@ -48,12 +48,15 @@
     end
   end
 
-  it "removes incompatible sprockets require statements" do
+  it "comments out incompatible sprockets require statements" do
     assert_file("app/assets/javascripts/application.js") do |contents|
-      refute_match("//= require_tree .", contents)
+      refute_match(%r{//= require_tree .$}, contents)
+      assert_match(%r{// require_tree .$}, contents)
```
robwise: see other comment about escaping the period

lib/generators/react_on_rails/base_generator.rb
yorzi: Yeah, you right... But it covers the case correctly here when there is a `$` in the ending. Anyway, it's better to code with preciseness. I will fix it.

justin808:
Thanks @yorzi!

Issue #128, opened 2015-11-27
Create NPM module with the API (v2.0.0)


justin808:
I had a detailed discussion with @alexfedoseev.

ASAP, I'm going to do the following:

Introduce this API via a node module, react_on_rails:

```js
import ReactOnRails from 'react_on_rails';
ReactOnRails.register('HelloWorldApp', HelloWorldAppServer);
```

This will work for both `serverGlobals.jsx` and `clientGlobals.jsx`.

We'll put the registered components on `ReactOnRails.components['componentName']` rather than polluting the global namespace.

- [x] Add api `ReactOnRails.register(componentName, component, generator_function = false)`. We can set the fact a component is really a generator_function with the 3rd argument.
- [ ] We'll also make the other ReactOnRails API's available via the node module.
- [x] We'll use the same repo as the gem as we want a single release script that releases both the gem and
- [x] startup will check that /client/package.json contains react_on_rails, and that the version number is the same as the gem, and quit if it is not.
- [x] convert react_on_rails.js to use babel.
- [x] automatically export ReactOnRails to the global namespace (no explicit export in webpack needed). The entry point of ReactOnRails will be required for the client and server bundles.
- [x] OPTIONAL -- for backwards compatibility, do we check if the component is declared on the global namespace? Probably.

CC: @alexfedoseev, @dylangrafmyre, @robwise  

-----

Example places this will be used:

https://github.com/shakacode/react_on_rails-generator-results-v1.0.3/pull/111/files#diff-3f5b3fdbb5756c6fb057f19689e92a1aR2

https://github.com/shakacode/react_on_rails-generator-results-v1.0.3/pull/112/files#diff-d012fe328a02dd847c8241e3d58d6a2aR3

```js
import ReactOnRails from 'react_on_rails';
import HelloWorldAppServer from './HelloWorldAppServer';
ReactOnRails.register('HelloWorldApp', HelloWorldAppServer);
```

justin808:
https://github.com/shakacode/react_on_rails-generator-results-v1.0.3/pull/111/files#diff-3f5b3fdbb5756c6fb057f19689e92a1aR2

justin808:
@alexfedoseev, @jbhatab, @robwise, anybody: Please review this list. What do you think about allowing the specification of the generator_function in the JS code:

- [ ] Add api `ReactOnRails.register(componentName, component, generator_function = false)`. We can set the fact a component is really a generator_function with the 3rd argument.

[...]

Issue #92, opened 2015-11-06
`examples` folder


robwise:
The idea is to have a Rakefile script that runs through the generator with differing options, putting each result into an `examples` folder. We would then implement a hidden `dev_tests` (or whatever we want to call it) generator that the script makes sure to always pass. A hidden generator is one that is not visible to the public but is nonetheless able to be invoked if you know the command.

This generator will create general sanity tests to ensure various essential functionalities of each result application is working correctly. We could ensure that our CI environment re-generates these results and runs these tests, thereby preventing a scenario where we pull something that breaks the generator apps. 

Whether or not we include the `examples` directory in source control I guess is optional, but I'd be worried that they would sometimes be stale since they would rely on the user making sure to re-generate of their changes.

- [x] create Rakefile script to generate apps into `examples` folder
- [x] ensure that Coveralls, the various linters (actually, we could keep the JS linters), are set up to ignore it
- [x] create hidden `dev_test_generator`
- [x] have generator results script use this option when generating the apps
- [x] have Rakefile include generated app tests when running test suite
- [ ] *move spec/dummy tests to the `dev_test_generator` and delete dummy apps (optional because this wouldn't work as far as ensuring React v13 compatibility)*
- [x] ensure our CI environment handles all of this appropriately
- [x] refactor so that we do not unnecessarily rebuild the app, re-run bundle install, or re-run npm install when we don't need to do so
- [ ] incorporate the ability to generate the apps in a separate location and push them to a given GitHub (basically the generator results functionality)

justin808:
@robwise No need to make it hidden probably. Useful to show what happens.

justin808:
@robwise should we close this?

robwise:
@justin808 this has been on hold while I've been working on other things, but it's actually just about ready to go after I make some performance optimizations. Should be really cool! This is what allows us to  have CI ensure the generator apps actually work!

robwise:
You can preview the branch here: https://github.com/shakacode/react_on_rails/tree/examples-folder

robwise:
Closing for now

PR #148, opened 2015-12-07
WIP -- conversion to ES6, npm for react_on_rails

justin808:
2.0.0-rc.x on NPM and 2.0.0.rc.x on rubygems!

`ReactOnRails.register({react1, react2});` to integrate with Rails! No more react_on_rails.js asset!

* https://www.npmjs.com/package/react-on-rails
* https://rubygems.org/gems/react_on_rails/versions/2.0.0.beta.2

-----------

## Here's the pending checklist:

- [x] Fix Redux (no server rendering) generator
- [x] Fix example apps to be symlinked to node package via the refresh script and rename to symlink-node-package
- [x] ensure symlink-node-package doesn't blow up if examples folder is not present
- [x] Ensure CI works (@dylangrafmyre)
- [x] Implement logic to check node package version against ruby gem version and warn if different
- [x] add JS tests and integrate into CI
- [x] have local runs of tests use poltergeist, not selenium 
- [x] change generators to match up with what's in spec/dummy
- [x] update docs
- [x] Ensure we are linting example apps' JS

See #128


justin808:
@yorzi Please take a look. I could not get this to start yet. Maybe you'll see the issue.

@alexfedoseev can you please take a look as well, both the setup and the ES6 conversion.

justin808:
@alexfedoseev Please review.
@yorzi Please see my last commit message and pick up from there.

yorzi:
@justin808 :+1: awesome work!

justin808:
@alexfedoseev the node module name of react_on_rails might be awkward. Should it be ReactOnRails? or `react-on-rails`?

```js
import ReactOnRails from 'react_on_rails';
```

alexfedoseev:
`react-on-rails` is more common in npm.

node_package/scripts/release
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+npm version $1 -m "Bump version to %s"
+git push -u origin master
```
alex35mil, line 4: I changed it to just `git push`, b/c betas can be released from branches.

node_package/src/react_on_rails.js
```diff
@@ -0,0 +1,277 @@
...
+import React from 'react';
+
+const context = ((typeof window !== 'undefined') && window) ||
+  ((typeof global !== 'undefined') && global) || this;
+
+let ReactOnRails;
+
+const turbolinksInstalled = (typeof Turbolinks !== 'undefined');
+
+function provideClientReact() {
```
alex35mil: Maybe move it to small modules/utils like in `bootstrap-loader`?

node_package/src/react_on_rails.js
```diff
@@ -0,0 +1,277 @@
...
+  if (typeof context.ReactDOM === 'undefined') {
+    if (React.version >= '0.14') {
```
alex35mil: It's better to change string to number (`0.14`) to avoid conversion.

alexfedoseev:
@justin808 LGTM! Few comments.

rstudner:
Looks good!

justin808:
@yorzi Check out @alexfedoseev suggestions. I'm OK if you apply them. 

Also, take a look at: https://github.com/umdjs/umd

We need to be sure we're properly installing the event handlers only ONCE. Probably can use some debug code until we're sure.

Let me know if you have questions.

justin808:
@yorzi Here's the list to do right now: https://github.com/shakacode/react_on_rails/commit/0ae1e8d5709391b008f14dba547eb62bd4b18522

 Not yet testing the registerComponent API
- [ ] Provides warning messages if not registering

TODO:
- [ ] Change examples to use

import ReactOnRails from 'react_on_rails';
ReactOnRails.registerComponent(name, component);

- [ ] Look into moving the 'is generator function' part from the view helper
  to the JS definition
- [ ] Add JS tests with Tape
- [ ] Fix generators
- [ ] Add scenario for not registering and be sure that we still use the
  globally defined one.

alexfedoseev:
@yorzi Also if you'll decide to split sources of this module into few smaller sub-modules, webpack has build-in functionality to wrap build into UMD wrapper ([link](https://webpack.github.io/docs/configuration.html#output-librarytarget). Ping me if you have any questions.

yorzi:
@alexfedoseev I had same idea to split current js into several small ones. Great hints!

node_package/src/react_on_rails.js
yorzi: @alexfedoseev assume `React.version == '1.4.3'`, it seems (`React.version >= 0.14`) will return `false`, but (`React.version >= '0.14'`) will return `true`.

node_package/src/react_on_rails.js
```diff
@@ -0,0 +1,277 @@
...
+
+function reactOnRailsPageLoaded() {
+  forEachComponent(render);
+}
+
+function reactOnRailsPageUnloaded() {
+  forEachComponent(unmount);
+}
+
+const components = {};
```
yorzi: I can not understand this variable here. If we attach components to this `components` const, how should it work? (comparing to previous window.component_name or global.component_name) cc @justin808 @alexfedoseev

spec/dummy/client/app/startup/clientGlobals.jsx
```diff
@@ -6,9 +8,16 @@ import HelloWorldApp from './HelloWorldApp';
...
 // This is an example of how to render a React component directly, without using Redux
-window.HelloWorld = HelloWorld;
+// window.HelloWorld = HelloWorld;
 window.HelloWorldWithLogAndThrow = HelloWorldWithLogAndThrow;
 window.HelloWorldES5 = HelloWorldES5;
 window.ReduxApp = ReduxApp;
 window.HelloWorldApp = HelloWorldApp;
 window.RouterApp = ClientRouterApp;
+
+ReactOnRails.registerComponent('HelloWorld', HelloWorld);
```
yorzi: It's not working as expected. Debugged from this call, it shows we defined `registerComponent` incorrectly.

node_package/tests/react_on_rails.spec.js
```diff
@@ -0,0 +1,12 @@
+var test = require('tape');
```
yorzi: Never wrote JS testing before. Is that possible to use ES6 syntax for testing code? cc: @justin808

node_package/src/react_on_rails.js
```diff
@@ -0,0 +1,277 @@
...
+
+ReactOnRails = {
```
yorzi: Should we do something like `export default ReactOnRails;` for other file's `import`? Hmm, but it doesn't work when I added it.

node_package/src/react_on_rails.js
alex35mil: @yorzi Reasonable :+1:

node_package/tests/react_on_rails.spec.js
alex35mil: Check out this repo for the start: https://github.com/shakacode/js-promises-testing

node_package/src/react_on_rails.js
alex35mil: @yorzi It's not necessary. By invoking this module (in entry point of webpack config for example), you call function on line 277, which makes things happen.

node_package/src/react_on_rails.js
justin808: We'll remove support for 0.13 in any case.

Since we're deprecating the old API, we'll be calling this 2.0.0, and it makes sense to remove 0.13 support at that point.

lib/generators/react_on_rails/templates/base/base/client/app/bundles/HelloWorld/startup/globals.jsx.tt
```diff
@@ -1,5 +1,6 @@
+import ReactOnRails from 'react_on_rails';
```
justin808: We might change this to:

``` js
import ReactOnRails from 'react-on-rails';
```

yorzi:
There are some errors which might relate to `babel-loader` when using `react-on-rails` in dummy app. Log here: https://gist.github.com/yorzi/6bec9e9d757afca4c582 cc: @justin808

alexfedoseev:
@yorzi Try to update to Babel 6 (https://github.com/shakacode/react-webpack-rails-tutorial/commit/d67548b473708300364ddb9fa1ec227b9dce00e4)

robwise:
@yorzi It may just be easier to move the tests from the dummy app to the `dev_tests` generator so they will get run on all of the example apps. You will have to refactor some of them since some are server-rendering specific, so the dev_tests generator will not copy those over for server-rendering examples. Then we could just get rid of the dummy apps. That way we don't have this problem in the future, because they get stale very easily. This was one of the todo items in https://github.com/shakacode/react_on_rails/issues/92

justin808:
@alexfedoseev @yorzi @robwise I just got it working with NPM! Tests will probably not pass remotely unless I push a version to npm. So I'll push 0.0.1 of react-on-rails.

justin808:
@alexfedoseev Any suggestions?

Here's the pending checklist:

- [ ] Allow specifying {generator: true} in `register` call.
- [ ] Check version number of package.json for react-on-rails and be sure it matches the gem version, quit otherwise.
- [ ] add JS tests and integrate into CI
- [ ] change generators to match up with what's in spec/dummy



 `register` or `registerComponent`?

```js
ReactOnRails.registerComponent('HelloWorld', HelloWorld, { generator: true });
ReactOnRails.registerComponent('HelloWorldWithLogAndThrow', HelloWorldWithLogAndThrow);
```

or 

```js
ReactOnRails.register('HelloWorld', HelloWorld, { generator: true });
ReactOnRails.register('HelloWorldWithLogAndThrow', HelloWorldWithLogAndThrow);
```

lib/generators/react_on_rails/templates/base/base/client/app/bundles/HelloWorld/startup/globals.jsx.tt
robwise: +1 for the hyphens

node_package/src/ReactOnRails.js
```diff
@@ -0,0 +1,35 @@
+import clientStartup from './clientStartup';
+import handleError from './handleError';
+import ComponentStore from './ComponentStore';
+import serverRenderReactComponent from './serverRenderReactComponent';
+import buildConsoleReplay from './buildConsoleReplay';
+
+const context = ((typeof window !== 'undefined') && window) ||
+  ((typeof global !== 'undefined') && global) || this;
```
robwise: could we extract a variable or two just for readability; `context` is a boolean?

node_package/src/buildConsoleReplay.js
```diff
@@ -0,0 +1,23 @@
...
+export default function buildConsoleReplay() {
+  let consoleReplay = '';
+
+  const history = console.history;
+
+  if (history && history.length > 0) {
+    history.forEach(msg => {
+      const stringifiedList = msg.arguments.map(arg => {
+        try {
+          return (typeof arg === 'string' || arg instanceof String) ? arg : JSON.stringify(arg);
```
robwise: is this ternary for performance?  Couldn't we just run arg through `stringify` no matter what, or does that not work?

node_package/src/clientStartup.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+function render(el) {
+  const componentName = el.getAttribute('data-component-name');
+  const domId = el.getAttribute('data-dom-id');
+  const props = JSON.parse(el.getAttribute('data-props'));
+  const trace = JSON.parse(el.getAttribute('data-trace'));
+  const generatorFunction = JSON.parse(el.getAttribute('data-generator-function'));
+  const expectTurboLinks = JSON.parse(el.getAttribute('data-expect-turbo-links'));
```
robwise, line 26: who is putting this attribute here, Rails or react_on_rails? Either way, checking for this is a cool idea. Belt and suspenders!

node_package/src/clientStartup.js
```diff
@@ -0,0 +1,84 @@
...
+
+  if (!turbolinksInstalled && expectTurboLinks) {
+    console.warn('WARNING: NO TurboLinks detected in JS, but it is in your Gemfile');
+  }
+
+  try {
+    const domNode = document.getElementById(domId);
+    if (domNode) {
+      const reactElementOrRouterResult = createReactElement({ componentName, props,
```
robwise: this is a stupid style thing, but just below on 42 when your object literal is multiple lines, you put each property on its own line, but here you don't. I prefer the each on its on line once wrapping is started, but obviously this is a very minor thing.

node_package/src/clientStartup.js
```diff
@@ -0,0 +1,84 @@
...
+        domId, trace, generatorFunction });
+      if (isRouterResult(reactElementOrRouterResult)) {
+        throw new Error('You returned a server side type of react-router error: ' +
+          JSON.stringify(reactElementOrRouterResult) +
+          '\nYou should return a React.Component always for the client side entry point.');
```
robwise, line 45: This part is a bit confusing for me, we are expecting a React.Component, but we check for a router result, and if it's a router result, we assume it's a server-side router error?

node_package/src/clientStartup.js
```diff
@@ -0,0 +1,84 @@
+import ReactDOM from 'react-dom';
+import handleError from './handleError';
+import createReactElement from './createReactElement';
+import isRouterResult from './isRouterResult';
+
+const REACT_ON_RAILS_COMPONENT_CLASS_NAME = 'js-react-on-rails-component';
+const turbolinksInstalled = (typeof Turbolinks !== 'undefined');
```
robwise, line 8: where is `Turbolinks` coming from, global? This passes the linters?

node_package/src/createReactElement.js
```diff
@@ -0,0 +1,21 @@
...
+import ReactOnRails from './ReactOnRails';
+
+export default function createReactElementOrRouterResult(
+  { componentName,
+    props,
+    domId,
+    trace,
+    generatorFunction,
+    location }) {
+  if (trace) {
```
robwise, line 22: Another stupid style comment, I think Alex usually does it something like this (and I can't remember if you can put trailing commas here, but I'm thinking you can): 

```
export default function createReactElementOrRouterResult({
  componentName,
  props,
  domId,
  trace,
  generatorFunction,
  location
}) {
  if (trace) {
```

spec/dummy/client/webpack.client.js
```diff
@@ -2,8 +2,9 @@ const path = require('path');
 
 module.exports = {
   entry: [
+    'es5-shim/es5-shim',
+    'es5-shim/es5-sham',
```
robwise: lines 5 and 6 are missing from spec/dummy/client/package.json

rakelib/run_rspec.rake
```diff
@@ -17,11 +17,6 @@ namespace :run_rspec do
     run_tests_in("spec/dummy", env_vars: "DRIVER=selenium_firefox")
```
robwise, line 16: heads up we're overriding the env var here, we should just delete this and rely on the rails_helper?

node_package/src/ReactOnRails.js
justin808: context is either `window` or `global`.

node_package/src/buildConsoleReplay.js
justin808: This is for safely printing args during server rendering.

justin808:
```
JSON.stringify("abc")
""abc""
```

node_package/src/clientStartup.js
justin808, line 8: Global from rails. Ask the rails guys if they use linters.

node_package/src/clientStartup.js
justin808, line 26: react_on_rails

node_package/src/clientStartup.js
justin808, line 45: line 33 checks

justin808:
@alexfedoseev @yorzi @mapreal19 @josiasds @jbhatab Take a look at this, what Rob put together.

http://codepen.io/robwise/pen/wMwzGo

I think this is good enough to remove the need to specify generator function.

```js
// SECRET SAUCE
function typeChecker(func) {
  const prototypeKeys = String(Object.keys(func.prototype));
  if (prototypeKeys.startsWith("constructor")) {
    return "es5Component";
  } else if (prototypeKeys.includes("render")) {
    return "es6Component";
  } else if (prototypeKeys === "") {
    return "generatorFunction";
  } else {
    return "Error";
  }
}

// SETUP
const es5Component = React.createClass({
  render: function() {
    return (<div>ES5 React Component</div>);
  }
});

const es6Component = class ES6Component extends React.Component {
  render() {
    return (<div>ES6 Component</div>);
  }
};

const foobarComponent = React.createClass({
  render: () => {
    return (<div>Component for Generator Function</div>);
  }
});

let generatorFunction = () => {
  return (<foobarComponent/>);
};
```

justin808:
I think we're pretty close.

I like this syntax for component registration:

```js
import ReactOnRails from 'react-on-rails';

import HelloWorld from '../components/HelloWorld';
import HelloWorldWithLogAndThrow from '../components/HelloWorldWithLogAndThrow';
import HelloWorldES5 from '../components/HelloWorldES5';
import HelloWorldApp from './HelloWorldApp';

import ReduxApp from './ClientReduxApp';
import RouterApp from './ClientRouterApp';
import HelloString from '../non_react/HelloString';

ReactOnRails.register({
  HelloWorld,
  HelloWorldWithLogAndThrow,
  HelloWorldES5,
  ReduxApp,
  HelloWorldApp,
  RouterApp,
});
```

justin808:
@yorzi @robwise @alexfedoseev I got the auto-detection working. Probably a bit more, but this should be close!

Big puzzle is how I can run the JS tests without installing react react-dom at the top level, as that will conflict with the inclusion of react in spec/dummy.

robwise:
TODO:

- [ ] Fix Redux (no server rendering) generator
- [ ] Fix example apps to be symlinked to node package so you don't have to run refresh script (then delete the refresh script)
- [ ] Ensure CI works
- [ ] Implement logic to check node package version against ruby gem version and warn if different

robwise:
@justin808 Hey I don't think we need to add a `ci` task, everything should already be getting done via the `rake` task. This way what you run locally with just `rake` and what CI runs is the same thing.

thewoolleyman:
Yeah, +1 to the default rake task doing everything CI does, it's what I would expect in a library

justin808:
@robwise I think the default should do "ci". Otherwise, how does one know what the description of the default is?

robwise:
@justin808 https://github.com/shakacode/react_on_rails/blob/master/Rakefile

robwise:
@justin808 finally got it passing in CI. Currently we are using selenium firefox as the driver for spec/dummy and poltergeist for the generated example apps. We could:

A) Keep it as is
B) have @dylangrafmyre yak shave on getting spec/dummy to work with poltergeist
C) change generator apps to also use selenium firefox just for consistency's sake

justin808:
Selenium is firefox is fine for generated apps on CI.

justin808:
@robwise I've released rc.1 and I'm done with the tests. I think any remaining JS tests are largely covered by our integration tests.

I'd say if no serious issues are uncovered this week, we ship later in the week!

justin808:
@robwise @alexfedoseev How about we change the official linter setup to use: https://github.com/shakacode/style-guide-javascript/tree/master/packages/eslint-config-shakacode

alexfedoseev:
@justin808 That's why we've made it :)

robwise:
@justin808 @alexfedoseev Yep, definitely! And we need to test if the example apps are actually getting linted once and for all.

bensie:
Looks like RC2 was tagged on GitHub, but not released to Rubygems or NPM. Can you do this or should I just wait for RC3 / final release?

justin808:
@bensie Give me a couple hours. I'll push up RC3!

justin808:
@bensie Please see 6bfc6ae or http://forum.shakacode.com/t/using-react-on-rails/451. Feedback appreciated.

bensie:
@justin808 I'll give it a go shortly...thanks!

justin808:
I'm going to squash commits and merge to master tomorrow!

Issue #138, opened 2015-12-01
Generator Warning Messages


robwise:
Currently the generator warning messages happen at the beginning of the output when running the generator and then quickly get pushed off screen by the noise the generator creates from making all of the files and such. We should probably move them to the end or something.

aaronvb:
@robwise I did some poking around last night and was able to get the InstallGenerator warning messages to output at the end by adding an error attribute to the InstallGenerator class. However, I'm still thinking of a way to get it to work with the rest of the generators.

The problem I see with the Generator classes is that they output top to bottom, and in the order we invoke them in the InstallGenerator. By using the same method above in each Generator, individual errors will output after each invoke, which is still not at the end.

Do you have any ideas?

robwise:
@aaronvb Yeah I was thinking about this problem too, certainly a monkey-patch option would be to make some sort of class in the `task_helper.rb` file and just add any error messages to an array that the `InstallGenerator` prints out at the very end.

justin808:
How about if we keep some variable to to keep an array of string messages to print at the end?

aaronvb:
@justin808 @robwise That's what I was thinking also. Will post an update soon.

justin808:
@aaronvb Any update on this issue? Thanks so much for your help!

justin808:
@aaronvb @robwise Is this ready to go?

aaronvb:
@justin808 @robwise I'm making the changes rob suggested right now. Will be pushing it soon.

aaronvb:
Let's close this issue.

justin808:
@aaronvb Thanks for your help with this one!

PR #158, opened 2015-12-11
Fixes Generator Warning Messages #138

aaronvb:
Added an ‘errors’ attribute to the install generator class which gets passed, as an option, to the generators that output errors.

Errors outputted get pushed into the errors attr array. At the end of the install any errors in the attr array get printed to the terminal.

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -1,7 +1,9 @@
 require "rails/generators"
 require File.expand_path("../generator_helper", __FILE__)
+require File.expand_path("../generator_errors", __FILE__)
 
 include GeneratorHelper
+include GeneratorErrors
 
 module ReactOnRails
   module Generators
@@ -62,7 +64,11 @@ module ReactOnRails
           /app/assets/javascripts/generated/*
         DATA
 
-        dest_file_exists?(".gitignore") ? append_to_file(".gitignore", data) : puts_setup_file_error(".gitignore", data)
+        if dest_file_exists?(".gitignore")
+          append_to_file(".gitignore", data)
+        else
+          GeneratorErrors.add_error(return_setup_file_error(".gitignore", data))
+        end
       end
 
       def update_application_js
--- a/lib/generators/react_on_rails/generator_helper.rb
+++ b/lib/generators/react_on_rails/generator_helper.rb
@@ -17,6 +17,14 @@ module GeneratorHelper
     puts "\n#{data}\n"
   end
 
+  def return_setup_file_error(file, data)
+    error = ""
+    error << "** #{file} was not found.\n"
+    error << "Please add the following content to your #{file} file:\n"
+    error << "\n#{data}\n"
+    error
+  end
+
   def empty_directory_with_keep_file(destination, config = {})
     empty_directory(destination, config)
     keep_file(destination)
--- a/lib/generators/react_on_rails/install_generator.rb
+++ b/lib/generators/react_on_rails/install_generator.rb
@@ -1,4 +1,8 @@
 require "rails/generators"
+require File.expand_path("../generator_helper", __FILE__)
+require File.expand_path("../generator_errors", __FILE__)
+include GeneratorHelper
+include GeneratorErrors
 
 module ReactOnRails
   module Generators
@@ -44,12 +48,20 @@ module ReactOnRails
                    aliases: "-b"
 
       def run_generators # rubocop:disable Metrics/CyclomaticComplexity
-        return unless installation_prerequisites_met?
+        unless installation_prerequisites_met?
+          print_errors
+          return
+        end
         warn_if_nvm_is_not_installed
         invoke_generators
+        print_errors
       end
 
       private
+
+      def print_errors
+        GeneratorErrors.errors.each { |errors| puts errors }
+      end
 
       def invoke_generators # rubocop:disable Metrics/CyclomaticComplexity
         invoke "react_on_rails:base"
@@ -72,14 +84,14 @@ module ReactOnRails
         return false unless `which npm`.blank?
         error = "** npm is required. Please install it before continuing."
         error << "https://www.npmjs.com/"
-        puts error
+        GeneratorErrors.add_error(error)
       end
 
       def missing_node?
         return false unless `which node`.blank?
         error = "** nodejs is required. Please install it before continuing."
         error << "https://nodejs.org/en/"
-        puts error
+        GeneratorErrors.add_error(error)
       end
 
       def uncommitted_changes?
@@ -87,12 +99,12 @@ module ReactOnRails
         status = `git status`
         return false if status.include?("nothing to commit, working directory clean")
         error = "** You have uncommitted code. Please commit or stash your changes before continuing"
-        puts error
+        GeneratorErrors.add_error(error)
       end
 
       def warn_if_nvm_is_not_installed
         return true unless `which nvm`.blank?
-        puts "** nvm is advised. Please consider installing it. https://github.com/creationix/nvm"
+        GeneratorErrors.add_error("** nvm is advised. Please consider installing it. https://github.com/creationix/nvm")
       end
     end
   end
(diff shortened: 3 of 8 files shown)


lib/generators/react_on_rails/generator_helper.rb
```diff
@@ -17,6 +17,16 @@ def puts_setup_file_error(file, data)
     puts "\n#{data}\n"
   end
 
+  def return_setup_file_error(file, data)
+    data = <<-DATA.strip_heredoc
```
aaronvb: This output creates weird spacing in my terminal. Any suggestions to fix this?

```
      ** .gitignore was not found.
      Please add the following content to your .gitignore file:

      # React on Rails
npm-debug.log
node_modules

# Generated js bundles
/app/assets/javascripts/generated/*
```

aaronvb:
Looks like I have some rubocop errors to fix.

lib/generators/react_on_rails/base_generator.rb
```diff
@@ -62,7 +62,7 @@ def update_git_ignore
           /app/assets/javascripts/generated/*
         DATA
 
-        dest_file_exists?(".gitignore") ? append_to_file(".gitignore", data) : puts_setup_file_error(".gitignore", data)
+        dest_file_exists?(".gitignore") ? append_to_file(".gitignore", data) : options.errors.push(return_setup_file_error(".gitignore", data))
```
justin808: It seems semantically odd to use `options` as the place to attach errors. Is this what any other gems use? Maybe a class variable would be better.

@samnang What's your opinion?

lib/generators/react_on_rails/generator_helper.rb
justin808: @aaronvb We ran into this a lot. I prefer to just skip the `strip_heredoc` and remove the leading spaces.

lib/generators/react_on_rails/base_generator.rb
aaronvb: @justin808 The problem is rails generators don't act like normal ruby classes. They get invoked, and each public method gets called from top to bottom. 

By using options, I can pass the array object from the install generator to any of the invoked generators and have them add errors.

Also, I couldn't find any information or examples on this issue.

justin808: How about if we create a singleton of our own class to that supports an api to add errors and later can be called to get the errors.

Putting errors on the options is really hacky.

aaronvb: @justin808 I agree, it does seem hacky. I'll explore your suggestion. Thanks for the feedback!

samnang: I agree with @justin808, we could create our own class. Either singleton or use one global object should be fine.

robwise: Please put it an array inside a class in the generators_helper.rb file.

justin808:
@aaronvb Any update?

aaronvb:
Updated with a singleton class.

lib/generators/react_on_rails/generator_helper.rb
robwise: @justin808 @aaronvb This code is a little confusing to me because you're assigning the parameter variable `data` to a variable called `data` that is set to a string that interpolates `data`, and then you add a call to `data` at the end, but I think this returns the same thing that would have been returned had you left it out. 

Anyway, the reason this indentation is happening is because `strip_heredoc` only strips to the leftmost non-space character found out of all the lines, and since you are including `#{data}` which has lines in it that are completely left-aligned, then strip-heredoc can't really do anything.

Example: 

```
 <<-TEXT.strip_heredoc
                  some text
                    some more text
TEXT 

# ouputs:
some text
   some more text
```

```
 <<-TEXT.strip_heredoc
                  some more text
                    some more text also
some text
TEXT 

# outputs:
                  some more text
                    some more text also
some text
```

aaronvb:
Updated with tests. Still not sure what to do about the output formatting.

lib/generators/react_on_rails/generator_helper.rb
```diff
@@ -17,6 +17,14 @@ def puts_setup_file_error(file, data)
...
   end
 
+  def return_setup_file_error(file, data)
+    error = ""
+    error << "** #{file} was not found.\n"
+    error << "Please add the following content to your #{file} file:\n"
+    error << "\n#{data}\n"
+    error
+  end
+
```
aaronvb, line 27: Renamed the string to 'error' for clarity. 

Also removed the strip_heredoc method per @robwise comment on data already being left justified causing it to not work here.

Output looks like this now:

```
** .gitignore was not found.
Please add the following content to your .gitignore file:

# React on Rails
npm-debug.log
node_modules

# Generated js bundles
/app/assets/javascripts/generated/*
```

aaronvb:
Removing WIP in title.

spec/react_on_rails/support/generator_spec_helper.rb
```diff
@@ -8,7 +8,7 @@
 # Expects an array of strings, such as "--redux"
 def run_generator_test_with_args(args, options = {})
   prepare_destination # this completely wipes the `destination` directory
-  simulate_existing_file(".gitignore")
+  simulate_existing_file(".gitignore") unless options.key?(:has_errors)
```
robwise: I would use `fetch(:has_errors, false)` here, as this would be the wrong behavior if I passed `has_errors: false`.

Also, I wouldn't call it `has_errors`, I would just do it like: `gitignore: false` as this is easier to reason about

spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -190,4 +190,9 @@
     before(:all) { run_generator_test_with_args([], assets_rb: true) }
     include_examples "base_generator:base", assets_rb: true
   end
+
+  context "with missing files to trigger errors" do
+    before(:all) { run_generator_test_with_args([], has_errors: true) }
```
robwise: see comment in generator spec helper about has_errors

spec/react_on_rails/support/generator_spec_helper.rb
aaronvb: Okay thanks, will update with these changes.

lib/generators/react_on_rails/generator_helper.rb
```diff
@@ -48,3 +56,19 @@ def copy_file_and_missing_parent_directories(source_file, destination_file = nil
     copy_file source_file, destination_file
   end
 end
+
+module GeneratorErrors
```
justin808: @aaronvb Let's put this in it's own file. Any reason to put in generator_helper.rb?

lib/generators/react_on_rails/install_generator.rb
```diff
@@ -72,14 +79,14 @@ def missing_npm?
         return false unless `which npm`.blank?
         error = "** npm is required. Please install it before continuing."
         error << "https://www.npmjs.com/"
-        puts error
+        GeneratorErrors.add_error error
```
justin808: parens

lib/generators/react_on_rails/install_generator.rb
```diff
@@ -44,13 +46,18 @@ class InstallGenerator < Rails::Generators::Base
                    aliases: "-b"
 
       def run_generators # rubocop:disable Metrics/CyclomaticComplexity
-        return unless installation_prerequisites_met?
+        return output_errors unless installation_prerequisites_met?
```
justin808: I really don't like this as output_errors looks like a return value.

Let's rename the method to `print_errors`

```
unless installation_prerequisites_met?
  print_errors
  return
end
```

@samnang @mapreal19 @alexfedoseev 

How do feel about the pattern of `return fn` where `fn` returns nil.

lib/generators/react_on_rails/generator_helper.rb
aaronvb: @justin808 @robwise suggested to put this in the GeneratorHelper module. I can move it to a separate file.

aaronvb:
Updated! Pending checks

justin808:
Thanks @aaronvb

PR #165, opened 2015-12-17
Only call hide! if defined 

flynfish:
Fixes issue on rails 3.2

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -6,7 +6,7 @@ include GeneratorHelper
 module ReactOnRails
   module Generators
     class BaseGenerator < Rails::Generators::Base # rubocop:disable Metrics/ClassLength
-      hide!
+      hide! if defined? hide!
       source_root(File.expand_path("../templates", __FILE__))
 
       # --redux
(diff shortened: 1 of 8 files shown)


justin808:
@robwise @flynfish what is the effect of not running hide?

flynfish:
@justin808 it's just a convenience method so I don't think there are too many side effects:

http://www.rubydoc.info/docs/rails/4.1.7/Rails%2FGenerators%2FBase.hide!

robwise:
@justin808 @flynfish it makes it so that when you run the command to get a list of generators, the sub-generators that we didn't intend for users to call directly will not appear

justin808:
@flynfish Can you please see if you can call the implementation from Rails 3.x:

```
# File railties/lib/rails/generators/base.rb, line 52
      def self.hide!
        Rails::Generators.hide_namespace self.namespace
      end
```

http://apidock.com/rails/v4.0.2/Rails/Generators/Base/hide!/class

justin808:
@flynfish Any update?

flynfish:
I've been on vacation but I'll try to get to this soon!

justin808:
@flynfish I will try to do a release later today, so if you have a good workaround for this issue, please let me know.

flynfish:
@justin808 unfortunately I haven't had time to try your suggestion and I
won't be able to get to it until next week

justin808:
@flynfish Any update?

flynfish:
@justin808 Closing in preference to #182

PR #172, opened 2015-12-28
Updated the server rendering pool to support precompiled assets.

martyphee:
This PR updates the `ServerRenderingPool` to handle including the `react_on_rails.js` file properly in production.

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -165,7 +165,7 @@ module ReactOnRails
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "stylesheets")
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "images")
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "fonts")
-          Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
+          Rails.application.config.assets.precompile += %w( generated/server-bundle.js react_on_rails )
         DATA
         assets_intializer = File.join(destination_root, "config/initializers/assets.rb")
         if File.exist?(assets_intializer)
--- a/lib/react_on_rails/server_rendering_pool.rb
+++ b/lib/react_on_rails/server_rendering_pool.rb
@@ -108,9 +108,9 @@ var console = { history: [] };
       end
 
       def react_on_rails
-        return ::Rails.application.assets['react_on_rails.js'] unless Rails.env.production?
+        return ::Rails.application.assets["react_on_rails.js"] unless Rails.env.production?
 
-        File.read(Rails.root.join('public', 'assets', Rails.application.assets_manifest.assets['react_on_rails.js']))
+        File.read(Rails.root.join("public", "assets", Rails.application.assets_manifest.assets["react_on_rails.js"]))
       end
     end
   end
--- a/lib/react_on_rails/server_rendering_pool.rb
+++ b/lib/react_on_rails/server_rendering_pool.rb
@@ -76,7 +76,7 @@ module ReactOnRails
           base_js_code = <<-JS
 #{console_polyfill}
           #{bundle_js_code};
-#{::Rails.application.assets['react_on_rails.js']};
+#{react_on_rails};
           JS
           ExecJS.compile(base_js_code)
         else
@@ -105,6 +105,12 @@ var console = { history: [] };
   };
 });
         JS
+      end
+
+      def react_on_rails
+        ::Rails.application.assets['react_on_rails.js'] if Rails.env.development?
+
+        File.read(Rails.root.join('public', 'assets', Rails.application.assets_manifest.assets['react_on_rails.js']))
       end
     end
   end
--- a/lib/react_on_rails/server_rendering_pool.rb
+++ b/lib/react_on_rails/server_rendering_pool.rb
@@ -108,7 +108,7 @@ var console = { history: [] };
       end
 
       def react_on_rails
-        ::Rails.application.assets['react_on_rails.js'] if Rails.env.development?
+        return ::Rails.application.assets['react_on_rails.js'] unless Rails.env.production?
 
         File.read(Rails.root.join('public', 'assets', Rails.application.assets_manifest.assets['react_on_rails.js']))
       end
--- a/spec/react_on_rails/support/shared_examples/base_generator_examples.rb
+++ b/spec/react_on_rails/support/shared_examples/base_generator_examples.rb
@@ -95,7 +95,7 @@ shared_examples "base_generator:base" do |options|
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "stylesheets")
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "images")
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "fonts")
-      Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
+      Rails.application.config.assets.precompile += %w( generated/server-bundle.js react_on_rails )
     EXPECTED
     assert_file("config/initializers/assets.rb") { |contents| assert_match(expected, contents) }
   end


rstudner:
(insert happy face here)

martyphee:
In the meantime this monkey patch in `initializers/react_on_rails.rb` will do the same thing

```
module ReactOnRails
  class ServerRenderingPool
    class << self
      def create_js_context
        server_js_file = ReactOnRails.configuration.server_bundle_js_file
        if server_js_file.present? && File.exist?(server_js_file)
          bundle_js_code = File.read(server_js_file)
          base_js_code = <<-JS
#{console_polyfill}
          #{bundle_js_code};
#{react_on_rails};
          JS
          ExecJS.compile(base_js_code)
        else
          if server_js_file.present?
            msg = "You specified server rendering JS file: #{server_js_file}, but it cannot be "\
              "read. You may set the server_bundle_js_file in your configuration to be \"\" to "\
              "avoid this warning"
            Rails.logger.warn msg
            puts msg
          end
          ExecJS.compile("")
        end
      end

      def react_on_rails
        return ::Rails.application.assets['react_on_rails.js'] if Rails.env.development?

        File.read(Rails.root.join('public', 'assets', Rails.application.assets_manifest.assets['react_on_rails.js']))
      end
    end
  end
end

**Comment by justin808** [comment]
@martyphee Can you try to fix the test failures?
@rstudner Can you take responsibility for code review?

@martyphee We're deploying this for production already on Heroku. And I've used the current mechanism in several other projects.

**NOTE**: within a few weeks, I'll release v2.0, and this problem is N/A as the react_on_rails.js comes from npm. Please see https://github.com/shakacode/react_on_rails/pull/148.

**Comment by martyphee** [comment]
The current mechanism works fine without server rendering, but for SEO
we'll need it.

**Comment by justin808** [comment]
@martyphee We're doing server rendering at http://www.reactrails.com. 

Right click and view source.

**Comment by martyphee** [comment]
How is that working for you then?  When I deployed with server rendering I
was getting the error.  It's working now (www.mybiabbrewery.com/volumes).

**Comment by justin808** [comment]
@martyphee What is "the error"? I'll invite you to our slack room if you send me an email.

I'm hesitant to change this in the current version unless we better understand why the current deployment of https://github.com/shakacode/react-webpack-rails-tutorial/ deploys just fine to Heroku and why yours and @thewoolleyman have issues. Maybe it's b/c of Rails 5 or a new version of sprockets?

**Comment by martyphee** [comment]
Pretty sure I found it was because of the sprockets version.

`::Rails.application.assets['react_on_rails.js']` wasn't returning anything.

**Comment by justin808** [comment]
@martyphee Does this change affect those using the older version of sprockets? Note, I updated the gems on the https://github.com/shakacode/react-webpack-rails-tutorial/ recently.

**Comment by thewoolleyman** [comment]
@justin808 @martyphee Yes as I noted on #170  `Rails.application.assets` will always be nil when `config.assets.compile` is set to false in the Rails config (which is the normal case for a prod env).  And yes, this is a recent change in Sprockets that has broken other people as noted here: https://github.com/rails/sprockets-rails/issues/237

So, my only suggestion would be rather than basing this on production env like this (https://github.com/shakacode/react_on_rails/pull/172/files#diff-10c7b74a600ef3ecde9dca86ed42c16eR111), if we could somehow directly detect the `config.assets.compile` setting and base the decision off that, it would be better, and work even if they have that flag turned off in non-prod envs.

**Comment by martyphee** [comment]
Sounds like a good idea.  Should be doable.

File: lib/react_on_rails/server_rendering_pool.rb [hunk 1]
```diff
@@ -106,6 +106,12 @@ def console_polyfill
 });
         JS
       end
+
+      def react_on_rails
+        return ::Rails.application.assets["react_on_rails.js"] unless Rails.env.production?
```
justin808, line 111: Maybe we should just check the return value, and if not sufficient, then use the next statement? I'm concerned that we'll break those using older versions of Sprockets. Additionally, production environments could have other names than "production". I've often deployed apps with "staging" as the environment.

I think we should memoize this and be flexible.

Plus, this will all go away in v2.0 as ReactOnRails will be within the bundle_js_code.

spec/react_on_rails/support/shared_examples/base_generator_examples.rb
```diff
@@ -95,7 +95,7 @@
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "stylesheets")
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "images")
       Rails.application.config.assets.paths << Rails.root.join("client", "assets", "fonts")
-      Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
+      Rails.application.config.assets.precompile += %w( generated/server-bundle.js react_on_rails )
```
justin808, line 98: This seems odd, as the rails installation should have the Gem's JS file. We're just probably looking in the wrong place.

lib/generators/react_on_rails/base_generator.rb
```diff
@@ -165,7 +165,7 @@ def append_to_assets_initializer
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "stylesheets")
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "images")
           Rails.application.config.assets.paths << Rails.root.join("client", "assets", "fonts")
-          Rails.application.config.assets.precompile += %w( generated/server-bundle.js )
+          Rails.application.config.assets.precompile += %w( generated/server-bundle.js react_on_rails )
```
justin808, line 168: There's many other gems that provide JS files (take jquery-rails, for example).

Do they also put them in the precompile list?

spec/react_on_rails/support/shared_examples/base_generator_examples.rb
martyphee, line 98: Why is this odd?  Unless the file is specified in the list it won't be generated during pre-compile and not loadable.  Not sure how to look up the file from the gems.

lib/generators/react_on_rails/base_generator.rb
martyphee, line 168: It won't be loaded.  You have to add it manually to the precompile list or
include it in the `application.js`.

[image: Inline image 1]

spec/react_on_rails/support/shared_examples/base_generator_examples.rb
justin808, line 98: Maybe this is just the way that the new Sprockets expects such files?

martyphee, line 98: I'm using an old sprokets on another project and we have to add the file
individually if we want them precompiled.

justin808:
Can I get confirmation that this does not break old sprockets users. We also have some Rails 3 users.

Within a couple of days, I'm going to release https://github.com/shakacode/react_on_rails/pull/148

justin808:
@martyphee, etc. If we can get word this doesn't break the old version of sprockets, I'll merge this.

martyphee:
I haven't had time to validate it.  I'll try to this week on a current
project.

justin808:
@martyphee I'll cut a build as soon as you get back to me.

martyphee:
Sorry for the delay.  I'll try to get this confirmed today.

rstudner:
@martyphee i'm trying to verify as well (just going through the setup pains for 'getting there' hah).  What version of rails/sprockets were you planning on testing against?

martyphee:
Lates of the 3.x branch.

gem 'rails', '~> 3.2', '>= 3.2.22'

rstudner:
Okay, right on.  I tried to just monkey patch my local react_on_rails gem to have your changes but it doesn't like that, so have to get a bit more formal on pulling things down :)

otherwise I get this beauty:
09:06 $ rails generate react_on_rails:install
Running via Spring preloader in process 31479
fatal: Not a git repository (or any of the parent directories): .git
** You have uncommitted code. Please commit or stash your changes before continuing

justin808:
@rstudner @martyphee I need to get confirmation that this doesn't break for Rails 3.2 and Rails 4.

I'm about to merge v2 into master, and this change is not relevant for v2 as JS assets come from npm.

martyphee:
I can't get the install to run with rails 3.x.  Ruby 2.2.3.

[WARNING] Could not load generator
"generators/react_on_rails/base_generator". Error: undefined method `hide!'
for ReactOnRails::Generators::BaseGenerator:Class.

/Users/martinphee/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/react_on_rails-1.2.0/lib/generators/react_on_rails/base_generator.rb:9:in
`<class:BaseGenerator>'

/Users/martinphee/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/react_on_rails-1.2.0/lib/generators/react_on_rails/base_generator.rb:8:in
`<module:Generators>'

/Users/martinphee/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/react_on_rails-1.2.0/lib/generators/react_on_rails/base_generator.rb:7:in
`<module:ReactOnRails>'

/Users/martinphee/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/react_on_rails-1.2.0/lib/generators/react_on_rails/base_generator.rb:6:in
`<top (required)>'

/Users/martinphee/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:251:in
`require'

thewoolleyman:
I'm going to try out the examples with the latest versions in a Rails 3 app, and I'll report if I get the same failures.

justin808:
Is this issue related: https://github.com/rails/sprockets-rails/commit/8fc9e0b446dd0ddc812ce6a88d32727e3015016c

This is what we had to do in our production project (diff):

```
-Rails.application.config.assets.precompile += %w(application_*.* *-client-bundle.*)
 +
 +if Rails.env.production?
 +  Rails.application.config.assets.precompile += %w(application_prod.* *-client-bundle.*)
 +else
 +  Rails.application.config.assets.precompile += %w(application_dev.* *-client-bundle.*)
 +end
  ```


We have these versions, and this issue just came up recently.
```
    sprockets (3.5.2)
      concurrent-ruby (~> 1.0)
      rack (> 1, < 3)
    sprockets-rails (3.0.0)
      actionpack (>= 4.0)
      activesupport (>= 4.0)
      sprockets (>= 3.0.0)
```

CC: @samnang @josiasds

justin808:
Big thanks to @thewoolleyman who's going to be shepherding this one with me.

samnang:
@justin808 I'm not pretty sure because the use case, most of times people will require react_on_rails in their application.js, why do we need to add into precompile list.

justin808:
@samnang this issue will only affect v1.x users, not the new v2.x which uses npm.

martyphee:
@thewoolleyman I'm running into the same issues with this PR.  Not sure if I can get it working with Rails 3.

justin808:
v2 is out. This issue is not valid any more. Please update to v2.

justin808:
Note, there are issues with sprockets v3 that I've not yet worked out. I could not update the https://github.com/shakacode/react-webpack-rails-tutorial/ to sprockets v3. If anybody can try that, much appreciated.

PR #182, opened 2016-01-04
Change hide! to the implementation to support rails 3.2

flynfish:
This allows the generators to work in rails 3.2 because the hide! method only exists in rails 4.


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -6,7 +6,7 @@
 module ReactOnRails
   module Generators
     class BaseGenerator < Rails::Generators::Base # rubocop:disable Metrics/ClassLength
-      hide!
+      Rails::Generators.hide_namespace self.namespace
```
justin808: Let's use parens for the method call.

Any doc link for this?

flynfish: Is this what you are looking for?
http://apidock.com/rails/Rails/Generators/hide_namespace/class

justin808: Yep! In both 4.2.1 and 3.2.13!

http://apidock.com/rails/v3.2.13/Rails/Generators/hide_namespace/class

http://apidock.com/rails/v4.2.1/Rails/Generators/hide_namespace/class

justin808:
Please make the paren change, and I'll merge.

flynfish:
@justin808 done

PR #192, opened 2016-01-11
Fix #189 Generator Error for application.scss

### LINKED ISSUE #189: Generator Error for application.scss
When I was creating tests for the generator errors, I noticed I could not get the application.scss exist? error to show up when skipping the creation of it in the test setup using ```application_css: false```

Here we check if ```application.scss``` exists:
https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/bootstrap_generator.rb#L56

However, it will always exist because of the previous method above it, so it's impossible to get the error:

https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/bootstrap_generator.rb#L23

Do you think the add error is necessary?

(cc: @robwise)

aaronvb:
Removed the File.exists? check for application.scss because it gets created if it doesn’t exist in the method before it.

(cc @robwise)

### DIFF
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -53,11 +53,7 @@ module ReactOnRails
 
         application_scss = File.join(destination_root, "app/assets/stylesheets/application.scss")
 
-        if File.exist?(application_scss)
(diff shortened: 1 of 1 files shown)


justin808:
Looked good.

PR #204, opened 2016-01-14
Improve package json template

robwise:
Also makes some fixes to the README.

This improves the recent hotfix so that we can match against all types of gem versions, including `beta` or `rc` suffixed ones.

### DIFF
--- a/spec/react_on_rails/generators/install_generator_spec.rb
+++ b/spec/react_on_rails/generators/install_generator_spec.rb
@@ -1,4 +1,5 @@
-require File.expand_path("../../support/generator_spec_helper", __FILE__)
+require_relative "../support/generator_spec_helper"
+require_relative "../support/version_test_helpers"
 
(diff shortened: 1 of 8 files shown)


spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -210,7 +211,7 @@
   end
 
   context "with missing files to trigger errors" do
-    it "GeneratorMessages has the missing file error" do
+    specify "GeneratorMessages has the missing file error" do
```
justin808, line 214: what is specify?

justin808:
@robwise CI failing.

spec/react_on_rails/generators/install_generator_spec.rb
robwise, line 214: @justin808 it's the same thing as `it` basically, just reads more like english in this case because GeneratorMessages isn't the subject

PR #205, opened 2016-01-15
Clarify Generator Messages

aaronvb:
Updated the 'errors' wording used in GeneratorMessages to 'messages' because this singleton is also outputting warning and info messages.


justin808:
LGTM!

PR #213, opened 2016-01-19
Better to return created reactElement.

DimaZab:
If I use only React component and I want to manually set state, maybe better to return rendered reactElement?

```javascript
  render(name, props, domNodeId) {
    const reactElement = createReactElement({ name, props, domNodeId });
    return ReactDOM.render(reactElement, document.getElementById(domNodeId));
  }

// then I can
  var renderedElement = ReactOnRails.render('Element', {}, 'app');
  renderedElement.setState({ option: value });

```

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/213)

### DIFF
--- a/node_package/src/ReactOnRails.js
+++ b/node_package/src/ReactOnRails.js
@@ -29,10 +29,11 @@ ctx.ReactOnRails = {
    * @param name Name of your registered component
    * @param props Props to pass to your component
    * @param domNodeId
+   * @returns {virtualDomElement} Reference to your component's backing instance
    */
   render(name, props, domNodeId) {
     const reactElement = createReactElement({ name, props, domNodeId });
-    ReactDOM.render(reactElement, document.getElementById(domNodeId));
+    return ReactDOM.render(reactElement, document.getElementById(domNodeId));
   },
 
   /**
--- a/node_package/tests/ReactOnRails.test.js
+++ b/node_package/tests/ReactOnRails.test.js
@@ -1,0 +1,26 @@
+/* eslint-disable react/no-multi-comp */
+import test from 'tape';
+import React from 'react';
+import ReactOnRails from '../src/ReactOnRails';
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+import JsDom from 'jsdom';
+
+if (!canUseDOM) {
+  global.document = JsDom.jsdom('<div id="root"></div>');
+  global.window = document.defaultView;
+}
+
+test('ReactOnRails render returns a virtual DOM element for component', (assert) => {
+  assert.plan(1);
+  const R1 = React.createClass({
+    render() {
+      return (
+        <div> WORLD </div>
+      );
+    },
+  });
+  ReactOnRails.register({ R1 });
+  const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalInstance._currentElement.type;
+  assert.deepEqual(actual, R1,
+    'ReactOnRails render should return a virtual DOM element for component');
+});
--- a/README.md
+++ b/README.md
@@ -356,7 +356,7 @@ If you are using [react-rails](https://github.com/reactjs/react-rails) in your p
 
 - Remove the 'react-rails' gem from your Gemfile.
 
-- Remove the generated generated lines for react-rails in your application.js file. 
+- Remove the generated generated lines for react-rails in your application.js file.
 
 ```
 //= require react
@@ -377,6 +377,7 @@ Note: If you have components from react-rails you want to use, then you will nee
 + [React Router](docs/additional_reading/react_router.md)
 + [RSpec Configuration](docs/additional_reading/rspec_configuration.md)
 + [Server Rendering Tips](docs/additional_reading/server_rendering_tips.md)
++ [Rails View Rendering from Inline JavaScript](docs/additional_reading/rails_view_rendering_from_inline_javascript.md)
 + [Tips](docs/additional_reading/tips.md)
 + [Tutorial for v2.0](docs/tutorial-v2.md)
 + [Turbolinks](docs/additional_reading/turbolinks.md)
--- a/docs/additional_reading/rails_view_rendering_from_inline_javascript.md
+++ b/docs/additional_reading/rails_view_rendering_from_inline_javascript.md
@@ -1,0 +1,26 @@
+# Using ReactOnRails in JavaScript
+You can easily render React components in your JavaScript with `render` method that returns a [reference to the component](https://facebook.github.io/react/docs/more-about-refs.html) (virtual DOM element).
+
+```js
+// componentName - name of your registered component;
+// props - Object which contains the properties to pass to the react object;
+// elementId - id of an element where we render our React component;
+ReactOnRails.render(componentName, props, elementId)
+```
+
+## Why do we need this?
+Imagine that we have some event with jQuery, it allows us to set component state manually.
+
+```html
+<input id="input" type="range" min="0" max="100" />
+<div id="root"></div>
+
+<script>
+  var input = $("#input");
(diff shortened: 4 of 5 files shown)


justin808:
@DimaZab I've never seen that pattern. What's the use case? Jade?

@alexfedoseev any opinion?

justin808:
@DimaZab Maybe push up small test repo demonstrating this?

alexfedoseev:
Also never seen that pattern. Use case will be nice.

DimaZab:
@justin808, @alexfedoseev https://github.com/DimaZab/example here is an example, see README.

And if we look in react source code we will see that render method return created react element.

```javascript
  render: function (nextElement, container, callback) {
    return ReactMount._renderSubtreeIntoContainer(null, nextElement, container, callback);
  }
```

alexfedoseev:
@DimaZab @justin808 It makes sense :+1:

justin808:
OK -- so no harm in doing this. Yikes, and yuck for mixing jQuery with React! :smile:

justin808:
@DimaZab can you please add a JS test for this. I think JS will be sufficient. rather than a full integration test.

DimaZab:
@justin808 I agree about jQuery and React but it's only for the first time.

DimaZab:
@justin808 I've try to write test for this method:

```javascript
  render(name, props, domNodeId) {
    const reactElement = createReactElement({ name, props, domNodeId });
    return ReactDOM.render(reactElement, document.getElementById(domNodeId));
  },
```

But get an error: 'document is not defined'. So, can I use something like mock-browser (https://www.npmjs.com/package/mock-browser) or may be you know better way?

justin808:
@DimaZab Sure -- you'll need to use something like that. Thanks!

DimaZab:
@justin808 done =)

node_package/tests/ReactOnRails.test.js
```diff
@@ -0,0 +1,26 @@
...
+  assert.plan(1);
+  const R1 = React.createClass({
+    render() {
+      return (
+        <div> WORLD </div>
+      );
+    },
+  });
+  ReactOnRails.register({ R1 });
+  const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalInstance._currentElement.type;
```
justin808, line 23: can we name this "actualElement" or "actualComponent"?

node_package/tests/ReactOnRails.test.js
```diff
@@ -0,0 +1,26 @@
...
+  assert.deepEqual(actual, R1,
```
justin808, line 24: are we comparing elements or components?

https://facebook.github.io/react/docs/glossary.html

node_package/src/ReactOnRails.js
```diff
@@ -31,7 +31,7 @@ ctx.ReactOnRails = {
    */
   render(name, props, domNodeId) {
     const reactElement = createReactElement({ name, props, domNodeId });
-    ReactDOM.render(reactElement, document.getElementById(domNodeId));
+    return ReactDOM.render(reactElement, document.getElementById(domNodeId));
```
justin808, line 36: We need to fix the JS Doc, and we should have a link to a file in the /docs directory that explains why you'd do this.

node_package/tests/ReactOnRails.test.js
justin808, line 24: @alexfedoseev CC

node_package/tests/ReactOnRails.test.js
robwise, line 23: @justin808 I think we should stick with the JSX testing convention, which is to leave it as `actual`

justin808:
@DimaZab, I'd like to get the merged as soon as you complete the changes I've requested. Thanks.

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-:-K8mBl_pEbCfwmmyUm55:1216942195)*

DimaZab:
@justin808 done

justin808:
Some additional small items.

---

Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[docs/additional_reading/javascript_rendering.md, line 26 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8v6vNxhaJqQzsBC557:-K8v6vNxhaJqQzsBC558:1966322473) ([raw file](https://github.com/shakacode/react_on_rails/blob/679ce318ac83eba30a5c5cff85a00423d9c4de11/docs/additional_reading/javascript_rendering.md#L26)):*
Need empty line at end of file per red marker.

---

*[node_package/src/ReactOnRails.js, line 31 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8v67kkPQiwW6OYknx5:-K8v67kkPQiwW6OYknx6:-30761891) ([raw file](https://github.com/shakacode/react_on_rails/blob/679ce318ac83eba30a5c5cff85a00423d9c4de11/node_package/src/ReactOnRails.js#L31)):*
we need to put the return value here, @DimaZab

---

*[node_package/tests/ReactOnRails.test.js, line 24 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8fIJUMdH3cHqlL8nvI-r1-24:-K8v6NNsETmMMDzImnZF:-1566343114) ([raw file](https://github.com/shakacode/react_on_rails/blob/7e77cf96d995acacbb4598d8388d11113a3cd133/node_package/tests/ReactOnRails.test.js#L24)):*
I'm not 100% sold that we're using the right terminology. However, it's not critical. @jbhatab? 

---

*[README.md, line 380 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8v6ewmSaU1sznhXPFp:-K8v6ewmSaU1sznhXPFq:1967368621) ([raw file](https://github.com/shakacode/react_on_rails/blob/679ce318ac83eba30a5c5cff85a00423d9c4de11/README.md#L380)):*
Let's rename to 

Rails View Rendering from Inline JavaScript

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-:-K8v71oh9B8UGeH-irg4:1325687161)*

justin808:
Reviewed 5 of 6 files at r2.
Review status: 2 of 5 files reviewed at latest revision, 6 unresolved discussions.

---

*[docs/additional_reading/rails_view_rendering_from_inline_javascript.md, line 7 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8z5mWB8sTEWPi88xvS:-K8z5mWB8sTEWPi88xvT:1430480309) ([raw file](https://github.com/shakacode/react_on_rails/blob/386599a68d24af7729bc98e740b3b48d23a6f50e/docs/additional_reading/rails_view_rendering_from_inline_javascript.md#L7)):*
We need to state the return value here.

We should copy the jsdoc from the JS file.

---

*[node_package/src/ReactOnRails.js, line 32 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/213#-K8z5wofdg-lbY2Kv3sM:-K8z5wofdg-lbY2Kv3sN:-1706355024) ([raw file](https://github.com/shakacode/react_on_rails/blob/386599a68d24af7729bc98e740b3b48d23a6f50e/node_package/src/ReactOnRails.js#L32)):*
This should go into the doc rails_view_rendering_from_inline_javascript.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/213)*

DimaZab:
@robwise Thank you)

PR #215, opened 2016-01-20
Support multiple components on a page using the same Redux store

justin808:
## Multiple React Components on a Page with One Store
You may wish to have 2 React components share the same the Redux store. For example, suppose you want a React component for your header, and underneath the header, you might have a conventional Rails page or another React component. With 2 React components, header and body ones, we'd like them both to share the same Rails view props hydrated Redux store, and we'd like this to all work with Turbolinks to minimize reloading the JavaScript. In the below examples, suppose our Redux store is called appStore, and you have 2 React top level components, "Comments" and "Blogs". You might be wondering on why we'd want to do this. Here's a few reasons:

1. As much as we love React and Redux, it's still way slower and more expensive than creating simple pages using standard Rails. We like the hybrid approach where we can focus on React + Redux for super high fidelity UX, while using Rails for simpler needs.
2. We want a React component for the header as we plan to have notification indicators in our header.

Main Layout:
```erb
  <% redux_store("appStore", props) %>;
  <%= render_component("Header") %>
  yield
```

Comments "App" view
```erb
  <%= render_component("Comments") %>
```

Blog "App" view
```erb
  <%= render_component("Blogs") %>
```

Inside your ReactOnRails registered component definition, where you'd normally do the store hookup, you instead get the store from ReactOnRails, like this:

```js
  // getStore will initialize the store if not already initialized, so creates or retrieves store
  const appStore = ReactOnRails.getStore("appStore");
  return (
    <Provider store={appStore}>
      <NonRouterCommentsContainer />
    </Provider>
  );
```

In either your store definition or your registration files, supposing you created a store called `appStore`, you will call:

```
ReactOnRails.registerStore({
  appStore
});
```

Note, this is a **storeCreator**, meaning that it is a function that takes props and returns a store.

So what will be happening?

The view helper `initialize_store(store_name, props)` will replace any cached value of store and rebuilds store by calling the storeGenerator with the props passed from Rails. This will be done with inline JavaScript to ensure that it happens after the page loaded/changed callback from the DOM or TurboLinks, which then renders the components.

Then the components are created as "pure render functions" (aka generator functions) which means they take props and return a React.Component. In the case of using `initializeStore`, we will not be passing any props to the component. The component will instead call this to get the store.

```js
const store = ReactOnRails.getStore("appStore");
```

Note, we will not be doing any partial updates to the Redux store. When the page content loads, we will rehydrate the store with whatever props are placed on the page.

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/215)


README.md
```diff
@@ -64,6 +64,7 @@ Please see [Getting Started](#getting-started) for how to set up your Rails proj
 + [Generator](#generator)
     - [Understanding the Organization of the Generated Client Code](#understanding-the-organization-of-the-generated-client-code)
     - [Redux](#redux)
+      - [Multiple React Components on a Page with One Store](#multiple-react-components-on-a-page-with-one-store)
```
robwise, line 79: we should make sure we put additions to the README in the proper place. I put this under `Redux`

README.md
```diff
@@ -275,6 +276,56 @@ If you have used the `--redux` generator option, you will notice the familiar ad
 
 Note the organizational paradigm of "bundles". These are like application domains and are used for grouping your code into webpack bundles, in case you decide to create different bundles for deployment. This is also useful for separating out logical parts of your application. The concept is that each bundle will have it's own Redux store. If you have code that you want to reuse across bundles, including components and reducers, place them under `/client/app/lib`.
 
+#### Multiple React Components on a Page with One Store
+You may wish to have 2 React components share the same the Redux store. For example, if your navbar is a React component, you may want it to use the same store as your component in the main area of the page. You may even want multiple React components in the main area, which allows for greater modularity. In addition, you may want this to work with Turbolinks to minimize reloading the JavaScript.
```
robwise: I noticed this section kept switching from 2nd person to 1st person plural, and I guess the rest of the README is 2nd person, so I switched it all to that.

README.md
```diff
@@ -275,6 +276,56 @@ If you have used the `--redux` generator option, you will notice the familiar ad
...
+return (
+  <Provider store={appStore}>
+    <NonRouterCommentsContainer />
+  </Provider>
+);
+```
+
+From your Rails view, you can use the provided helper `initialize_store(store_name, props)` to create a fresh version of the store (because it may already exist if you came from visiting a previous page).
+
+Components are created as [stateless function(al) components](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions). Since you can pass in initial props via `initializeStore`, you do not need to pass any props directly to the component. Instead, the component hydrates by connecting to the store.
```
robwise: Don't confuse `stateless function component` with `pure render component`. They are slightly different. Technically, I can still use the ES6 class syntax to create a component, use a pure render mixin, and it's a pure render component. A stateless function component specifically means a function that just returns some JSX. I linked to the corresponding section in the React docs.

README.md
```diff
@@ -275,6 +276,56 @@ If you have used the `--redux` generator option, you will notice the familiar ad
...
+```
+
+When registering your component with React on Rails, you can get the store via `ReactOnRails.getStore`:
+
+```js
+// getStore will initialize the store if not already initialized, so creates or retrieves store
+const appStore = ReactOnRails.getStore("appStore");
+return (
+  <Provider store={appStore}>
+    <NonRouterCommentsContainer />
```
robwise: what is `NonRouterCommentsContainer`??

README.md
```diff
@@ -275,6 +276,56 @@ If you have used the `--redux` generator option, you will notice the familiar ad
...
+  </Provider>
+);
+```
+
+From your Rails view, you can use the provided helper `initialize_store(store_name, props)` to create a fresh version of the store (because it may already exist if you came from visiting a previous page).
+
+**app/views/layouts/application.html.erb**
+```erb
+...
+<% initialize_store("appStore", props) %>;
```
robwise: This is a problem. We are initializing the store with props inside of the application layout, but those props need to have data specific to each different page (in this case, the initial data for comments and blogs)?

robwise:
We also need to explain why you would want to have everything in a single store instead of just having multiple stores.

README.md
justin808: correct

robwise: @justin808 lol this was a rhetorical question! You can't pass in page-specific props in a layout file without some type of yield block! Alternatively we can recommend calling the `initialize_store` method from the view, and just make sure to always pass in navbar props.

justin808: @robwise Yes -- the props come from the controller.

robwise: @justin808 so you are going to need to ensure that you create an instance variable in your controller for every page, even static pages that have a navbar, and it always must be named exactly the same thing as what you have in your layout file that's creating the store? We need to document all of this.

justin808:
I addressed those. Please review the latest commit. Feel free to commit directly for small typos.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-:-K8mELnzkcm0vDlw56np:-54150192)*

robwise:
Reviewed 1 of 1 files at r3.

---

*[README.md, line 280 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8o5Tig_SATZmSNl2fR:-K8o5Tig_SATZmSNl2fS:1567794868) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L280)):*
`share the same the same` needs fixing

---

*[README.md, line 280 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oD7Z3iY0DGQhRjnuV:-K8oD7Z4-G2ypyBYdGj0:1077054869) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L280)):*
In the bit you added, "this" is referring to using Turbolinks, when really you meant it to refer to using the same store.

---

*[README.md, line 282 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oDPYFFAcMqkfG2W_T:-K8oDPYFFAcMqkfG2W_U:1732367643) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L282)):*
You're using first person all of the sudden, please stick to the same narrative. Also, I thought we all agreed a long time ago that naming these things `App` was a bad idea because they are not actually apps, and you can have more than one on the same page.

---

*[README.md, line 304 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oDnidlYBQ7rziS_df:-K8oDnidlYBQ7rziS_dg:-804750474) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L304)):*
It also means you are going to have to pass props for the `Navbar` in every single controller, besides the props for whatever the controller actually is called. I think we can therefore enhance this, such as providing `@react_props` in the application controller and then doing a merge in the controller on that instance variable.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

justin808:

---

*[README.md, line 304 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oDnidlYBQ7rziS_df:-K8pvlxU1v__trEd8Yd_:-795562235) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L304)):*
???

Lost here

content_for:

is probably good for this part

include a partial for the header in the target view

and that partial has a content_for

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

robwise:

---

*[README.md, line 304 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oDnidlYBQ7rziS_df:-K8q3MY6-u2BhctMukhW:134676122) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L304)):*
@justin808 So say you are on a message inbox page, where your main component is an inbox. The Navbar and the Inbox component will want to share the same store, since the Navbar has an "unread messages" icon that shows the count of unread messages, and you want this number to update when you read a message in the inbox. However, the Navbar has other data that needs to be fed to it that has nothing to do with messages, perhaps such as URLs for the navigation, other notifications, user avatar, etc.

So you would either need to pass both the inbox props and the Navbar props in to the `initialize_store` helper inside of the inbox view, or you would need to initialize the store in the layout where the navbar is and pass in the Navbar props there, along with a `content_for :store` type thing that you would then pass in from the inbox view, which I think is what you're suggesting here and I agree with.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

justin808:

---

*[README.md, line 304 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-K8oDnidlYBQ7rziS_df:-K8rDJxo15JOpRUNakGc:1339575113) ([raw file](https://github.com/shakacode/react_on_rails/blob/dfecaf1191ab0729ad93aef7c20f9e9776685de4/README.md#L304)):*
OK -- so no change to the docs? It's clearly up to the dev on how to organize the code!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

justin808:
@alexfedoseev @robwise This is ready for review! Tests added. Server rendering working.

Only to-do is to 

- [ ] refactor `react_on_rails_helper.rb` to use method objects.

justin808:
3.0.0.beta.1 is on https://rubygems.org/gems/react_on_rails
3.0.0-beta.1 is on: https://www.npmjs.com/package/react-on-rails

Feedback appreciated.

alexfedoseev:
Looks awesome! Let's try it out :+1: 

---

Review status: 0 of 44 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-:-KA17ATfjQ1elaGYvCGi:1618958574)*

alexfedoseev:
Reviewed 40 of 44 files at r4, 4 of 5 files at r5, 2 of 2 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

justin808:
@robwise Once you code review, I'll merge. I'll probably update the https://github.com/shakacode/react-webpack-rails-tutorial/ as well.

robwise:
Reviewed 40 of 44 files at r4, 4 of 5 files at r5, 2 of 2 files at r6.

---

*[node_package/src/StoreRegistry.js, line 65 \[r6\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-KA3UwYBV6Y7TgQrvven:-KA3UwYC6zY8wbS4AawL:-529796308) ([raw file](https://github.com/shakacode/react_on_rails/blob/b51b34ba7d95cc162abf4279b108ac0d257aac3a/node_package/src/StoreRegistry.js#L65)):*
"Get **a** Map"

---

*[node_package/src/StoreRegistry.js, line 73 \[r6\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-KA3V26urn_YJ5GLkr37:-KA3V26urn_YJ5GLkr38:-529796308) ([raw file](https://github.com/shakacode/react_on_rails/blob/b51b34ba7d95cc162abf4279b108ac0d257aac3a/node_package/src/StoreRegistry.js#L73)):*
"Get **a** Map"

---

*[README.md, line 305 \[r6\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/215#-KA3T5AE6cVYHbK6KhpB:-KA3T5AE6cVYHbK6KhpC:1030688017) ([raw file](https://github.com/shakacode/react_on_rails/blob/b51b34ba7d95cc162abf4279b108ac0d257aac3a/README.md#L305)):*
"need to make **a** function"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/215)*

Issue #211, opened 2016-01-18
GeneratorMessages Warning


aaronvb:
Not sure if this is a problem on my end, but when I was writing updated tests for GeneratorMessages I could not get the `GeneratorMessages.add_warning` test to pass.

```
Failure/Error: Rainbow("WARNING: #{msg}").orange
       
       NoMethodError:
         undefined method `orange' for "WARNING: warning":Rainbow::Presenter
```

I then tested a warning message in one of the generators and the same error popped up. You should be able to reproduce this error if you add this to any generator.

```
def print_example_warning_message
   GeneratorMessages.add_warning("example warning message")
end
```

Which results in the same error:
```
format_warning': undefined method `orange' for "WARNING: example warning message":Rainbow::Presenter (NoMethodError)
```

The error points to this method in `generator_messages.rb`.
```
def format_warning(msg)
  Rainbow("WARNING: #{msg}").orange
end
```

I went through the rainbow docs and orange is a supported x11 color so I'm not sure why this is happening.

aaronvb:
Also note that if change orange to yellow(or another base color), it works normally.

justin808:
yellow is too light for a white background

orange worked fine for @robwise and me.

justin808:
@aaronvb still an issue?

aaronvb:
@justin808 Yes, I've pushed a branch in my repo with the new tests and still get the same error.

https://github.com/aaronvb/react_on_rails/tree/add_tests_to_generator_message

These are the tests I added: https://github.com/aaronvb/react_on_rails/commit/427ceac43a720fe1dbcd44cf2d2ff8d789810a84

And the test error that I keep getting:

```
1) GeneratorMessages has a method that can add warnings
     Failure/Error: Rainbow("WARNING: #{msg}").orange
     
     NoMethodError:
       undefined method `orange' for "WARNING: Test warning":Rainbow::Presenter
     # ./lib/generators/react_on_rails/generator_messages.rb:28:in `format_warning'
     # ./lib/generators/react_on_rails/generator_messages.rb:12:in `add_warning'
     # ./spec/react_on_rails/generators/generator_messages_spec.rb:13:in `block (2 levels) in <top (required)>'
```

[...]

PR #217, opened 2016-01-21
Add Tests to Generator Messages

aaronvb:
- Added new tests to Generator Messages.
- Fixed color bug with rainbow by using the orange RGB value instead. #211 
- Added require simplecov helper.


justin808:
Nice work @aaronvb!

PR #230, opened 2016-01-25
Bump rainbow gem version to >= 2.1.0.

aaronvb:
In one of my previous PR's we had to do a 'hacky' fix to get X11 colors to output with the GeneratorMessages. https://github.com/shakacode/react_on_rails/pull/217

I was able to fix the problem with the rainbow gem and get a PR merged: https://github.com/sickill/rainbow/issues/27. The author updated the version of the gem to 2.1.0.

- New version of rainbow fixes bug with X11 colors.
- Use ‘orange’ instead of RGB color.


robwise:
LGTM

PR #248, opened 2016-02-01
Added Foreman as a dependency.

FearMediocrity:
[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/248)


justin808:
@FearMediocrity Most devs have foreman installed globally. Any reason we should add to the gemspec? for contributors?

I could see maybe adding this to what's generated. Maybe it's already there.

FearMediocrity:
I added it because if you follow your instructions I had to manually. ( I was using a sandboxed app to try it out).

If a dev already uses it then it will be an already satisfied dependency, if not it will save a minute of head scratching for the beginner coming from React to Rails.

justin808:
Thanks.

PR #251, opened 2016-02-01
Optimize efficiency of EnsureAssetsCompiled functionality

robwise:
This commit changes the EnsureAssetsCompiled feature that ensures
RSpec tests are run with the latest webpack bundles. Previously,
webpack bundles would be rebuilt every test run, whether or not
they actually needed to be. While one could around this by running
webpack in the background to rebuild the static assets, doing so
required users to name their webpack build script exactly correct
as the technique relied on a `pgrep`.

Instead, the functionality checks the mdate on each generated webpack
bundle/asset against the mdate of every file inside of the `client`
directory, which is all encapsulated by the standard library method
`FileUtils.uptodate?` and is very performant. This eliminates the
need to check for an existing webpack process because the webpack
bundles will be up to date anyway if the latter is running.

This required adding some fixtures for proper testing. These can be
found in `spec/react_on_rails/fixtures/webpack_assets`.

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/251)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,16 @@ Contributors: please follow the recommendations outlined at [keepachangelog.com]
 ## [Unreleased]
 ##### Added
 - Added forman to gemspec in case new dev does not have it globally installed. [#248](https://github.com/shakacode/react_on_rails/pull/248)
+##### Changed
+- Changed the EnsureAssetsCompiled feature that ensures RSpec tests are run with the latest webpack bundles. Previously, webpack bundles would be rebuilt every test run, whether or not they actually needed to be. While one could around this by running webpack in the background to rebuild the static assets, doing so required users to name their webpack build script exactly correct as the technique relied on a `pgrep`.
+
+  Instead, the new functionality checks the mdate on each generated webpack bundle/asset against the mdate of every file inside of the `client` directory, which is all encapsulated by the standard library method `FileUtils.uptodate?` and is very performant. This eliminates the need to check for an existing webpack process because the webpack bundles will be up to date anyway if the latter is running.
+
+  Users need not make any changes to their code to benefit from the enhancement, assuming all of their generated files are output to `javascripts/generated`, `stylesheets/generated`, `fonts/generated`, or `images/generated`. [#251](https://github.com/shakacode/react_on_rails/pull/251)
 
 ## [2.3.0] - 2016-02-01
 ##### Added
-- Added polyfills for `setInterval` and `setTimeout` in case other libraries expect these to exist. 
+- Added polyfills for `setInterval` and `setTimeout` in case other libraries expect these to exist.
 - Added much improved debugging for errors in the server JavaScript webpack file.
 - See [#244](https://github.com/shakacode/react_on_rails/pull/244/) for these improvements.
 
--- a/lib/react_on_rails.rb
+++ b/lib/react_on_rails.rb
@@ -7,11 +7,12 @@ require "react_on_rails/server_rendering_pool"
 require "react_on_rails/engine"
 require "react_on_rails/version_syntax_converter"
 require "react_on_rails/ensure_assets_compiled"
+require "react_on_rails/webpack_assets_status_checker"
 require "react_on_rails/git_utils"
 require "react_on_rails/utils"
 
 module ReactOnRails
   def self.configure_rspec_to_compile_assets(config, metatag = :js)
-    config.before(:example, metatag) { ReactOnRails::EnsureAssetsCompiled.check_built_assets }
+    config.before(:example, metatag) { EnsureAssetsCompiled.build.call }
   end
 end
--- a/lib/react_on_rails/ensure_assets_compiled.rb
+++ b/lib/react_on_rails/ensure_assets_compiled.rb
@@ -1,31 +1,47 @@
 module ReactOnRails
-  module EnsureAssetsCompiled
-    def self.check_built_assets
-      return if @checks_complete
-      puts "Checking for existing webpack bundles before running tests."
-      build_assets_for_type("client")
-      build_assets_for_type("server") if ReactOnRails.configuration.server_bundle_js_file.present?
-      @checks_complete = true
+  class EnsureAssetsCompiled
+    COMPILED_DIR_NAMES = %w(javascripts stylesheets fonts images).freeze
+
+    def self.build
+      client_dir = Rails.root.join("client")
+      compiled_dirs = COMPILED_DIR_NAMES.map { |dir| Rails.root.join("app", "assets", dir, "generated") }
+      checker = WebpackAssetsStatusChecker.new(client_dir: client_dir, compiled_dirs: compiled_dirs)
+      compiler = WebpackAssetsCompiler.new
+      new(checker, compiler)
     end
 
-    def self.build_assets_for_type(type)
-      unless running_webpack_watch?(type)
-        puts "Building Webpack #{type}-rendering assets..."
-        build_output = `cd client && npm run build:#{type}`
-        if build_output =~ /error/i
-          fail "Error in building assets!\n#{build_output}"
-        end
+    attr_reader :webpack_assets_checker, :webpack_assets_compiler, :assets_have_been_compiled
 
-        puts "Webpack #{type}-rendering assets built."
-      end
+    def initialize(webpack_assets_checker, webpack_assets_compiler)
+      @webpack_assets_compiler = webpack_assets_compiler
+      @webpack_assets_checker = webpack_assets_checker
+      @assets_have_been_compiled = false
     end
 
-    def self.running_webpack_watch?(type)
-      running = `pgrep -fl '\\-w \\-\\-config webpack\\.#{type}\\.rails\\.build\\.config\\.js'`
-      if running.present?
-        puts "Webpack is running for #{type}-rendering assets, skipping rebuild => #{running.ai}"
-        return true
+    def call
+      should_skip_compiling = assets_have_been_compiled || @webpack_assets_checker.up_to_date?
+      webpack_assets_compiler.compile unless should_skip_compiling
+      @assets_have_been_compiled = true
+    end
+  end
+
+  class WebpackAssetsCompiler
+    def compile
+      compile_type(:client)
+      compile_type(:server) if ReactOnRails.configuration.server_bundle_js_file.present?
+    end
+
+    private
+
+    def compile_type(type)
+      puts "\n\nBuilding Webpack #{type}-rendering assets..."
+      build_output = `cd client && npm run build:#{type}`
+
+      if build_output =~ /error/i
+        fail "Error in building assets!\n#{build_output}"
       end
+
+      puts "Webpack #{type}-rendering assets built.\n\n"
     end
   end
 end
--- a/lib/react_on_rails/webpack_assets_status_checker.rb
+++ b/lib/react_on_rails/webpack_assets_status_checker.rb
@@ -1,0 +1,45 @@
+require "rake"
+
+module ReactOnRails
+  class WebpackAssetsStatusChecker
+    attr_reader :client_dir, :compiled_dirs
+
+    def initialize(args = {})
+      @compiled_dirs = args.fetch(:compiled_dirs)
+      @client_dir = args.fetch(:client_dir)
+    end
+
+    def up_to_date?
+      return false unless assets_exist?
+      all_compiled_assets.all? { |asset| FileUtils.uptodate?(asset, client_files) }
+    end
+
+    private
+
+    def all_compiled_assets
+      @all_compiled_assets ||= make_file_list(make_globs(compiled_dirs)).to_ary
+    end
+
+    def client_files
+      @client_files ||= make_file_list(make_globs(client_dir)).to_ary
+    end
+
+    def make_globs(dirs)
+      Array(dirs).map { |dir| File.join(dir, "**", "*") }
+    end
+
+    def assets_exist?
+      all_compiled_assets.to_ary.size > 0
+    end
+
+    def make_file_list(glob)
+      FileList.new(glob) do |fl|
+        fl.exclude(".DS_Store")
+        fl.exclude(".keep")
+        fl.exclude("thumbs.db")
+        fl.exclude(".")
+        fl.exclude("..")
+      end
+    end
+  end
+end
--- a/spec/react_on_rails/ensure_assets_compiled_spec.rb
+++ b/spec/react_on_rails/ensure_assets_compiled_spec.rb
@@ -1,0 +1,41 @@
+require_relative "simplecov_helper"
+require_relative "spec_helper"
+
+class WebpackAssetsCompilerDouble
+  attr_reader :times_ran
+
+  def initialize
+    @times_ran = 0
+  end
+
+  def compile
+    @times_ran += 1
+  end
+end
+
+module ReactOnRails
+  describe EnsureAssetsCompiled do
+    let(:compiler) { WebpackAssetsCompilerDouble.new }
+    let(:ensurer) { EnsureAssetsCompiled.new(checker, compiler) }
+
+    context "when assets are not up to date" do
+      let(:checker) { double_webpack_assets_checker(up_to_date: false) }
+
+      it "compiles the webpack bundles" do
+        expect { ensurer.call }.to change { compiler.times_ran }.from(0).to(1)
+      end
+    end
+
+    context "when assets are up to date" do
+      let(:checker) { double_webpack_assets_checker(up_to_date: true) }
+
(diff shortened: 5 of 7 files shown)


robwise:
Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 3 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TK8MXQiRjPnFljW1K:-K9TK8MXQiRjPnFljW1L:-1932651568) ([raw file](https://github.com/shakacode/react_on_rails/blob/11b3a8b02ede9e433f71a5b0521427450a29825e/lib/react_on_rails/ensure_assets_compiled.rb#L3)):*
@justin808 does it make sense that I check all of these folders for generated assets?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

justin808:
If I'm running the static assets, I don't want the check to even run, per the fact that process is running, which is super fast.

What's going to happen if there's a 1000 js files? How fast is this?

New rails uses some file system notification to track changes. We should maybe have some background process that watches and touches some file that says dirty.

But if you're going to do that, you might as well just check for the webpack -w processes.

---

Reviewed 13 of 14 files at r1, 1 of 1 files at r2.

---

*[CHANGELOG.md, line 10 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TOLa6qO7UEPRdz4NY:-K9TOLa6qO7UEPRdz4NZ:-1261125758) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/CHANGELOG.md#L10)):*
files can be out of date, and wepack -w might be in flight

why remove the process check? slow?

---

*[CHANGELOG.md, line 12 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TOSdaPsIFDb2XYSmw:-K9TOSdaPsIFDb2XYSmx:1104750275) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/CHANGELOG.md#L12)):*
These should ultimately be configurable, as should the npm run tasks names.

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 3 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TK8MXQiRjPnFljW1K:-K9TOcebOSiorSfXTcHv:-80086463) ([raw file](https://github.com/shakacode/react_on_rails/blob/11b3a8b02ede9e433f71a5b0521427450a29825e/lib/react_on_rails/ensure_assets_compiled.rb#L3)):*
I don't think fonts and images get generated, but they are dependencies.

I don't understand assets.

this needs to be configurable in the ReactOnRails.rb file

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 41 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TP3m99MBt9P44q6f4:-K9TP3m99MBt9P44q6f5:-1834007503) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/lib/react_on_rails/ensure_assets_compiled.rb#L41)):*
convention over configuration

build:client
build:server

right?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-:-K9TPg9BwvLVjE7703Y6:1923094566)*

robwise:
@justin808 The check is super-fast and short-circuits on first failure. This way we don't rely on `pgrep`ing for a process name that must be exactly correct for it to work. This was an issue on an internal project that slipped through and I was able to fix it only because I knew the internal workings of the code and where to look for what the process had to be named.

---

---

*[CHANGELOG.md, line 10 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TOLa6qO7UEPRdz4NY:-K9TWUrR81KMNYiT4wo_:800313397) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/CHANGELOG.md#L10)):*
@justin808 The process check is brittle and adds complexity for end-users to know what to name the process. The only way that the files are out of date and webpack -w is in flight would be if you are running your tests without waiting for your webpack process to finish compiling, which would cause a failure for either technique.

---

*[CHANGELOG.md, line 12 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TOSdaPsIFDb2XYSmw:-K9TX3aRfkRoXkiLpf0N:-267749782) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/CHANGELOG.md#L12)):*
@justin808 Yeah I agree, these are basically hard-coded folder names and therefore we are making assumptions here, but I vote for a future PR if and when someone has an issue with it

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 3 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TK8MXQiRjPnFljW1K:-K9TXILSEmHtdZSMK5Fo:-1241324534) ([raw file](https://github.com/shakacode/react_on_rails/blob/11b3a8b02ede9e433f71a5b0521427450a29825e/lib/react_on_rails/ensure_assets_compiled.rb#L3)):*
@justin808 `assets` is a bug, good catch, that should be `stylesheets`. Should I delete fonts and images then?

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 41 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TP3m99MBt9P44q6f4:-K9TXWq-x4A2TTiYjL4N:485826380) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/lib/react_on_rails/ensure_assets_compiled.rb#L41)):*
@justin808 Yeah this is the other kind of hard-coded thing I am doing that is making an assumption. We *assume* your static build process is named `build:client` and `build:server` (if you have server-rendering). This would be no big deal to allow for the user to override a default value or whatever, but it's just more and more noise for end-users. Should I implement the overriding and use defaults or something for this as well as the generated folder names?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-:-K9TW3tRS9I0DzA41yGG:1874749325)*

samnang:
*[spec/react_on_rails/support/fixtures_helper.rb, line 3 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VE--YYpJPsrKH9WIS:-K9VE--YYpJPsrKH9WIT:-300928769) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/support/fixtures_helper.rb#L3)):*
Shouldn't it just a constant instead method?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

samnang:
*[spec/react_on_rails/support/fixtures_helper.rb, line 7 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VEQqeTzy2di412Ol1:-K9VEQqgOt9esUMe6VUh:-1033786378) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/support/fixtures_helper.rb#L7)):*
Why don't we just accept just call file_path? And we assume they will be relative path from fixtures directory instead of either support file name and array to join?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

samnang:
*[spec/react_on_rails/webpack_assets_status_checker_spec.rb, line 14 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VSJCM7uMxP7FJJuCv:-K9VSJCNUKqp33Msh0xs:-1680451484) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/webpack_assets_status_checker_spec.rb#L14)):*
No need to take any action. Be careful of overused `let`

https://robots.thoughtbot.com/lets-not

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

samnang:
@robwise I'm not pretty sure, I see couple of files under fixes got changes, did they get change because of `touch`, if so we need to think about the test again. Each test run shouldn't be different.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-:-K9VTM819gnatnZt_wD7:1597633744)*

samnang:
*[lib/react_on_rails/webpack_assets_status_checker.rb, line 41 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VUcs6MczfUNRziyT7:-K9VUcs6MczfUNRziyT8:2114959276) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/lib/react_on_rails/webpack_assets_status_checker.rb#L41)):*
I'm not sure, it might be a bit brittle if they have some other generated files.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

justin808:
Rob, super fast for how many files? 

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 3 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TK8MXQiRjPnFljW1K:-K9Vn4CMd1q-R9u23Eph:-1179446010) ([raw file](https://github.com/shakacode/react_on_rails/blob/11b3a8b02ede9e433f71a5b0521427450a29825e/lib/react_on_rails/ensure_assets_compiled.rb#L3)):*
Probably, because any change to those would be reflected by a change to the stylesheets directory.j

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 41 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TP3m99MBt9P44q6f4:-K9VnIo-aZssPFjT1c41:1117187385) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/lib/react_on_rails/ensure_assets_compiled.rb#L41)):*
Ideally, these values should not be hard coded.

And for production, we have different settings for building.

---

*[spec/react_on_rails/support/fixtures_helper.rb, line 3 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VE--YYpJPsrKH9WIS:-K9Vn_LKM34upXusLa8h:1814048775) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/support/fixtures_helper.rb#L3)):*
could be

---

*[spec/react_on_rails/webpack_assets_status_checker_spec.rb, line 19 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9Vo-3ADuxuDIWpnbhb:-K9Vo-3B8EnqqY2O75UO:-286772437) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/webpack_assets_status_checker_spec.rb#L19)):*
you only need to touch one file, and
you should check each one in separate tests.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-:-K9VoAkMEpm2ZyO-hFGt:-1712250263)*

robwise:
@samnang I was able to fix the `mtime` issue by just using `before` blocks to touch the appropriate files.

@justin808 We are currently using the same technique to check all files in the `lib/generators/react_on_rails` folder recursively, and it's pretty instant.

---

Review status: 13 of 14 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 3 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TK8MXQiRjPnFljW1K:-K9TvSuC51oh0uP48PEh:1465301416) ([raw file](https://github.com/shakacode/react_on_rails/blob/11b3a8b02ede9e433f71a5b0521427450a29825e/lib/react_on_rails/ensure_assets_compiled.rb#L3)):*
Done. I am leaving fonts and images pending your response as it can't hurt for now.

---

*[lib/react_on_rails/webpack_assets_status_checker.rb, line 41 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VUcs6MczfUNRziyT7:-K9YPAqDLZQ6M83chVhm:-1700668522) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/lib/react_on_rails/webpack_assets_status_checker.rb#L41)):*
Huh? These aren't generated files, these are the exceptions that we don't care about

---

*[spec/react_on_rails/support/fixtures_helper.rb, line 3 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VE--YYpJPsrKH9WIS:-K9YPMScQ--XyOWQcyLo:-230059517) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/support/fixtures_helper.rb#L3)):*
I'm in the Avdi Grimm school of methods instead of constants

---

*[spec/react_on_rails/support/fixtures_helper.rb, line 7 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VEQqeTzy2di412Ol1:-K9YPUgwnKT--QNhGS7M:612966483) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/support/fixtures_helper.rb#L7)):*
@samnang It's terser when I call the helper so I don't have to keep doing `File.join` everywhere, it's not DRY

---

*[spec/react_on_rails/webpack_assets_status_checker_spec.rb, line 19 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9Vo-3ADuxuDIWpnbhb:-K9YPmIEMURd2LiVtisP:1028607995) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/webpack_assets_status_checker_spec.rb#L19)):*
@justin808 it checks the whole array, so it would be redundant to test each one. Fixed with touching only one of them.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-:-K9YOjyK6QyxqW-0zWI6:646372146)*

robwise:
Review status: 13 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/ensure_assets_compiled.rb, line 41 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9TP3m99MBt9P44q6f4:-K9YQOgyDhPqJS3-33Da:204136399) ([raw file](https://github.com/shakacode/react_on_rails/blob/ba2e38ee973c3710dc5b30b8cf3bf1f3bc3cf944/lib/react_on_rails/ensure_assets_compiled.rb#L41)):*
@justin808 ....so what's the answer to my question?

---

*[spec/react_on_rails/webpack_assets_status_checker_spec.rb, line 14 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/251#-K9VSJCM7uMxP7FJJuCv:-K9YQVU5HP4sNltYf9M3:-1123401828) ([raw file](https://github.com/shakacode/react_on_rails/blob/f9946d00a39f894db082afb7a59ab35c7bfed657/spec/react_on_rails/webpack_assets_status_checker_spec.rb#L14)):*
@samnang I've seen that thoughtbot article before. My way is more DRY, the only argument against it is readability, which is the same argument you could make about any abstraction to remove duplication. These guys should just use Minitest.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/251)*

PR #252, opened 2016-02-03
Update docs

robwise:
Fixes two minor docs issues

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/252)

### DIFF
--- a/docs/additional_reading/rspec_configuration.md
+++ b/docs/additional_reading/rspec_configuration.md
@@ -9,9 +9,13 @@ RSpec.configure do |config|
   ReactOnRails.configure_rspec_to_compile_assets(config)
 ```
 
-You can pass an RSpec metatag as an optional second parameter to this helper method if you want this helper to run on examples other than where `js: true` (default). The helper will compile webpack files at most once per test run.
+You can pass an RSpec metatag as an optional second parameter to this helper method if you want this helper to run on examples other than where `js: true` (default). The helper will compile webpack files at most once per test run. The helper will not compile the webpack files unless they are out of date (stale).
 
-If you do not want to be slowed down by re-compiling webpack assets from scratch every test run, you can call `npm run build:client` (and `npm run build:server` if doing server rendering) to have webpack recompile these files in the background, which will be much faster. The helper looks for these processes and will abort recompiling if it finds them to be running.
+Please take note of the following:
+- This utility assumes your build tasks are `npm run build:client` and `npm run build:server` and do not have the `--watch` option enabled.
+- By default, the webpack processes look for the `app/assets/javascripts/generated` and `app/assets/stylesheets/generated` folders. If these folders are missing, are empty, or contain files with `mtime`s older than any of the files in your `client` folder, the helper will recompile your assets. You can override this inside of `config/initializers/react_on_rails.rb` by passing an array of filepaths (relative to the root of the app) to the `generated_assets_dirs` configuration option.
+
+If you want to speed up the re-compiling process, you can call `npm run build:dev:client` (and `npm run build:dev:server` if doing server rendering) to have webpack run in "watch" mode and recompile these files in the background, which will be much faster when making incremental changes than compiling from scratch.
 
 If you want to use a testing framework other than RSpec, we suggest you implement similar logic.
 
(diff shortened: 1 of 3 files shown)


docs/additional_reading/rspec_configuration.md
```diff
@@ -9,9 +9,9 @@ RSpec.configure do |config|
...
+You can pass an RSpec metatag as an optional second parameter to this helper method if you want this helper to run on examples other than where `js: true` (default). The helper will compile webpack files at most once per test run. The helper will not compile the webpack files unless they are out of date (stale).
 
-If you do not want to be slowed down by re-compiling webpack assets from scratch every test run, you can call `npm run build:client` (and `npm run build:server` if doing server rendering) to have webpack recompile these files in the background, which will be much faster. The helper looks for these processes and will abort recompiling if it finds them to be running.
+If you want to speed up the re-compiling process, you can call `npm run build:dev:client` (and `npm run build:dev:server` if doing server rendering) to have webpack run in "watch" mode and recompile these files in the background, which will be much faster when making incremental changes than compiling from scratch.
```
justin808, line 18: What do you auto-run? you can't call the watch ones.

We should either:
1. Make this configurable
2. Make sure to print out the task names to use

Also, we need to be sure we're tracking if server rendering is on and not run that one.

docs/additional_reading/rspec_configuration.md
```diff
@@ -9,9 +9,9 @@ RSpec.configure do |config|
   ReactOnRails.configure_rspec_to_compile_assets(config)
 ```
 
-You can pass an RSpec metatag as an optional second parameter to this helper method if you want this helper to run on examples other than where `js: true` (default). The helper will compile webpack files at most once per test run.
+You can pass an RSpec metatag as an optional second parameter to this helper method if you want this helper to run on examples other than where `js: true` (default). The helper will compile webpack files at most once per test run. The helper will not compile the webpack files unless they are out of date (stale).
```
justin808, line 12: the webpack generated bundle files

docs/additional_reading/rspec_configuration.md
justin808, line 18: @robwise ?

justin808:
Looking good. Added some comments.

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/252#-:-K9k0uRNlmv2KKKP4xO1:-1618081089)*

robwise:
Reviewed 1 of 3 files at r1.

---

*[docs/additional_reading/rspec_configuration.md, line 12 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/252#-K9aFAnHA5Y7eqQ_eH3V-r1-12:-K9mP-FAKV3g3wR54xIm:1078130600) ([raw file](https://github.com/shakacode/react_on_rails/blob/c1abf03f4020f088075e9107122ba5e26ab39e82/docs/additional_reading/rspec_configuration.md#L12)):*
@justin808 but it's not just generated bundles, it puts assets as well, right?

---

*[docs/additional_reading/rspec_configuration.md, line 14 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/252#-K9aFAnHA5Y7eqQ_eH3V-r1-14:-K9mPDNaBJgQYifswrE5:1077686531) ([raw file](https://github.com/shakacode/react_on_rails/blob/c1abf03f4020f088075e9107122ba5e26ab39e82/docs/additional_reading/rspec_configuration.md#L14)):*
@justin808 Sorry I thought this was clear from https://github.com/shakacode/react_on_rails/pull/253.

Yes I am checking for server-rendering. As of https://github.com/shakacode/react_on_rails/pull/253 I am doing this via a new utility method that essentially just checks to see if server-bundle-js file is not blank in configuration.

As far as the scripts we use to generate the bundles, it's still exactly the same as your original code: https://github.com/shakacode/react_on_rails/pull/253/files#diff-337396057535e726020f7b53e9f6de7dR52

The webpack build `npm run build:client` and if server rendering is enabled , `npm run build:server`

I have added a note to this doc about what script it is running, although I guarantee that if we change the name of the build scripts in the future we will forget to update this. The task will fail if these scripts don't exist and we log the failure output so it should be obvious to the user.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/252)*

robwise:
This PR goes along with #253

justin808:
This is in #253.

PR #262, opened 2016-02-08
Fix Bootstrap Sass Append to Gemfile

### LINKED ISSUE #256: append to Gemfile
in my case Gemfile had
```ruby
group :test do
  #...
end
```
And what i've got after running generator is:
```ruby
group :test do
  #...
endgem 'botstrap-saas'
```

aaronvb:
Fixes #256 

If the gemfile does not have an empty line at the end, append_to_file will append ‘gem bootstrap-sass’ to the last line.

- Added a line break to the beginning of ‘gem bootstrap-sass’ to ensure it’s on a new line.

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/262)

### DIFF
--- a/lib/generators/react_on_rails/bootstrap_generator.rb
+++ b/lib/generators/react_on_rails/bootstrap_generator.rb
@@ -62,7 +62,7 @@ module ReactOnRails
       end
 
       def add_bootstrap_sprockets_to_gemfile
-        append_to_file("Gemfile", "gem 'bootstrap-sass'\n")
+        append_to_file("Gemfile", "\ngem 'bootstrap-sass'\n")
       end
 
       def add_bootstrap_sprockets_to_application_js


lib/generators/react_on_rails/bootstrap_generator.rb
```diff
@@ -62,7 +62,7 @@ def strip_application_scss_of_incompatible_sprockets_statements
       end
 
       def add_bootstrap_sprockets_to_gemfile
-        append_to_file("Gemfile", "gem 'bootstrap-sass'\n")
+        append_to_file("Gemfile", "\ngem 'bootstrap-sass'\n")
```
aaronvb, line 65: Following the same convention we used here: https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/base_generator.rb#L137

justin808:
Thanks @aaronvb!

PR #264, opened 2016-02-10
Turbolinks 5 Support

aaronvb:
- Added turbolinksShim to support older versions of Turbolinks.
- Updated dummy to use Turbolinks 5

***

I came up with two options to getting Turbolinks 5 in; update the event listeners or add a shim so we can have support for both the classic and Turbolink 5 events. I decided on the latter, based on the one in the Turbolinks 5 docs, that translates the classic events to the new Turbolinks 5 events giving support for classic and Turbolinks 5.

https://github.com/turbolinks/turbolinks#should-i-use-turbolinks-5-or-turbolinks-classic
https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/compatibility.coffee

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/264)


justin808:
@aaronvb my gut tells me that we want to detect which version of turbolinks is in use, rather than using a shim. Did you have a PR for doing the non-shim way?

aaronvb:
@justin808 That was my initial idea but I haven't found an easy way to detect the javascript version of Turbolinks. I can see there are completely new functions between classic and 5, and I can do a comparison with that, but that seems hacky. Do you know of a way to figure out the version of Turbolinks loaded in the DOM?

aaronvb:
@justin808 Forgot to note that the shim is based on https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/compatibility.coffee

justin808:
You should be able to determine this in 2 ways:

1. On the ruby side, check the VERSION in the gem
2. On the JS side, there's probably some function names on the Turbolinks object that differ between v2 and v5.

aaronvb:
@justin808 since this is being handled in the npm_module I'll have to use the second option. I'll make another PR with the version check method. Thanks for the feedback!

justin808:
See https://github.com/turbolinks/turbolinks/issues/6

justin808:
Closed in favor of #270

PR #270, opened 2016-02-11
Add Turbolinks 5 Support Version Check

aaronvb:
- Add version check by using feature detection.
- Update dummy app to use Turbolinks 5

***

Alternate method to #264

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/270)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ install:
   - npm install -g npm
   - npm install -g poltergeist
   - bundle install
+  - ENABLE_TURBOLINKS_5=TRUE bundle install
   - npm install
   - rake dummy_apps
   - rake examples
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,10 @@ gem "selenium-webdriver"
 gem "spring"
 gem "sqlite3"
 gem "therubyracer"
-gem "turbolinks"
+if ENV["ENABLE_TURBOLINKS_5"].nil? || ENV["ENABLE_TURBOLINKS_5"].strip.empty?
+  gem "turbolinks", "2.5.3"
+else
+  gem "turbolinks", "~> 5.0.0.beta"
+end
 gem "uglifier", ">= 2.7.2"
 gem "web-console", "~> 2.0"
--- a/docs/additional_reading/turbolinks.md
+++ b/docs/additional_reading/turbolinks.md
@@ -1,7 +1,7 @@
 # Turbolinks
 
 * See [Turbolinks on Github](https://github.com/rails/turbolinks)
-* Currently support 2.5.x of Turbolinks. We plan to update to Turbolinks 5 soon.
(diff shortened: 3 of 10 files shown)


justin808:
@aaronvb this looks great

Couple things:

1. Update the 

https://github.com/shakacode/react_on_rails/blob/master/docs%2Fadditional_reading%2Fturbolinks.md

with some info on this.


2. Can we have travis run both? I'm pretty sure we can with 2 env setups (@dylangrafmyre)


I'd like to get this into v3. I'll merge tomorrow once I hear back.

aaronvb:
@justin808 I couldn't get rspec to recognize two different versions of turbolinks because it seems to always default back to the turbolinks in the root Gemfile. 

I tried using an ENV variable wrapped around the gem include and having a new rake task invoke it, but that didn't work.

```ruby
if ENV["DISABLE_TURBOLINKS"].nil? || ENV["DISABLE_TURBOLINKS"].strip.empty?
  if ENV["TURBOLINKS_5"].nil? || ENV["TURBOLINKS_5"].strip.empty?
    gem 'turbolinks', '2.5.3'
  else
   gem 'turbolinks', '~> 5.0.0.beta' 
  end
end
```

I also tried setting up a 2nd dummy app with turbolinks 5 in the gemfile and it would still install the turbolinks from the root Gemfile. I know @robwise wrote something about the gems in the root Gemfile being used for the dummy app and travis-ci, maybe he can shed some light here.

aaronvb:
Update: I may have found a way test both versions of Turbolinks. Running on CI right now.

.travis.yml
```diff
@@ -26,6 +26,7 @@ install:
   - npm install -g npm
   - npm install -g poltergeist
   - bundle install
+  - ENABLE_TURBOLINKS_5=TRUE bundle install
```
aaronvb, line 29: Any gems that are used in spec/dummy need to be installed before running the tests because Travis-CI will look for the source here when running bundle install in the spec/dummy folder. This basically tells travis-ci to install turbolinks classic and turbolinks 5.

Gemfile
```diff
@@ -29,6 +29,10 @@ gem "selenium-webdriver"
 gem "spring"
 gem "sqlite3"
 gem "therubyracer"
-gem "turbolinks"
+if ENV["ENABLE_TURBOLINKS_5"].nil? || ENV["ENABLE_TURBOLINKS_5"].strip.empty?
```
aaronvb, line 32: This environment variable is passed in the travis-ci config.

aaronvb:
Working on doc update.

justin808:
:lgtm_strong: 

:clap: :clap: :clap: :clap: :clap: :clap: :clap: 

---

Reviewed 7 of 8 files at r1, 3 of 3 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/270#-:-KAUAmJ4qYBDPsKs9sVy:-1991634907)*

PR #284, opened 2016-02-19
fix build comment in generator

ttrmw:
Pretty sure this comment was meant to include `:client`?

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/284)


aaronvb:
@ttrmw You are correct, `npm run build:dev` is not a valid run command. The client webpack should be run with `npm run build:dev:client`.

justin808:
Thanks @ttrmw

PR #315, opened 2016-03-07
expose reactOnRailsPageLoaded to ReactOnRails

SqueezedLight:
[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/315)

### DIFF
--- a/docs/additional-reading/turbolinks.md
+++ b/docs/additional-reading/turbolinks.md
@@ -28,6 +28,24 @@ Turbolinks 5 is now being supported. React on Rails will automatically detect wh
 
 For more information on Turbolinks 5: [https://github.com/turbolinks/turbolinks](https://github.com/turbolinks/turbolinks)
 
+### async script loading
+Generally async script loading can be done like:
+```erb
+  <%= javascript_include_tag 'application', async: Rails.env.production? %>
+```
+If you use ```document.addEventListener("turbolinks:load", function() {...});``` somewhere in your code, you will notice, that Turbolinks 5 does not fire ```turbolinks:load``` on initial page load. A quick workaround is to use ```defer``` instead of ```async```:
+```erb
+  <%= javascript_include_tag 'application', defer: Rails.env.production? %>
+```
+More information on this issue can be found here: https://github.com/turbolinks/turbolinks/issues/28
+
+When loading your scripts asynchronously you may experience, that your Components are not registered correctly. Call ```ReactOnRails.reactOnRailsPageLoaded()``` to re-initialize like so:
+```
+  document.addEventListener("turbolinks:load", function() {
+    ReactOnRails.reactOnRailsPageLoaded();
+  });
+```
+
 ## Troubleshooting
 To turn on tracing of Turbolinks events, put this in your registration file, where you register your components.
 
--- a/node_package/src/ReactOnRails.js
+++ b/node_package/src/ReactOnRails.js
@@ -1,4 +1,4 @@
-import clientStartup from './clientStartup';
+import * as ClientStartup from './clientStartup';
 import handleError from './handleError';
 import ComponentRegistry from './ComponentRegistry';
 import StoreRegistry from './StoreRegistry';
@@ -61,6 +61,10 @@ ctx.ReactOnRails = {
     if (Object.keys(options).length > 0) {
       throw new Error('Invalid options passed to ReactOnRails.options: ', JSON.stringify(options));
     }
+  },
+
+  reactOnRailsPageLoaded() {
+    ClientStartup.reactOnRailsPageLoaded();
   },
 
   ////////////////////////////////////////////////////////////////////////////////
@@ -175,6 +179,6 @@ ctx.ReactOnRails = {
 
 ReactOnRails.resetOptions();
 
-clientStartup(ctx);
+ClientStartup.clientStartup(ctx);
 
 export default ctx.ReactOnRails;
--- a/node_package/src/clientStartup.js
+++ b/node_package/src/clientStartup.js
@@ -85,7 +85,7 @@ You should return a React.Component always for the client side entry point.`);
   }
 }
 
-function reactOnRailsPageLoaded() {
+export function reactOnRailsPageLoaded() {
   debugTurbolinks('reactOnRailsPageLoaded');
 
   forEachStore(initializeStore);
@@ -103,7 +103,7 @@ function reactOnRailsPageUnloaded() {
   forEachComponent(unmount);
 }
 
-export default function clientStartup(context) {
+export function clientStartup(context) {
   const document = context.document;
 
   // Check if server rendering


justin808:
Looking great!

---

Reviewed 4 of 4 files at r1.

---

*[docs/additional_reading/turbolinks.md, line 39 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJaadr99Rec3bZPSkT:-KCJaadsx3Tp3TUWN5VJ:1559906488) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L39)):*
Is this an issue for TurboLinks 2.5? We need to indicate that.

---

*[docs/additional_reading/turbolinks.md, line 46 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJb77zTz6Mqw5pLj0a:-KCJb77zTz6Mqw5pLj0b:1005888068) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L46)):*
I'm a little confused by this. I'm guessing that this is 100% needed, or not, and we should change https://github.com/shakacode/react_on_rails/blob/master/node_package%2Fsrc%2FclientStartup.js#L146

---

*[node_package/src/ReactOnRails.js, line 64 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJbyGIy8dN5nibVdlG:-KCJbyGIy8dN5nibVdlH:1548518597) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/node_package/src/ReactOnRails.js#L64)):*
While this probably works, it's confusing that the inner name matches the outer name. Let's change the import name for `reactOnRailsPageLoaded` to 

```js
import * as ClientStartup from './clientStartup';
```

And then we can use

```js
ClientStartup.clientStartup()
ClientStartup.reactOnRailsPageLoaded()
```

Any suggestions?

---

*[spec/dummy/Gemfile.lock, line 75 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJdih8AGPFanspaiNE:-KCJdih8AGPFanspaiNF:-1326771124) ([raw file](https://github.com/shakacode/react_on_rails/blob/08fe39b5cde9ad6190d3698a6183a1f5a132e94b/spec/dummy/Gemfile.lock#L75)):*
why are you removing this?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-:-KCJdoc2VGH2HuCvbpVh:-781840010)*

justin808:

---

*[docs/additional_reading/turbolinks.md, line 46 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJb77zTz6Mqw5pLj0a:-KCJdybDBI_xR85xYPXd:-1602905941) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L46)):*
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/315)*

SqueezedLight:

---

*[docs/additional_reading/turbolinks.md, line 39 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJaadr99Rec3bZPSkT:-KCKSeAyhCYs_VLVzkGv:-1846851064) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L39)):*
You mean the workaround to use defer instead of async? In previous versions of my app i used Turbolinks 2.5 and it worked with async. The page:load event in Turbolinks 2.5 does not fire on initial page load anyway. In contrast turbolinks:load is meant to fire on initial page load, but does not in case of loading with async. So the defer workaround helps. Maybe it is a good idea to mention here that Turbolinks 5 is a complete rebuild and events changed a lot - with reference to the Turbolinks 5 page?

---

*[docs/additional_reading/turbolinks.md, line 46 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJb77zTz6Mqw5pLj0a:-KCKUPDOniYFRMP7uUMc:185116400) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L46)):*
I didn't test it for a lot of different use cases and do not know if it is needed 100%. But at least it is needed in case of using third-party libraries that interact with ReactOnRails components. What would you suggest?

---

*[spec/dummy/Gemfile.lock, line 75 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJdih8AGPFanspaiNE:-KCKRx_56bBPTXAxUkBy:-460158724) ([raw file](https://github.com/shakacode/react_on_rails/blob/08fe39b5cde9ad6190d3698a6183a1f5a132e94b/spec/dummy/Gemfile.lock#L75)):*
I'm having problems with bundling 'capybara-webkit' in my dev env. Thats why i removed it from the spec/dummy/gemfile. I accidentally pushed this change, sorry.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/315)*

justin808:
Reviewed 2 of 2 files at r2.

---

*[node_package/src/ReactOnRails.js, line 64 \[r2\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCMUuXgwO6LnDzf1o51:-KCMUuXgwO6LnDzf1o52:1052976030) ([raw file](https://github.com/shakacode/react_on_rails/blob/9621a1636bde350351b93b5eb0e2d2509f6e7b54/node_package/src/ReactOnRails.js#L64)):*
Yes, better! I hope you agree!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/315)*

justin808:
Let's address the "can we change the clientStartup method" to solve so of the TurboLinks issues. Thanks!

---

---

*[docs/additional_reading/turbolinks.md, line 39 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJaadr99Rec3bZPSkT:-KCMV7bluw0KS-SYFW_9:1569950577) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L39)):*
Yes -- the more details on this the better.

However I think we should be supporting the turbolinks part without the need to have the extra code. The problem is that the library won't know (at least I think not) if the script was loaded async or defer.

---

*[docs/additional_reading/turbolinks.md, line 46 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-KCJb77zTz6Mqw5pLj0a:-KCMVTJFLOv37prSjE-g:528732367) ([raw file](https://github.com/shakacode/react_on_rails/blob/40c97e23e10a6e48ce334a34fc35f3e6c7083fde/docs/additional_reading/turbolinks.md#L46)):*
We need to do a little more research into this. I'm going to also be investigating Turbolinks5 soon.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/315#-:-KCMVd63CVIyv2vRRvXw:1554320718)*

justin808:
@SqueezedLight I'm going to try looking into this issue over the weekend. If you have any additional info, please let me know.

justin808:
@SqueezedLight Please rebase on top of master to a single commit and let me know if you're ready. I'm about to release v4.

Guoxweii:
+ 1

SqueezedLight:
@justin808 sorry for the delay again. All done!

justin808:
Given: 


I'm still not totally sure on the docs for this feature.

I'm concerned that somebody will be running the loaded handler code twice in some cases.

Thoughts?

SqueezedLight:
In the docs i wrote: "When loading your scripts asynchronously you may experience, that your Components are not registered correctly. Call ```ReactOnRails.reactOnRailsPageLoaded()``` to re-initialize like so:"

If i understand you correctly you insist on making it much more clear that this is optional and only in case components are not registered correctly, right? 
I see the explanation in the docs more as a hint in case someone runs into the same problem i did (components are not registered). The problem itself is not limited to using Turbolinks or loading scripts via async or defer. So i guess the right place to document this would be a specific doc file where all methods of the ReactOnRails object are documented and ```reactOnRailsPageLoaded()``` is just one of them. Right now there is no such file, is it?? Thats why i added it to the turbolinks docs, mainly because i thought this is the place where most people would run into this problem. Beside that, i referenced the issue i opened at the turbolinks repo.

To be honest i'm feeling like i do not really get the severeness of your concerns ;) What i do know is, that without ```ReactOnRails.reactOnRailsPageLoaded()``` i had to dump this gem in production apps. It is that important for me and i guess there are more people with this need. As i said, i have the slight feeling i do not get your concerns right. If i'm mistaken i'm happy to learn ;)

justin808:
> In the docs i wrote: "When loading your scripts asynchronously you may experience, that your Components are not registered correctly. Call ReactOnRails.reactOnRailsPageLoaded() to re-initialize like so:"
> If i understand you correctly you insist on making it much more clear that this is optional and only in case components are not registered correctly, right? 

It's not about "registration" but script loading settings and Turbolinks, which I'm researching heavily this week. Possibly the version of Turbolinks matters (2.x or 5.x)

> I see the explanation in the docs more as a hint in case someone runs into the same problem i did (components are not registered). The problem itself is not limited to using Turbolinks or loading scripts via async or defer. So i guess the right place to document this would be a specific doc file where all methods of the ReactOnRails object are documented and reactOnRailsPageLoaded() is just one of them. Right now there is no such file, is it?? Thats why i added it to the turbolinks docs, mainly because i thought this is the place where most people would run into this problem. Beside that, i referenced the issue i opened at the turbolinks repo.

That's good. I'm going to take the API docs, ruby and JS, and put them into separate files, very soon.

> To be honest i'm feeling like i do not really get the severeness of your concerns ;) What i do know is, that without ReactOnRails.reactOnRailsPageLoaded() i had to dump this gem in production apps. It is that important for me and i guess there are more people with this need. As i said, i have the slight feeling i do not get your concerns right. If i'm mistaken i'm happy to learn ;)

Basically, the gem has gotten pretty complicated. For every new option we add, a newbie must consider "do I need to use that option?". One of the goals of ReactOnRails is that the component loading should just work. 

Basically, I can handle the remaining to-do's (but help is always appreciated):

1. Expose this, as a means to work around these issues.
1. Research any special cases that require this, seeing if we can modify the file. [clientStartup.js](https://github.com/shakacode/react_on_rails/blob/master/node_package%2Fsrc%2FclientStartup.js) to accomdate for these cases. @SqueezedLight can you provide me a sample git repo that I can clone for this purpose? Or just details are great on how I modify the /spec/dummy sample app.

justin808:
@jbhatab @alexfedoseev @thewoolleyman @robwise Agree?

justin808:
@SqueezedLight Any word?

I really want to accept this, but I'm mostly concerned that we'll try to double call the render sometimes.

This code runs when the JS from the bundle is loaded. It installs the handler.

I'm thinking that if we install turbolinks via npm in the client bundles, then we won't need this.

https://github.com/shakacode/react_on_rails/blob/master/node_package%2Fsrc%2FclientStartup.js#L106

```js
export default function clientStartup(context) {
  const document = context.document;

  // Check if server rendering
  if (!document) {
    return;
  }

  // Tried with a file local variable, but the install handler gets called twice.
  if (context.__REACT_ON_RAILS_EVENT_HANDLERS_RAN_ONCE__) {
    return;
  }

  context.__REACT_ON_RAILS_EVENT_HANDLERS_RAN_ONCE__ = // eslint-disable-line no-param-reassign
    true;

  debugTurbolinks('Adding DOMContentLoaded event to install event listeners.');

  document.addEventListener('DOMContentLoaded', () => {
    // Install listeners when running on the client (browser).
    // We must do this check for turbolinks AFTER the document is loaded because we load the
    // Webpack bundles first.

    if (!turbolinksInstalled()) {
      debugTurbolinks(
        'NOT USING TURBOLINKS: DOMContentLoaded event, calling reactOnRailsPageLoaded'
      );
      reactOnRailsPageLoaded();
    } else {
      if (turbolinksVersion5()) {
        debugTurbolinks(
          'USING TURBOLINKS 5: document added event listeners turbolinks:before-cache and ' +
          'turbolinks:load.'
        );
        document.addEventListener('turbolinks:before-cache', reactOnRailsPageUnloaded);
        document.addEventListener('turbolinks:load', reactOnRailsPageLoaded);
      } else {
        debugTurbolinks(
          'USING TURBOLINKS 2: document added event listeners page:before-unload and ' +
          'page:change.');
        document.addEventListener('page:before-unload', reactOnRailsPageUnloaded);
        document.addEventListener('page:change', reactOnRailsPageLoaded);
      }
    }
  });
}
```

justin808:
@SqueezedLight I'm going to put a new release out. We need to rebase this on top of master make sure it passes CI. Any thoughts on my prior questions?

SqueezedLight:
@justin808 oh man i totally lost track of this issue. I'm under very heavy work load at the moment.

To be honest i'm more or less unable to make a qualified statement to your questions. A lot has changed in this gem and the version i run is not the current one (3.0.6). I tried to update my code several times in the last 4 weeks - but it completely breaks a lot of (well established) things - asset loading for example. The gaps between releases are huge. Too huge for me to follow on the side. That is why i'm not in the position to make a helpful statement.
On top of that it seems i can not get a fresh master copy to work. I'm afraid i'm out of this game as long as i find more time to debug.

Sorry for not being more of a help.

justin808:
@SqueezedLight I'm going to look at this more in the coming week.

justin808:
@SqueezedLight I should have time in the next few days to check this one out. If you can rebase on top of master, that would be appreciated!

justin808:
Closed as it's merged in #417

PR #319, opened 2016-03-08
Remove use of heroku-buildpack-multi

esauter5:
I was trying to use `heroku-buildpack-multi` and was getting failures when trying to push to Heroku. This led me to realize that Heroku natively supports using multiple build packs now, so there is no need to use `heroku-buildpack-multi` or have to worry about a `.buildpack` file. I simplified the instructions and removed references to `.buildpack` in the code. Now, all you have to worry about is running this from the CLI:

```
heroku buildpacks:set heroku/ruby
heroku buildpacks:add --index 1 heroku/nodejs
```

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/319)

### DIFF
--- a/docs/additional_reading/heroku_deployment.md
+++ b/docs/additional_reading/heroku_deployment.md
@@ -2,7 +2,6 @@
 The generator has created the necessary files and gems for deployment to Heroku. If you have installed manually, you will need to provide these files yourself:
 
 + `Procfile`: used by Heroku and Foreman to start the Puma server
-+ `.buildpacks`: used to install Ruby and Node environments
 + `12factor` gem: required by Heroku
 + `'puma'` gem: recommended Heroku webserver
 + `config/puma.rb`: Puma webserver config file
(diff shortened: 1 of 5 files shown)


esauter5:
I'm now noticing that the docs previously described that you could use `heroku-buildpack-multi` or set up multiple buildpacks via the toolbelt. I actually mistakenly merged these ideas together in a previous update on the docs. I still agree it's best to just remove `heroku-buildpack-multi` in general and have one simple way to set this up. Thoughts?

justin808:
@esauter5 This looks great! Can you find one other person in our slack room to confirm these instructions work? Thanks!

stevesuh:
When you run those CLI commands does the .buildpack file get updated?  I have an existing .buildpack file with dependenices to heroku-buildpack-apt and heroku-bundle-config.

esauter5:
AFAIK, the `.buildpack` file is only used by `heroku-buildpack-multi`. In this case, I'm suggesting removing that. Then you just run those two commands and the buildpacks are updated on heroku. You don't need to worry about `.buildpack`.

This brings an interesting point though. I guess I hadn't thought about the possibility that people still might need other buildpacks. May need to update the docs a little more to detail this option

justin808:
@esauter5 Keep me posted.

esauter5:
@justin808 Will do. Hope to get to this in the next day or so.

esauter5:
This should be good. I added something at the end of the heroku deployment section pointing to heroku-buildpack-multi in case people need a custom buildpack option.

docs/additional_reading/heroku_deployment.md
```diff
@@ -10,18 +9,19 @@ The generator has created the necessary files and gems for deployment to Heroku.
...
-heroku buildpacks:set https://github.com/heroku/heroku-buildpack-multi
+heroku buildpacks:set heroku/ruby
+heroku buildpacks:add --index 1 heroku/nodejs
 ```
 
-Heroku will now be able to use the multiple buildpacks specified in `.buildpacks`.
-
 For more information, see [Using Multiple Buildpacks for an App](https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app)
 
+If for some reason you need custom buildpacks that are not officially supported by Heroku ([see this page](https://devcenter.heroku.com/articles/buildpacks)), we recommend checking out [heroku-buildpack-multi](https://github.com/ddollar/heroku-buildpack-multi).
```
justin808, line 23: I think this is the one you want: https://github.com/heroku/heroku-buildpack-multi

justin808:
This looks good. One suggestion added.

esauter5:
The old docs recommended and linked to ddollar's version, but then used the heroku one in an example. The heroku version is just a fork of ddollar's and appears to be deprecated according to [https://github.com/heroku/heroku-buildpack-multi/pull/7#issuecomment-151463964](comment).

I tried using heroku's previous and was not able to get it to work. I'm open to linking to whichever though.

justin808:
@esauter5 Great catch! Yes, just last fall, the strategy changed. And the ddollar one has been more recently updated.

lib/generators/react_on_rails/heroku_deployment_generator.rb
```diff
@@ -10,8 +10,7 @@ class HerokuDeploymentGenerator < Rails::Generators::Base
 
       def copy_heroku_deployment_files
         base_path = "heroku_deployment"
-        %w(.buildpacks
-           Procfile
+        %w(Procfile
```
justin808, line 13: we need to delete the .buildpacks file also.

justin808:
Let's:

1. rebase on top of master
2. remove .buildpacks

Then I'll release the next version!

Great job @esauter5!

lib/generators/react_on_rails/heroku_deployment_generator.rb
esauter5, line 13: Ahh good catch! That file was hiding from me because it was a dot file.

esauter5:
Thanks for the help @justin808! This should be good.

justin808:
:100: 

:lgtm_strong: 

---

Reviewed 2 of 4 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/319#-:-KD0i1NHIVW6X94kh5xE:647196688)*

PR #321, opened 2016-03-09
Remove the babel package from the devDependencies

lucke84:
Starting with a freshly installed copy of this project's repo, `npm run build` fails trying to install `babel` as the package no longer exists (see http://jamesknelson.com/the-six-things-you-need-to-know-about-babel-6/). Removing it solves the problem.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/321)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "doc": "docs"
   },
   "devDependencies": {
-    "babel": "^6.3.26",
     "babel-cli": "^6.4.5",
     "babel-core": "^6.4.5",
     "babel-eslint": "^5.0.0-beta8",


justin808:
@lucke84 Any chance that you can look at my branch: 

https://github.com/shakacode/react_on_rails/tree/new-api-shared-redux-stores

I'd like to get this merged to master, but CI is not passing. I fixed many issues there already.

I think somehow the version of babel got out of sync..

justin808:
@lucke84 Can you also please take a look at this CI failure: https://travis-ci.org/shakacode/react_on_rails/builds/115007962

justin808:
@lucke84 I'll back this out tomorrow if we can't get CI to pass. I'll probably have some time this weekend to research. If you can find some fix before, that would be amazing.

lucke84:
Took a quick look, maybe it's worth trying and upgrade node to `5.6.0`?

PR #335, opened 2016-03-17
Changed ReactOnRails::Controller#redux_store API

justin808:
2nd parameter of props with default value of { } is changed to a named
parameter props with a default value of { }

Needed for consistency with the view helper redux_store API.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/335)


justin808:
@seoyoochan Please review!

Issue #317, opened 2016-03-07
Configure Server bundle file as param in react_component


lucke84:
Which are the implications of changing the path of server bundle file (or any other config) at runtime? 

To give you a bit of context: with Webpack we create 2 different server bundle files, one for desktop and the other one for mobile. Then in the `ApplicationController` we have a `before_action` hook to check whether server-side rendering (SSR) is turned on and which browser is requesting the page (via `browser` gem). A simplified version looks like this:

```ruby
class ApplicationController < ActionController::Base
  before_action :set_assets

  private

  def is_mobile?
    browser.mobile?
  end

  def set_assets
    if SERVER_SIDE_RENDERING
      ReactOnRails.configure do |config|
        bundle_file = is_mobile? ? SERVER_BUNDLE_MOBILE : SERVER_BUNDLE_DESKTOP
        Rails.logger.info "Bundle: [#{bundle_file}]"

        config.server_bundle_js_file = bundle_file
        config.prerender = SERVER_SIDE_RENDERING
        config.raise_on_prerender_error = Rails.env.development?
      end
    end
  end
end
```

where `SERVER_SIDE_RENDERING` is a boolean, and `SERVER_BUNDLE_MOBILE` and `SERVER_BUNDLE_DESKTOP` are path to the files; all these constants are properly set in their own initializers.

This works relatively smoothly, but every now and then the user on a mobile device sees the desktop SSR version for a second (while the assets get downloaded) and then the client side rendering kicks in showing the correct version. This never happens in development yet it does in production, therefore I'm thinking this might be due to some Rails classes reloading magic disabled in prod environment.

Anyhow, I wanted to hear somebody else's opinion on whether this is a good idea and which are the shortcomings you can think of.

justin808:
@lucke84 this is an excellent use case. Without a great deal of work, we could support a first class concept of different bundles for server rendering. We could do something like:

```erb
<%= react_component("MobileEnabledComponent", prerender: true, server_bundle: my_rails_helper_to_get_correct_bundle_name, props: @the_props) %>
```

Note, with v4, you will want to hydrate your props in the controller and have it render on the bottom of the page for maximum performance.

If this API works, either you or I could put this.

justin808:
@lucke84 This would be an excellent, EASY way to make a nice contribution.

lucke84:
Will be happy to :) I'll try and put together a PR somewhen this week.

justin808:
@lucke84 Please remember:

[...]

Issue #183, opened 2016-01-04
Consider integration with react-stdio for server rendering


justin808:
* @mjackson: https://github.com/mjackson/react-stdio
* @hassox: http://blog.overstuffedgorilla.com/render-react-with-phoenix/

@mjackson has created react-stdio. We should look into integrating this with react_on_rails.

@alexfedoseev: this looks promising 
@cowboyd: Any thoughts on this compared to what we currently use: https://github.com/cowboyd/therubyracer

mjackson:
Just for clarification, it's react-stdio, not react-studio :)

justin808:
@mjackson Thanks. Studio sounds kind of cool ;-).

Any thoughts on a Rails integration?

mjackson:
Probably the first step would be to create a Ruby wrapper for react-stdio that knows how to feed its stdin and parse its stdout. After that, integration with Rails should be fairly straightforward.

aaronvb:
I looked into implementing this into react_on_rails(and in react-rails). With our current server rendering method it's not going to work 'out of the box'. react-stdio takes a path to a component whereas we pass a string of JS code(with polyfills) to ExecJS(therubyracer).

I ended up writing a little react-stdio wrapper for Rails that could be helpful in pursuing this option: https://github.com/aaronvb/rails_react_stdio.

justin808:
@aaronvb If you get this working, you will be a

# HERO

justin808:
@aaronvb any word on this one?

aaronvb:
@justin808 I haven't put much work into this. I'm open to pairing up or starting a branch for collaboration on this.

justin808:
@aaronvb Any progress on this one? We really should try this one!

justin808:
Closed by #380

PR #343, opened 2016-03-22
Support for multiple server file bundles

lucke84:
Adds support for multiple server bundle js files, as a follow-up of this conversation: https://github.com/shakacode/react_on_rails/issues/317

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/343)


lucke84:
Please take a look at the PR: there's still some testing, clean-up and documentation work to do, but we would like to have a first round of feedback on:
* the approach, in general (including the pool-resetting triggering mechanism for development mode)
* the testing part (what would you add/change)
* the defaulting assumption (the first entry in the array on bundle js file is the default one, when not specified)
* miscellaneous (we tried to stick with the project's best practices and conventions, we might have missed something)

Looking forward to seeing this bit of functionality supported, we (and probably a lot of others) will benefit from this mechanism.

justin808:
Looks like a great start.

Let me know if you have questions.

---

Reviewed 4 of 8 files at r1, 22 of 22 files at r3.

---

*[README.md, line 225 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVSGqQzrx0R1wSik-p:-KDVSGqR8AWzibawf-21:1927103568) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d50159371b69042a430f75899802df601929c32/README.md#L225)):*
Why did you change this?

---

*[app/helpers/react_on_rails_helper.rb, line 87 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVSPg2qxBa-dLo9lZZ:-KDVSPg2qxBa-dLo9lZ_:205328007) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/app/helpers/react_on_rails_helper.rb#L87)):*
```
#   server_bundle_js_file: the name of the server bundle js file you want to use for rendering the component server side.
# If this is not specified, the default file specified in your config/initializers/react_on_rails.rb is used.
```

---

*[lib/generators/react_on_rails/base_generator.rb, line 150 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVSrZOWuRww9oZZiTW:-KDVSrZPMdfbG7V3CFEr:-2002164684) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d50159371b69042a430f75899802df601929c32/lib/generators/react_on_rails/base_generator.rb#L150)):*
`server.jsx` is too generic

---

*[lib/generators/react_on_rails/templates/base/base/config/initializers/react_on_rails.rb.tt, line 18 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVTBkIksv8pgZT3-UR:-KDVTBkIksv8pgZT3-US:-921442811) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/generators/react_on_rails/templates/base/base/config/initializers/react_on_rails.rb.tt#L18)):*
I'm not 100% sure we want to change the config for this. It's an easy migration, but I'm not sure it's common enough to have multiple files.

---

*[lib/react_on_rails/configuration.rb, line 13 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVTeVX1XuFP5CFcBmK:-KDVTeVX1XuFP5CFcBmL:-21217580) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/configuration.rb#L13)):*
```ruby
files +=  @configuration.server_bundle_js_files
```

That's what I normally do, but either would work. Use parens if you use a method call.

---

*[lib/react_on_rails/configuration.rb, line 43 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVUBkE17Y04yWiFWpu:-KDVUBkE17Y04yWiFWpv:-1382863365) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/configuration.rb#L43)):*
```ruby
def normalize_server_bundle_js_files
 @configuration.server_bundle_js_files = @configuration.server_bundle_js_files.map do |server_bundle_js_file|
      if server_bundle_js_file.include?(File::SEPARATOR)
        puts "[DEPRECATION] ReactOnRails: remove path from server_bundle_js_files in configuration. "\
          "All generated files must go in #{@configuration.generated_assets_dir}"
        server_bundle_js_file = File.basename(server_bundle_js_file)
      else
        server_bundle_js_file 
      end
    end
end
```

---

*[lib/react_on_rails/server_rendering_pool.rb, line 16 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVUk-WhAs-ub2FPUk2:-KDVUk-WhAs-ub2FPUk3:1974138719) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/server_rendering_pool.rb#L16)):*
```ruby
  @js_context_pools = 
     ReactOnRails.configuration.server_bundle_js_files.each_with_object({}) do |server_bundle_js_file, hash|
        hash[server_bundle_js_file] = ConnectionPool.new(options) do
          create_js_context(server_bundle_js_file)
        end
      end
```

---

*[lib/react_on_rails/server_rendering_pool.rb, line 74 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVVrQtyoWhx6vXm2mN:-KDVVrQtyoWhx6vXm2mO:1160574794) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/server_rendering_pool.rb#L74)):*
let's make an accessor method

js_context_pool_for_file

---

*[lib/react_on_rails/server_rendering_pool.rb, line 95 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVWGEkioDb8HWG5BCt:-KDVWGEkioDb8HWG5BCu:-597659646) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d50159371b69042a430f75899802df601929c32/lib/react_on_rails/server_rendering_pool.rb#L95)):*
this is a valid case that need to be checked. I.e., the code supports evaluating a piece of arbitrary JS on the server side.

Normally, you'd have your default server_js_file used, but there is a case where you don't have any server js file.

---

*[lib/react_on_rails/utils.rb, line 22 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVW_frBb5VZwRyHz-l:-KDVW_frBb5VZwRyHz-m:-1907423804) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/utils.rb#L22)):*
This makes sense. We need to be sure we document this well.

---

*[spec/dummy/client/webpack.server.js, line 7 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVXD3H0rNm9XfNQFnD:-KDVXD3H0rNm9XfNQFnE:1115614051) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/spec/dummy/client/webpack.server.js#L7)):*
See this for setting up the entry points:

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/webpack.client.base.config.js#L17

```js
  entry: {

    // See use of 'vendor' in the CommonsChunkPlugin inclusion below.
    vendor: [
      'babel-polyfill',
      'es5-shim/es5-shim',
      'es5-shim/es5-sham',
      'jquery',
    ],

    // This will contain the app entry points defined by webpack.hot.config and webpack.rails.config
    app: [
      './app/bundles/comments/startup/clientRegistration',
    ],
  },
```

---

*[spec/dummy/client/webpack.server.js, line 11 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVXL99XEAbdmJfMaP5:-KDVXL99XEAbdmJfMaP6:636418028) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/spec/dummy/client/webpack.server.js#L11)):*
I bet this is why you renamed serverRegistration. Please change it back.

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 15 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVXnqcy3cOJCrucqP3:-KDVXnqdL9W_vVXdkRfE:-2021650048) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/spec/dummy/client/webpack.server.rails.build.config.js#L15)):*
This looks correct. Maybe webpack.server was a file that should have been removed. It seems replaced by this file.

---

*[spec/dummy/client/app/startup/alternative-server.jsx, line 1 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVY2i_fG2uf-UC5G2F:-KDVY2i_fG2uf-UC5G2G:790824305) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/spec/dummy/client/app/startup/alternative-server.jsx#L1)):*
For the example, why not just include only the  one component that has a different server bundle?

Or else we really need to refactor this.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-:-KDVYEAETosvd8xVXDLA:807493757)*

justin808:
@jopdeklein @lucke84 We need to make sure the docs provide a really solid use case for why we need different bundles. I would think we maybe able to accomplish the same end result by passing in a prop from rails and having a few conditionals in the code.

In the case of client side bundles, there are clear reasons on why you'd separate the bundles:

1. `vendor-bundle.js` can change far less than your app code, so client browsers will have this file cached.
2. `app-bundle.js` can change frequently.
3. Some mobile pages may only want to load a bit of the js.

I don't see any of these applying to the server side.

Also see #287. Could that be a simpler solution.

Basically, before we add this feature, we need to be 100% sure that the extra complexity of this is useful.

Furthermore, we'll eventually move to using #183 for server rendering.

CC: @alexfedoseev @jbhatab @robwise @thewoolleyman

lucke84:
#287 is indeed something useful but if you need separate bundles in a more consistent way, it'll end up having the developers using this gem to hack a lot in order to achieve something that can work out of the box. At the same time, I agree on the fact this feature would introduce some complexity which has to be justified by a real-life, common need. Later on today we'll add a more structured comment to explain a proper use case. For now, I'm just about to update the PR with a couple changes to address part of your feedback (see specific comments on each item).

---

---

*[lib/generators/react_on_rails/templates/base/base/config/initializers/react_on_rails.rb.tt, line 18 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVTBkIksv8pgZT3-UR:-KDXh1HzFHlJY7N-Rs9x:-543371409) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/generators/react_on_rails/templates/base/base/config/initializers/react_on_rails.rb.tt#L18)):*
We were actually wondering what's the best way to deal with this. Maybe it's a bit old-school, but I like the consistency of a field's type, that's why I changed this to be always an array. Plus, at the moment it's probably not common at all because the framework doesn't support the functionality yet but, if well documented, I think people will start using it more and more often. 

Anyhow, what would you recommend? 
- To have a multi-type field (Ruby-style), which can be either string or array
- always the array with the assumption that the first file is the default
- an extra field to specify when using only one server bundle file

---

*[lib/react_on_rails/configuration.rb, line 13 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVTeVX1XuFP5CFcBmK:-KDXkax19UGTak95SkMQ:-236341455) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/configuration.rb#L13)):*
`concat` is better memory-wise, but again in this context `+=` is not a problem.

---

*[lib/react_on_rails/configuration.rb, line 43 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVUBkE17Y04yWiFWpu:-KDXiLN7goiqYNGnwmBf:952267012) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/configuration.rb#L43)):*
Happy to extract this, as long as it is a private method :) Plus, I think it should be a class method, shouldn't it?

---

*[lib/react_on_rails/server_rendering_pool.rb, line 16 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVUk-WhAs-ub2FPUk2:-KDXio1FxtGbfAIA6ROn:-272156301) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/lib/react_on_rails/server_rendering_pool.rb#L16)):*
`each_with_object` is handy but turns out it is not very efficient memory-wise. Since it doesn't get triggered that often (especially in prod) I'll change that.

---

*[lib/react_on_rails/server_rendering_pool.rb, line 95 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVWGEkioDb8HWG5BCt:-KDXvr4mMUh-LrtXAnc7:1920557896) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d50159371b69042a430f75899802df601929c32/lib/react_on_rails/server_rendering_pool.rb#L95)):*
Can you give an example of a situation like that? I think that being in the SSR context there should always be a bundle js file.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-:-KDY-mU9czRHS64giTwT:1504023319)*

jopdeklein:
Review status: 20 of 24 files reviewed at latest revision, 14 unresolved discussions.

---

*[README.md, line 225 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVSGqQzrx0R1wSik-p:-KDYHWRf_IxY9mWwmL4I:226842925) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d50159371b69042a430f75899802df601929c32/README.md#L225)):*
Indeed as you mention later we did this to output `server-bundle.js` and  `server-alternative-bundle.js` to avoid too many changes throughout the code. I think we can fix it in the webpack.config without changing anything else, I'll give it a shot.

---

*[spec/dummy/client/webpack.server.js, line 7 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/343#-KDVXD3H0rNm9XfNQFnD:-KDYJWo5aPePhD196lui:1534215167) ([raw file](https://github.com/shakacode/react_on_rails/blob/e9dd34d9f17ecf4ffef82e576d3a595e0e6ee9cc/spec/dummy/client/webpack.server.js#L7)):*
OK I see your point, but I am unsure where the outcome of this is used - just for development / hot version of the server rendering? I can set it up similar like we've done for the build but we'll need to find a way to load the correct one: 
```  
entry: {
    server: ['babel-polyfill', './app/startup/server'],
    'alternative-server': ['babel-polyfill', './app/startup/alternative-server']
}
```

Either way we have to discuss the need for different bundles, we'll try to sketch our scenario and we can debate whether our approach is the simplest and most useful, perhaps there are simpler paths to what we want to accomplish.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/343)*

justin808:
@jopdeklein Think about one big bundle and injecting a little JS code either before or after we render the components, create the stores.

Per #287...

justin808:
Given our discussion for #287, should we keep this PR open now? We can close and always re-open later if needed. @jopdeklein

justin808:
@jopdeklein @lucke84 Please see #345. I plan to merge later today. I think this should meet your needs.

justin808:
@lucke84 @jopdeklein Can we close this one?

jopdeklein:
@justin808 we're currently using our branch in production in order to fix a critical issue we were facing. We are not looking to deviate too far from your track, so I'll first try whether we can make it work with your proposal of a separate server bundle with a top level switch. If this works it should indeed be simpler than our bundle switching approach.

justin808:
@jopdeklein you probably don't mean a separate bundle, I'm guessing. You should be able to require whichever file you need dynamically. Use `require` and not `import`.

jopdeklein:
@justin808 You're right, I meant we have a separate `server`, `client-responsive` and `client-mobile` bundles. The server bundle `requires` the correct components for rendering on the server, while the client bundles are specialized for the corresponding device (basically the RESS approach).

I implemented your suggestions yesterday and it seems to work fine, and @lucke84 and I agree this is a simpler approach - feel free to close this PR.

PR #345, opened 2016-03-23
Providing rails_context always to generator functions (store and component)

justin808:
## Background:
As of v4, we provide the location as a second param to component generator functions after props, and **ONLY** on the server rendering. This is not in the current API for store generator functions.

### Needing the current url path for server rendering
Suppose you want to display a nav bar with the current navigation link highlighted by the URL. When you server render the code, you will need to know the current URL/path if that is what you want your logic to be based on. This could be added to props, or ReactOnRails can add this automatically as another param to the generator function (or maybe an additional object, where we'll consider other additional bits of system info from Rails, like maybe the locale, later).

### Needing the I18n.locale
Suppose you want to server render your react components with a the current Rails locale. We need to pass the I18n.locale to the view rendering.

## Details
We'll do a breaking change of changing the 2nd parameter from a string to an object, but this should affect very few users, as we only passed this for server rendering.

The new api will be to pass 2 params when generating React components or hydrating redux stores:

So if you register `MyAppComponent`, it will get called like:

```js
reactComponent = MyAppComponent(props, railsContext);
```
and for a store:

```js
reduxStore = MyReduxStore(props, railsContext);
```

The `railsContext` may have:

```ruby
     {
        # URL settings
        href: request.original_url,
        location: "#{uri.path}#{uri.query.present? ? "?#{uri.query}": ""}",
        scheme: uri.scheme, # http
        host: uri.host, # foo.com
        pathname: uri.path, # /posts
        search: uri.query, # id=30&limit=5

        # Locale settings
        i18nLocale: I18n.locale,
        i18nDefaultLocale: I18n.default_locale,
        httpAcceptLanguage: request.env["HTTP_ACCEPT_LANGUAGE"]
      }
```

Additionally, you can customize the values.

Set the class for the `rendering_extension`
```ruby
  config.rendering_extension = RenderingExtension
```

Implement it like this:

```ruby
module RenderingExtension

  # Return a Hash that contains custom values from the view context that will get passed to
  # all calls to react_component and redux_store for rendering
  def self.custom_context(view_context)
     {
       somethingUseful: view_context.session[:something_useful]
     }
  end
end
```

This would put a prop of `somethingUseful` in the railsContext passed to all react_component and redux_store calls.

Since you can't access the rails session from JavaScript (or other values available in the view rendering context), this might useful.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/345)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Please see [Getting Started](#getting-started) for how to set up your Rails proj
 + [How it Works](#how-it-works)
     - [Client-Side Rendering vs. Server-Side Rendering](#client-side-rendering-vs-server-side-rendering)
     - [Building the Bundles](#building-the-bundles)
+    - [Rails Context)(#rails-context)
     - [Globally Exposing Your React Components](#globally-exposing-your-react-components)
     - [ReactOnRails View Helpers API](#reactonrails-view-helpers-api)
     - [ReactOnRails JavaScript API](#reactonrails-javascript-api)
@@ -207,6 +208,84 @@ In the following screenshot you can see the 3 parts of React on Rails rendering:
 Each time you change your client code, you will need to re-generate the bundles (the webpack-created JavaScript files included in application.js). The included Foreman `Procfile.dev` will take care of this for you by watching your JavaScript code files for changes. Simply run `foreman start -f Procfile.dev`.
 
 On Heroku deploys, the `lib/assets.rake` file takes care of running webpack during deployment. If you have used the provided generator, these bundles will automatically be added to your `.gitignore` in order to prevent extraneous noise from re-generated code in your pull requests. You will want to do this manually if you do not use the provided generator.
+
+### Rails Context
+When you use a "generator function" to create react components or you used shared redux stores, you get 2 params passed to your function:
+
+1. Props that you pass in the view helper of either `react_component` or `redux_store`
+2. Rails contextual information, such as the current pathname. You can customize this in your config file.
+
+This information should be the same regardless of either client or server side rendering.
+
+While you could manually pass this information in as "props", the rails_context is a convenience because it's pass consistently to all invocations of generator functions.
+
+So if you register your generator function `MyAppComponent`, it will get called like:
+
+```js
+reactComponent = MyAppComponent(props, railsContext);
+```
+and for a store:
+
+```js
+reduxStore = MyReduxStore(props, railsContext);
+```
+
+The `railsContext` has: (see implementation in file react_on_rails_helper.rb for method rails_context for the definitive list).
+
+```ruby
+  {
+    # URL settings
+    href: request.original_url,
+    location: "#{uri.path}#{uri.query.present? ? "?#{uri.query}": ""}",
+    scheme: uri.scheme, # http
+    host: uri.host, # foo.com
+    pathname: uri.path, # /posts
+    search: uri.query, # id=30&limit=5
+
+    # Locale settings
+    i18nLocale: I18n.locale,
+    i18nDefaultLocale: I18n.default_locale,
+    httpAcceptLanguage: request.env["HTTP_ACCEPT_LANGUAGE"]
+  }
+```
+
+#### Use Cases
+##### Needing the current url path for server rendering
+Suppose you want to display a nav bar with the current navigation link highlighted by the URL. When you server render the code, you will need to know the current URL/path if that is what you want your logic to be based on. This could be added to props, or ReactOnRails can add this automatically as another param to the generator function (or maybe an additional object, where we'll consider other additional bits of system info from Rails, like maybe the locale, later).
+
+##### Needing the I18n.locale
+Suppose you want to server render your react components with a the current Rails locale. We need to pass the I18n.locale to the view rendering.
+
+
+#### Customization of the rails_context
+You can customize the values passed in the rails_context in your `config/initializers/react_on_rails.rb`
+
+
+Set the class for the `rendering_extension`:
+
+```ruby
+  config.rendering_extension = RenderingExtension
+```
+
+Implement it like this above in the same file. Create a class method on the module called `custom_context` that takes the `view_context` for a param.
+
+```ruby
+module RenderingExtension
+
+  # Return a Hash that contains custom values from the view context that will get merged with
+  # the standard rails_context values and passed to all calls to generator functions used by the
+  # react_component and redux_store view helpers
+  def self.custom_context(view_context)
+    {
+     somethingUseful: view_context.session[:something_useful]
+    }
+  end
+end
+```
+
+In this case, a prop and value for `somethingUseful` will go into the railsContext passed to all react_component and redux_store calls.
+
+Since you can't access the rails session from JavaScript (or other values available in the view rendering context), this might useful.
 
 ### Globally Exposing Your React Components
 Place your JavaScript code inside of the provided `client/app` folder. Use modules just as you would when using webpack alone. The difference here is that instead of mounting React components directly to an element using `React.render`, you **expose your components globally and then mount them with helpers inside of your Rails views**.
@@ -479,9 +558,15 @@ You may wish to have 2 React components share the same the Redux store. For exam
 
 Suppose the Redux store is called `appStore`, and you have 3 React components that each need to connect to a store: `NavbarApp`, `CommentsApp`, and `BlogsApp`. I named them with `App` to indicate that they are the registered components.
 
-You will need to make a function that can create the store you will be using for all components and register it via the `registerStore` method. Note, this is a **storeCreator**, meaning that it is a function that takes props and returns a store:
+You will need to make a function that can create the store you will be using for all components and register it via the `registerStore` method. Note, this is a **storeCreator**, meaning that it is a function that takes (props, location) and returns a store:
 
 ```
+function appStore(props, railsContext) {
+  // Create a hydrated redux store, using props and the railsContext (object with 
+  // Rails contextual information).
+  return myAppStore;
+}
+
 ReactOnRails.registerStore({
   appStore
 });
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -94,6 +94,7 @@ module ReactOnRailsHelper
     # server has already rendered the HTML.
     # We use this react_component_index in case we have the same component multiple times on the page.
 
+    # TODO: remove this
     options, props = parse_options_props(component_name, options, other_options)
 
     react_component_index = next_react_component_index
@@ -139,11 +140,13 @@ module ReactOnRailsHelper
                                   content_tag_options)
 
     # IMPORTANT: Ensure that we mark string as html_safe to avoid escaping.
-    <<-HTML.html_safe
+    result = <<-HTML.html_safe
 #{component_specification_tag}
     #{rendered_output}
     #{replay_console(options) ? console_script : ''}
     HTML
+
+    prepend_render_rails_context(result)
   end
 
   # Separate initialization of store from react_component allows multiple react_component calls to
@@ -165,7 +168,8 @@ module ReactOnRailsHelper
     else
       @registered_stores ||= []
       @registered_stores << redux_store_data
-      render_redux_store_data(redux_store_data)
+      result = render_redux_store_data(redux_store_data)
+      prepend_render_rails_context(result)
     end
   end
 
@@ -233,12 +237,31 @@ module ReactOnRailsHelper
 
   private
 
+  # prepend the rails_context if not yet applied
+  def prepend_render_rails_context(render_value)
+    return render_value if @rendered_rails_context
+
+    data = {
+      rails_context: rails_context
+    }
+
+    @rendered_rails_context = true
+
+    rails_context_content = content_tag(:div,
+                                        "",
+                                        id: "js-react-on-rails-context",
+                                        style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
+                                        data: data)
+    "#{rails_context_content}\n#{render_value}".html_safe
+  end
+
   def render_redux_store_data(redux_store_data)
-    content_tag(:div,
-                "",
-                class: "js-react-on-rails-store",
-                style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
-                data: redux_store_data)
+    result = content_tag(:div,
+                         "",
+                         class: "js-react-on-rails-store",
+                         style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
+                         data: redux_store_data)
+    prepend_render_rails_context(result)
   end
 
   def next_react_component_index
@@ -265,6 +288,7 @@ module ReactOnRailsHelper
 
     wrapper_js = <<-JS
 (function() {
+  var railsContext = #{rails_context.to_json};
 #{initialize_redux_stores}
   var props = #{props_string(props)};
   return ReactOnRails.serverRenderReactComponent({
@@ -272,7 +296,7 @@ module ReactOnRailsHelper
     domNodeId: '#{dom_id}',
     props: props,
     trace: #{trace(options)},
-    location: '#{request.fullpath}'
+    railsContext: railsContext
   });
 })()
     JS
@@ -314,11 +338,41 @@ module ReactOnRailsHelper
       memo << <<-JS
 reduxProps = #{props};
 storeGenerator = ReactOnRails.getStoreGenerator('#{store_name}');
-store = storeGenerator(reduxProps);
+store = storeGenerator(reduxProps, railsContext);
 ReactOnRails.setStore('#{store_name}', store);
       JS
     end
     result
+  end
+
+  # This is the definitive list of the default values used for the rails_context, which is the
+  # second parameter passed to both component and store generator functions.
+  def rails_context
+    @rails_context ||= begin
+      uri = URI.parse(request.original_url)
+      # uri = URI("http://foo.com/posts?id=30&limit=5#time=1305298413")
+
+      result = {
+        # URL settings
+        href: request.original_url,
+        location: "#{uri.path}#{uri.query.present? ? "?#{uri.query}" : ''}",
+        scheme: uri.scheme, # http
+        host: uri.host, # foo.com
+        pathname: uri.path, # /posts
+        search: uri.query, # id=30&limit=5
+
+        # Locale settings
+        i18nLocale: I18n.locale,
+        i18nDefaultLocale: I18n.default_locale,
+        httpAcceptLanguage: request.env["HTTP_ACCEPT_LANGUAGE"]
+      }
+
+      if ReactOnRails.configuration.rendering_extension
+        custom_context = ReactOnRails.configuration.rendering_extension.custom_context(self)
+        result.merge!(custom_context) if custom_context
+      end
+      result
+    end
   end
 
   def raise_on_prerender_error(options)
--- a/docs/additional-reading/react-router.md
+++ b/docs/additional-reading/react-router.md
@@ -6,12 +6,14 @@ However, when attempting to use server-rendering, it is necessary to take steps 
 If you are working with the HelloWorldApp created by the react_on_rails generator, then the code below corresponds to the module in `client/app/bundles/HelloWorld/startup/HelloWorldAppServer.jsx`.
 
 ```js
-const RouterApp = (props, location) => {
-  const store = createStore(props);
-
+const RouterApp = (props, railsContext) => {
   let error;
   let redirectLocation;
   let routeProps;
+  const { location } = railsContext;
+  
+  // create your hydrated store
+  const store = createStore(props);
 
   // See https://github.com/reactjs/react-router/blob/master/docs/guides/advanced/ServerRendering.md
   match({ routes, location }, (_error, _redirectLocation, _routeProps) => {
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -55,7 +55,8 @@ module ReactOnRails
       server_renderer_pool_size: 1,
       server_renderer_timeout: 20,
       skip_display_none: false,
-      webpack_generated_files: []
+      webpack_generated_files: [],
+      rendering_extension: nil
     )
   end
 
@@ -65,14 +66,15 @@ module ReactOnRails
                   :logging_on_server, :server_renderer_pool_size,
                   :server_renderer_timeout, :raise_on_prerender_error,
                   :skip_display_none, :generated_assets_dirs, :generated_assets_dir,
-                  :webpack_generated_files
+                  :webpack_generated_files, :rendering_extension
 
     def initialize(server_bundle_js_file: nil, prerender: nil, replay_console: nil,
                    trace: nil, development_mode: nil,
                    logging_on_server: nil, server_renderer_pool_size: nil,
                    server_renderer_timeout: nil, raise_on_prerender_error: nil,
                    skip_display_none: nil, generated_assets_dirs: nil,
-                   generated_assets_dir: nil, webpack_generated_files: nil)
+                   generated_assets_dir: nil, webpack_generated_files: nil,
+                   rendering_extension: nil)
       self.server_bundle_js_file = server_bundle_js_file
       self.generated_assets_dirs = generated_assets_dirs
       self.generated_assets_dir = generated_assets_dir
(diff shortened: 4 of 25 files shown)


justin808:
My gut tells me to make extra params an object, and leave the first param as the props object, for the JS code. Otherwise, people would have to update their JS app components. They still might have to if they were using the location 2nd param, but that's unlikely to be many people.

alexfedoseev:
> Maybe 2nd param can be an object? so we have can have { fullUrl: <full URL>, path: <only path>, query: <all query and hash params> }

:+1: 

> Should ReactOnRails be providing this? Certainly this could be added to props. But since this is universal, might make sense to have ReactOnRails do this.

It's needed for react-router on server, so we shouldn't rely on props from user here.

> Is the way that I added a store/reducer etc. correct for adding the location info?

Seems fine.

> How does this fit with with using redux-react-router? And our sort of app, where we use redux-react-router after the top level URL.

If `redux-react-router` is used, I guess we should inject location info into its store. Looks like it might be done by just putting it in initial state ([link](https://github.com/reactjs/react-router-redux/blob/master/src/reducer.js#L8)), but I haven't looked much into it, so not sure is this correct way.

justin808:
@alexfedoseev Given that file with this, how do we inject the current location?

```javascript
/**
 * This action type will be dispatched when your history
 * receives a location change.
 */
export const LOCATION_CHANGE = '@@router/LOCATION_CHANGE'

const initialState = {
  locationBeforeTransitions: null
}

/**
 * This reducer will update the state with the most recent location history
 * has transitioned to. This may not be in sync with the router, particularly
 * if you have asynchronously-loaded routes, so reading from and relying on
 * this state it is discouraged.
 */
export function routerReducer(state = initialState, { type, payload }) {
  if (type === LOCATION_CHANGE) {
    return { ...state, locationBeforeTransitions: payload }
  }

  return state
}
```

justin808:
maybe I return for the extra values:
```js
{ location: { href, pathname, hash, search } }
```

jbhatab:
@justin808 I like it as an object as the second variable with all of the location data. Not sure on a string of the full url vs the object. I feel like people are used to parsing urls but generally they get it from window that parses it kinda for them, so I could see it being nice broken up like that.

justin808:
Debating on whether or not to pass the location parts, as this library is very popular: https://www.npmjs.com/package/url-parse (550k downloads the past month).

```ruby
  def rails_context
    uri = URI.parse(request.original_url)
    # uri = URI("http://foo.com/posts?id=30&limit=5#time=1305298413")

    result = {
      # URL settings
      location: request.original_url,
      scheme: uri.scheme, # http
      host: url.host, # foo.com
      pathname: uri.path, # /posts
      search: uri.query, # id=30&limit=5
      hash: uri.fragment, # "time=1305298413"

      # Locale settings
      i18nLocale: I18n.locale,
      i18nDefaultLocale: I18n.default_locale,
      httpAcceptLanguage: request.env['HTTP_ACCEPT_LANGUAGE']
    }

    if ReactOnRails.configuration.rendering_extension
      custom_context = ReactOnRails.configuration.rendering_extension.custom_context(self)
      if custom_context
        result.merge!(custom_context)
      end
    end
    result
  end
```

This can set the locale quite nicely: https://github.com/iain/http_accept_language.

I'm considering allowing a RenderingExtension like this:

Set the class for the `rendering_extension`
```ruby
  config.rendering_extension = RenderingExtension
```

Implement it like this:

```ruby
module RenderingExtension

  # Return a Hash that contains custom values from the view context that will get passed to
  # all calls to react_component and redux_store for rendering
  def self.custom_context(view_context)
     {
       somethingUseful: view_context.session[:something_useful]
     }
  end
end
```

This would put a prop of `somethingUseful` in the railsContext passed to all react_component and redux_store calls.

Since you can't access the rails session from JavaScript (or other values available in the view rendering context), this might useful.

So why not put this in props?

The advantage to the `railsContext` is that this is consistent for all calls.

Feedback?

justin808:
@alexfedoseev @martyphee @jbhatab This is ready to merge once I add tests for the new functionality. The work on the view helper and the JS libraries should be complete, as well as to the /spec/dummy sample app.

Any comments on the API?

I hope to merge this to master on Saturday!

jbhatab:
@justin808 I think the api for the RenderingExtension thing is a bit odd. The concept is fine for flexibility. Can we just have extra params that are passed in on the helper calls get passed in automatically? Like a wildcard args situation.

justin808:
Review status: 0 of 21 files reviewed at latest revision, 25 unresolved discussions.

---

*[app/helpers/react_on_rails_helper.rb, line 149 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoUyQ6VyVuH_MGxZIm:-KDoUyQ74_X60xW9tph1:-1026345503) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L149)):*
this will make sure that we've rendered a hidden div  with the contextual information exactly ONCE.

---

*[app/helpers/react_on_rails_helper.rb, line 172 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoV4rrYQULoS7mP-2U:-KDoV4rrYQULoS7mP-2V:-1026345503) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L172)):*
this will make sure that we've rendered a hidden div  with the contextual information exactly ONCE.

---

*[app/helpers/react_on_rails_helper.rb, line 242 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoV8LqgHjBpKIfmkzB:-KDoV8LqgHjBpKIfmkzC:-1897301607) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L242)):*
Only once! Else just send back the dom string passed.

---

*[app/helpers/react_on_rails_helper.rb, line 255 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoVGKgGy1DL-l6SiF1:-KDoVGKgGy1DL-l6SiF2:1404107467) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L255)):*
We put the context div first.

---

*[app/helpers/react_on_rails_helper.rb, line 291 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoVLQcw9L7eAp1EuAL:-KDoVLQcw9L7eAp1EuAM:-2105916355) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L291)):*
put the railsContext in a variable, used in the JS call to server render

---

*[app/helpers/react_on_rails_helper.rb, line 348 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoYWymmtTqdOm9K5m1:-KDoYWymmtTqdOm9K5m2:-455473404) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L348)):*
The definitive list of what goes in the rails context

---

*[node_package/src/clientStartup.js, line 35 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoYmWzXymKiyHhd23J:-KDoYmWzXymKiyHhd23K:935270624) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/clientStartup.js#L35)):*
We consistently call generator functions with the railsContext parameter

---

*[node_package/src/clientStartup.js, line 51 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoYxy8m93LsSW8-4Cp:-KDoYxy8m93LsSW8-4Cq:935270624) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/clientStartup.js#L51)):*
We consistently call generator functions with the railsContext parameter

---

*[node_package/src/clientStartup.js, line 73 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZ-E-Ce9nBL6RmB_o:-KDoZ-E-Ce9nBL6RmB_p:935270624) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/clientStartup.js#L73)):*
We consistently call generator functions with the railsContext parameter

---

*[node_package/src/clientStartup.js, line 106 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZ4CFosi3MNs9gKQ0:-KDoZ4CFosi3MNs9gKQ1:-1101274440) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/clientStartup.js#L106)):*
only had one initializeStore function, so no need to make it a parameter

---

*[node_package/src/createReactElement.js, line 25 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZKTi0g-GS0DxeDsa:-KDoZKTi0g-GS0DxeDsb:-1064009412) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/createReactElement.js#L25)):*
This change makes tracing now log the props and context. 

I'll have to experiment to see if this is overkill.

---

*[node_package/src/createReactElement.js, line 33 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZSb-3aqJscwHFdL8:-KDoZSb-3aqJscwHFdL9:701493849) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/node_package/src/createReactElement.js#L33)):*
yep, call with the railsContext, and not location.

# BREAKING CHANGE RIGHT HERE

Yes, going from a string to an object.

CC: @jbhatab, @thewoolleyman, @martyphee, @alexfedoseev    

---

*[spec/dummy/app/controllers/pages_controller.rb, line 5 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZdouxyyiIOHTqpVs:-KDoZdouxyyiIOHTqpVt:-1733808323) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/app/controllers/pages_controller.rb#L5)):*
example of having value in session to put into rails context

---

*[spec/dummy/client/app/components/HelloWorld.jsx, line 14 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZqc_9LhSAVPDkxQt:-KDoZqc_9LhSAVPDkxQu:40347432) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorld.jsx#L14)):*
optional, bc we do not get this for standard invocation, only generator functions

---

*[spec/dummy/client/app/components/HelloWorld.jsx, line 57 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoZkVNawVjF6zNtRkf:-KDoZkVNawVjF6zNtRkg:1011979049) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorld.jsx#L57)):*
We only get the RailsContext for generator functions.

---

*[spec/dummy/client/app/components/HelloWorldContainer.jsx, line 9 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_0PeWF2t7X4YIXEz:-KDo_0PeWF2t7X4YIXF-:-428916364) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorldContainer.jsx#L9)):*
@alexfedoseev @robwise Is this a good pattern to pass in the railsContext as a 3rd param to the container?

---

*[spec/dummy/client/app/components/HelloWorldContainer.jsx, line 23 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_9Yp10chPv-9r9mY:-KDo_9Yp10chPv-9r9mZ:-117711042) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorldContainer.jsx#L23)):*
Here's where we map the state to props for railsContext

---

*[spec/dummy/client/app/components/HelloWorldRedux.jsx, line 10 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_HIsIoEGWR5kF3Ys:-KDo_HIsIoEGWR5kF3Yt:385232533) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorldRedux.jsx#L10)):*
In the redux case, I'm always using railsContext

---

*[spec/dummy/client/app/components/RailsContext.jsx, line 1 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_KtBGtzS4R30e_Xc:-KDo_KtBGtzS4R30e_Xd:-1997336192) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/RailsContext.jsx#L1)):*
Make a simple table of the context values for test verification and demonstration



---

*[spec/dummy/client/app/reducers/RailsContextReducer.jsx, line 1 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo__5XtPtPH2wR7i5a:-KDo__5XtPtPH2wR7i5b:-1946309390) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/reducers/RailsContextReducer.jsx#L1)):*
@alexfedoseev does it make sense to have a reducer if there's no actions? I guess this is how the state gets put together.

---

*[spec/dummy/client/app/reducers/reducersIndex.jsx, line 8 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_fIfbLIaTcXnhBcW:-KDo_fIfbLIaTcXnhBcX:1136172024) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/reducers/reducersIndex.jsx#L8)):*
This is how we get the railsContext into the redux store.

---

*[spec/dummy/client/app/startup/ClientReduxApp.jsx, line 20 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_mdYtrFrD1E8bKB-:-KDo_mdYtrFrD1E8bKB0:793089826) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ClientReduxApp.jsx#L20)):*
@alexfedoseev Please confirm this is the best way to get the railsContext put into the props for redux.

Sort of interesting that we slam the param in, and then make it its own "reducer domain".

Maybe there's a better pattern since the the rails_context really is immutable?

---

*[spec/dummy/client/app/startup/ServerReduxApp.jsx, line 21 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoa58UArngbCyNhixS:-KDoa58UArngbCyNhixT:-2144374342) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ServerReduxApp.jsx#L21)):*
The server side example code is duplicated. Wondering if there's a better way.

---

*[spec/dummy/client/app/startup/ServerRouterApp.jsx, line 11 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoa8iMlqTGd9mNByAi:-KDoa8iMlqTGd9mNByAj:-67534141) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ServerRouterApp.jsx#L11)):*
This is how we get the router location needed.

---

*[spec/dummy/config/initializers/react_on_rails.rb, line 9 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoaHbYVUNqDr4aHqt2:-KDoaHbYVUNqDr4aHqt3:1754606036) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/config/initializers/react_on_rails.rb#L9)):*
@martyphee This is a custom extension to the rails_context.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345)*

justin808:
@jbhatab You already have props for that. So I don't understand your use case.

---

Review status: 0 of 21 files reviewed at latest revision, 25 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-:-KDoaVpht8ztXcLT2eTz:-875512025)*

justin808:
AFAIK, my only remaining task on this PR is to update the /spec/dummy tests to confirm the values in the displayed table.



---

Review status: 0 of 21 files reviewed at latest revision, 25 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-:-KDofJAzNlngwD7pnw2p:-691940616)*

alexfedoseev:
@justin808 @jbhatab :lgtm: 

---

Reviewed 13 of 17 files at r2, 4 of 5 files at r3, 4 of 5 files at r4.
Review status: all files reviewed at latest revision, 25 unresolved discussions.

---

*[app/helpers/react_on_rails_helper.rb, line 348 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoYWymmtTqdOm9K5m1:-KDomf0gUvVLXxiEZI-L:-913168351) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/app/helpers/react_on_rails_helper.rb#L348)):*
This looks good! :+1: 

---

*[spec/dummy/client/app/components/HelloWorldContainer.jsx, line 9 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_0PeWF2t7X4YIXEz:-KDon6f26tZlj5bkJrPF:-1025520117) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/components/HelloWorldContainer.jsx#L9)):*
I don't see any downsides here.

---

*[spec/dummy/client/app/reducers/RailsContextReducer.jsx, line 1 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo__5XtPtPH2wR7i5a:-KDonNdoG8Z8Jle3vH_o:-1396648732) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/reducers/RailsContextReducer.jsx#L1)):*
It makes sense, we must have reducer to put some initial data (even if there's no actions).

---

*[spec/dummy/client/app/startup/ClientReduxApp.jsx, line 20 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_mdYtrFrD1E8bKB-:-KDor5LLJMZBoPDOJkKM:2063277702) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ClientReduxApp.jsx#L20)):*
I'm not a big fan of mutating args (I usually use `Object.assign` to merge objects), but it seems fine.

---

*[spec/dummy/client/app/startup/ServerReduxApp.jsx, line 21 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoa58UArngbCyNhixS:-KDorsrphTyEFPbAwFfP:-1078557511) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ServerReduxApp.jsx#L21)):*
Abstract it into module `composeInitialState`, which will be used by client and server startup components.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-:-KDosGIsEqEfqJvqjDaM:1982356996)*

justin808:
Review status: all files reviewed at latest revision, 25 unresolved discussions.

---

*[spec/dummy/client/app/startup/ClientReduxApp.jsx, line 20 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDo_mdYtrFrD1E8bKB-:-KDovSnRNpk61HIPm7tg:575024669) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ClientReduxApp.jsx#L20)):*
Done.

```js
export default (props, railsContext) => {
  const combinedReducer = combineReducers(reducers);
  const combinedProps = Object.assign({}, props, railsContext);
  const store = applyMiddleware(middleware)(createStore)(combinedReducer, combinedProps);
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345)*

justin808:
Thanks Alex!

---

Review status: all files reviewed at latest revision, 22 unresolved discussions.

---

*[spec/dummy/client/app/startup/ServerReduxApp.jsx, line 21 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDoa58UArngbCyNhixS:-KDoxFfO68EBzoROPn4f:978538665) ([raw file](https://github.com/shakacode/react_on_rails/blob/3d1c4602ffe036f2de9d647661a26d0733f2c199/spec/dummy/client/app/startup/ServerReduxApp.jsx#L21)):*
Whole file or just the creation of the store, to abstract.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-:-KDoxRBA4cs8IOCyYGSd:426914609)*

justin808:
Review status: 19 of 24 files reviewed at latest revision, 24 unresolved discussions.

---

*[spec/dummy/spec/features/rails_context_spec.rb, line 1 \[r5\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDqoWL7KfXzKM15y6iz:-KDqoWL8SP_C-3O3UvnZ:1812278756) ([raw file](https://github.com/shakacode/react_on_rails/blob/b06819d070f2a21670f61ca3ce0172944a236e04/spec/dummy/spec/features/rails_context_spec.rb#L1)):*
This file checks client side values

---

*[spec/dummy/spec/requests/server_render_check_spec.rb, line 61 \[r5\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/345#-KDqoa_v7lNfvyKxYzUS:-KDqoa_v7lNfvyKxYzUT:-1620573205) ([raw file](https://github.com/shakacode/react_on_rails/blob/b06819d070f2a21670f61ca3ce0172944a236e04/spec/dummy/spec/requests/server_render_check_spec.rb#L61)):*
we're checking that we have the correct values on the server side

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/345)*

martyphee:
Sorry for the late response.  This won’t solve our current needs.  We still need a way load the proper i18n file for server side rendering.

jbhatab:
@martyphee you can't pass the locale file into with this?

martyphee:
Maybe.  Right now we're depending on a global `i18n` object which is a wrapper around Jed

Jed is initialized like so:

```

justin808:
@martyphee, @jbhatab You should be able to put these localization setups in separate files using the locale name in the file name (or as the whole file name). Then just do a require (not an import) based on the file appropriate for the locale.

Since we cache the "context" of the whole server bundle file, this might be slightly more performant.

Any reason that won't work?

martyphee:
The translations are needed for server rendering, so they'll be needed when the component is generated?

PR #350, opened 2016-03-25
Removed unnecessary options

jbhatab:
This PR built to remove a few options from the installer and the express server.

### Code removed from installer
Heroku
Hot Module Reloader
Express Server
Bootstrap
Ruby and JS linting

If anyone's curious on my logic, I can explain why I removed each in more detail but overall I felt they were outside of the scope of the gem and created maintenance issues. 

### Documentation and example app
The documentation in the readme hasn't been updated to reflect and that will have to be a dialog on how we want to port that over. Most likely we just need to remove all the relevant documentation and create small 'how tos' for each of these topics, but we can just remove things for now and add the how tos later. In the same vein as this, I also think we need to decide how we are going to move forward with example applications. we have the spec/dummy app and the webpack tutorial. I think we should stick with a single place for these things so we can make maintenance more stable. 

### How the app is run in develop
I kept the Procfile to run the app and the npm run command in the root package json but I think we should remove both the root package json and the Procfile in favor of manually running both commands. People can setup their own dev enviroments. I think this is outside the scope of this PR though. The relevant messages that pop up when running the generator were changed to reflect this though.

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/350)


justin808:
@jbhatab Definitely say a few sentences in the PR description and commit comments.

Please let me know when when this is ready for review. Thanks.

---

Review status: 0 of 40 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/base_generator.rb, line 120 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDkhtACZvAAtv2a4VSL:-KDkhtACZvAAtv2a4VSM:1789396393) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L120)):*
How are the readme's part of the options to remove? Yes, these small readme's will get some updates. At the min, they should point toward the main docs.

---

*[lib/generators/react_on_rails/base_generator.rb, line 137 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDki3UGJvLUKLrnCGMr:-KDki3UGJvLUKLrnCGMs:-2018510195) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L137)):*
this is needed for server rendering

did you test this?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-:-KDkhjh2XhSIy_o9MG9f:-1175392011)*

jbhatab:
*[lib/generators/react_on_rails/base_generator.rb, line 137 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDki3UGJvLUKLrnCGMr:-KDl6W727Y6fhY-J33AH:-1415912557) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L137)):*
You actually added it to the gem dependency of react_on_rails so it wasn't needed and we did test this with rake run_rspec:examples.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/350)*

jbhatab:
Yeah sorry about that, added a comment. Let me know if you need more details, this is a big one.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-:-KDl6biMaHy8a5m7nwgR:-1077653961)*

jbhatab:
*[lib/generators/react_on_rails/base_generator.rb, line 120 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDkhtACZvAAtv2a4VSL:-KDl6gL8g6KKf3r2Rsy6:183678132) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L120)):*
@justin808 Maybe this belongs in a separate PR but I don't think these belong in the generated code. I'd like all necessary documentation to be in the readme in a getting started section.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/350)*

justin808:
Reviewed 40 of 40 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/base_generator.rb, line 120 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDkhtACZvAAtv2a4VSL:-KDl7TiJ855f7-QD4gGt:-945047399) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L120)):*
Let's leave in the place holders at least for now. Thanks.  That way I don't have to poke around for adding it back later. 

---

*[lib/generators/react_on_rails/base_generator.rb, line 137 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDki3UGJvLUKLrnCGMr:-KDl7bwgzh8WLtRfemOA:275434275) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L137)):*
OK.  Probably OK for now. there's issue with having this there if one wants to use jRuby, FYI.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react_on_rails/350)*

jbhatab:
*[lib/generators/react_on_rails/base_generator.rb, line 137 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/350#-KDki3UGJvLUKLrnCGMr:-KETQtx4pvkZVF58l8P8:-142982717) ([raw file](https://github.com/shakacode/react_on_rails/blob/8482885f22f9f89eb8a8f3ee2b9eeec725bd7458/lib/generators/react_on_rails/base_generator.rb#L137)):*
we can make a comment on that in the readme.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/350)*

PR #366, opened 2016-04-04
Address failure to sanitize console messages

justin808:
* Only happened with server rendering.
* Fix is to escape for closing script tag

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/366)

PR #370, opened 2016-04-05
Completes prior fix 366 for script sanitization

justin808:
* Some cases missed, per: http://stackoverflow.com/a/23983448/1009332
* Added serverRendering and port to railsContext
* Only logging attributes client side if tracing to avoid overwhelming
  server logs. On the client, one can optionally toggle expand the
  props. On the server, we stringified, and with a big set of props,
  that could be huge.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/370)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -48,7 +48,7 @@ Lint/HandleExceptions:
     - 'spec/dummy/bin/rake'
 
 Metrics/AbcSize:
-  Max: 26
+  Max: 28
 
 Metrics/CyclomaticComplexity:
   Max: 7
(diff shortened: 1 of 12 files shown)


justin808:
@alexfedoseev Take a quick peek at this one as well as #366.

justin808:
For example, one's code could conditionally log values if on the server, but not on the client. Basically, so long as the rendering comes out the same, React works. If you try to render the value of serverRendering, then React totally breaks.

alexfedoseev:
LGTM :+1:

PR #379, opened 2016-04-08
Updating to react 15

### LINKED ISSUE #378: Update to React 15

brucek:
Fixes #378

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/379)


lib/generators/react_on_rails/templates/base/base/client/package.json.tt
```diff
@@ -71,11 +71,11 @@
     <%- if options.redux? -%>
     "mirror-creator": "1.1.0",
     <%- end -%>
-    "react": "^0.14.8",
+    "react": "^15.0.0",
```
justin808: @brucek We need to change the peer dependency of the npm module to support either 14 or 15.

brucek:
I didn't look at what it would take to test against both versions, if you just meant the `package.json.tt` deps needed to support both I think this should do it.

PR #386, opened 2016-04-14
Fix file exists check for server bundle, which returns true for a dir.

timscott:
NOTE: Similar bugs may exist elsewhere in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/386)

### DIFF
--- a/lib/react_on_rails/server_rendering_pool.rb
+++ b/lib/react_on_rails/server_rendering_pool.rb
@@ -79,7 +79,7 @@ module ReactOnRails
 
       def create_js_context
         server_js_file = ReactOnRails::Utils.default_server_bundle_js_file_path
-        if server_js_file.present? && File.exist?(server_js_file)
+        if server_js_file.present? && File.file?(server_js_file)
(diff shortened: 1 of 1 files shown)


justin808:
http://www.gethourglass.com/blog/ruby-check-if-file-exists.html

Interesting one.

@samnang @mapreal19 Do you two like this change?

timscott:
@justin808  Ha, that's the same blog post I found. I did not know about this behavior of ```File.exist?``` I would call surprising.

mapreal19:
Looks good. Didn't know about this either. Is it possible to add a spec that confirm this bug?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/386#-:-KFNucqCwbf9ZlNoDnqT:-536050799)*

samnang:
Good catch. On Unix, everything is a file https://en.wikipedia.org/wiki/Everything_is_a_file

justin808:
Thanks @timscott!

Issue #381, opened 2016-04-11
Disable js-react-on-rails-component


danielhusar:
Hey, is it possible to disable outputting this element when rendering server side ?

justin808:
I don't understand the question.

justin808:
@danielhusar Please open another issue if you're still having problems.

danielhusar:
Hey, sorry I was meaning that hidden div with serialised props when doing server side rendering.
![screen shot 2016-05-06 at 09 44 00](https://cloud.githubusercontent.com/assets/1857751/15068423/1b88a2d6-136f-11e6-8338-1add26f9d175.png)

I use my own custom logic to retrieve state when doing server side rendering, and I would probably prefer something like `window.__INITIAL_STATE__` to store initial state.

justin808:
@danielhusar this is the code that bootstraps ReactOnRails: https://github.com/shakacode/react_on_rails/blob/master/node_package%2Fsrc%2FclientStartup.js

Feel free to submit a PR.

danielhusar:
Yep I know :) But Im not using that module, mainly because I migrated app from express, and I will need possibly to migrate it back so don't want to tie it too much to the rails.

Our mainly concerns are just seo for that hidden div, I can live with it for now :) If it came as issue I will fire the PR, cheers !

PR #387, opened 2016-04-14
Customizable DOM prefix for react components, stores, and contexts

anveo:
Requested in https://github.com/shakacode/react_on_rails/issues/381.

If there was a vulnerability, I suppose this could help stop attackers from fingerprinting what libraries are being used on a site. Some people also like to have control over their markup. If you think this is useful I could improve the documentation and add some tests.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/387)

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -107,7 +107,7 @@ module ReactOnRailsHelper
     component_specification_tag =
       content_tag(:div,
                   "",
-                  class: "js-react-on-rails-component",
+                  class: "#{ReactOnRails.configuration.dom_prefix}-component",
                   style: options.style,
                   data: options.data)
 
@@ -237,7 +237,7 @@ module ReactOnRailsHelper
 
     rails_context_content = content_tag(:div,
                                         "",
-                                        id: "js-react-on-rails-context",
+                                        id: "#{ReactOnRails.configuration.dom_prefix}-context",
                                         style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
                                         data: data)
     "#{rails_context_content}\n#{render_value}".html_safe
@@ -246,7 +246,7 @@ module ReactOnRailsHelper
   def render_redux_store_data(redux_store_data)
     result = content_tag(:div,
                          "",
-                         class: "js-react-on-rails-store",
+                         class: "#{ReactOnRails.configuration.dom_prefix}-store",
                          style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
                          data: redux_store_data)
     prepend_render_rails_context(result)
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -56,7 +56,8 @@ module ReactOnRails
       server_renderer_timeout: 20,
       skip_display_none: false,
       webpack_generated_files: [],
-      rendering_extension: nil
+      rendering_extension: nil,
+      dom_prefix: "js-react-on-rails"
     )
   end
 
@@ -66,7 +67,7 @@ module ReactOnRails
                   :logging_on_server, :server_renderer_pool_size,
                   :server_renderer_timeout, :raise_on_prerender_error,
                   :skip_display_none, :generated_assets_dirs, :generated_assets_dir,
-                  :webpack_generated_files, :rendering_extension
+                  :webpack_generated_files, :rendering_extension, :dom_prefix
 
     def initialize(server_bundle_js_file: nil, prerender: nil, replay_console: nil,
                    trace: nil, development_mode: nil,
@@ -74,7 +75,7 @@ module ReactOnRails
                    server_renderer_timeout: nil, raise_on_prerender_error: nil,
                    skip_display_none: nil, generated_assets_dirs: nil,
                    generated_assets_dir: nil, webpack_generated_files: nil,
-                   rendering_extension: nil)
+                   rendering_extension: nil, dom_prefix: nil)
       self.server_bundle_js_file = server_bundle_js_file
       self.generated_assets_dirs = generated_assets_dirs
       self.generated_assets_dir = generated_assets_dir
@@ -97,6 +98,7 @@ module ReactOnRails
 
       self.webpack_generated_files = webpack_generated_files
       self.rendering_extension = rendering_extension
+      self.dom_prefix = dom_prefix
     end
   end
 end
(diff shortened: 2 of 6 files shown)


justin808:
@anveo Can you please provide specifics on "a vulnerability"?

anveo:
I just meant that the default `js-react-on-rails-` prefix exposes that a site is using react_on_rails. Some people might not like that for aesthetic reasons.

The security angle is kinda contrived, but if there was a vulnerability in react_on_rails, then scanners might search for that tag. Just trying to think of reasons #381 might have been requested for.

justin808:
Let's see what other community members think of this one.

justin808:
I'm leaning toward accepting this one. @thewoolleyman, @samnang, @robwise, @alexfedoseev, thoughts?

robwise:
I don't see the relation to #381, is there someone who needs this? I guess I sort of can see the security through obscurity thing, but you can't change the `data-reactid` attributes that come from vanilla React, right? So you're going to be tipping your hand in that sense anyway?

alexfedoseev:
It makes sense for the case, when somebody found security hole in this gem and scans web for `js-react-on-rails-` to make something bad. I'd call it `wordpress case` lol. With that said I 👍 here.

thewoolleyman:
Sounds reasonable, +1

samnang:
👍

justin808:
@anveo VERY NICE work! I had zero code comments. Exactly as I would have done this.

---

Reviewed 2 of 4 files at r1, 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/387#-:-KGKSeIy0d7cxKVqoJSc:1277618924)*

justin808:
@anveo Can you please rebase this fix on top of master and I'll merge.

anveo:
@justin808 I rebased against master, but now 4 additional commits are showing up in this PR. Any ideas on how I can fix that?

justin808:
Your rebase did not go correctly.

We need to squash your commits into one commit.

You can do this a couple ways. It might be easiest to:

0. pull master (be updated)
1. create a new branch "my-branch"
2. git merge --squash "old-branch" into this branch
3. Once you verify, then you can rename the old branch to something else (or delete) and rename this new branch to match the one on this PR. Then do a "push -f".

anveo:
@justin808 Thanks for the advice. I ended up rebasing from upstream/master and force pushing. Seems to have corrected itself.

justin808:
:lgtm_strong: 

Merging if this passes CI.

---

Reviewed 6 of 6 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/387#-:-KGUaqWT3w-DSL_1HGSs:-20569596)*

justin808:
Reviewed 1 of 1 files at r8.

---

*[node_package/src/ReactOnRails.js, line 55 \[r8\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/387#-KGVByAOWRJVdIGMQc5r:-KGVByAOWRJVdIGMQc5s:918453978) ([raw file](https://github.com/shakacode/react_on_rails/blob/504a002ce27a656ec94c07069ce1314dd5fa1d9a/node_package/src/ReactOnRails.js#L55)):*
Now that I look at this, I really wonder if there's a way that this can only be in one spot, like the rails config file.

```
domPrefix: string  If specified, it must match ReactOnRails.config.dom_prefix
```

I thought a bit about this and the only sensical alternative would be to:


1. Have a JSON file read by both react_on_rails Ruby and the webpack config.
2. The webpack config can set a value for the global value.
3. The JS code will use the global value rather than the options.

@jbhatab and I were discussing a couple places where we needed both the webpack and the rails to have the same value, such as the name of the server rendering file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/387)*

justin808:
One more comment. The key question is if we could DRY this up so there's no chance of a misconfiguration between the JS and the Rails code.

I think we could with a little bit of work.

---

---

*[node_package/src/ReactOnRails.js, line 55 \[r8\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/387#-KGVByAOWRJVdIGMQc5r:-KGVEqsM-UAaBouYPkY5:-1147821248) ([raw file](https://github.com/shakacode/react_on_rails/blob/504a002ce27a656ec94c07069ce1314dd5fa1d9a/node_package/src/ReactOnRails.js#L55)):*
CC: @alexfedoseev

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/387#-:-KGVEaLtge3JD0F-aO2E:327710951)*

justin808:
Please see https://github.com/shakacode/react_on_rails/blob/master/docs%2Fcontributor-info%2Fcontributing.md. Notably, we need a changelog entry.

I think this OK for now. We can DRY this up later.

justin808:
@anveo @alexfedoseev @jbhatab @thewoolleyman @robwise @aaronvb @martyphee I'm sort of not convinced on the need for this, as the JS on the pages will include the ReactOnRails symbol. Given how easy it is to see that ReactOnRails is defined for a page:

```js
if (ReactOnRails) console.log("ReactOnRails is defined!");
```

I'm all for putting in security improvements, but I'm not convinced on this one.

Another issue is that components have a particular pattern of data attributes, such as `data-component-name`, `data-dom-id`, `data-props`.

robwise:
@justin808 that's basically what I said 15 days ago: https://github.com/shakacode/react_on_rails/pull/387#issuecomment-212351642

I should point out though, that most crawlers don't evaluate JavaScript, so saying that they could just run that JavaScript on the page is not the same as being able to read a server-rendered HTML attribute, as many crawlers won't be able to do the former but will the latter.

I also pointed out the `data-` attribute pattern, but that's a tip-off of React, not necessarily ReactOnRails. 

All of that being said, I still think it's a bit obscure.

justin808:
@alexfedoseev you commented here: https://github.com/shakacode/react_on_rails/pull/387#issuecomment-212358679

I'm tempted to close this. I'm trying to keep the docs and code simpler unless really justified.

justin808:
@jbhatab @alexfedoseev last call if any dissenters on closing this one. Yes, somebody could scan for websites using RoR, but we intend to make our platform NOT have security issues.

justin808:
Closing this for now. We can revisit this later if we feel this is worth doing.

PR #392, opened 2016-04-19
Removed required keyword arguments for Ruby 2.0 compatibility

zimmru:
react_on_rails currently requires Ruby 2.1, but it's really only a handful of lines that use required keyword arguments. This commit removes those so that react_on_rails is compatible with Ruby 2.0.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/392)


justin808:
@zimmru Any reason you can't move to Ruby 2.1?

robwise:
Seems like an easy win for extra compatibility though

ricochet2200:
For those of us that can't use rvm (no internet access, rpm requirements, etc.) or already have other gems that force a ruby version, changing ruby versions is a really big deal.  In my case 2.1 is a deal breaker.  Ideally this gem should match the [ruby requirements of rails 4](http://guides.rubyonrails.org/4_0_release_notes.html), which is "Ruby 2.0 preferred; 1.9.3+ required".  

With such a small number of rails projects using react, it seems wise to me to try to be as inclusive as possible to make it more mainstream.  I agree with @robwise, this seems like an easy win for the extra compatibility.

thewoolleyman:
I agree, it's a good change to make easier adoption.

However, I would want to see some sort of linter check added as part of this PR which would fail if any incompatible syntax is added, if that's possible.  Otherwise other incompatible stuff will just get added in the future.

justin808:
Sounds good -- Once this passes CI, I'll review and merge.

justin808:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/392)*

justin808:
@zimmru I can't merge this until it passes CI.

zimmru:
There were a bunch of recent commits that use Ruby 2.1 features, so it's no longer the handful of lines that are different. Probably not worth trying to maintain at this point.

justin808:
@zimmru Are you able to update to Ruby 2.1?

Are there any strong reasons that people want to use an advanced tool like React on Rails, with an old version of Ruby?

zimmru:
@justin808 It's, of course, easy to update my dev machine to Ruby 2.1. The issue is at work, where our servers have Ruby 2.0 on them. Updating them is a bit of a pain.

thewoolleyman:
@zimmru  Yep.  I feel your pain, but that's why Containers (e.g. Docker) and PaaS are all the rage and the direction of the future - persistent servers are difficult to move forward without effort and risk - even if you use tools like Chef and Ansible to automatically and declaratively update them.

But, it's a good direction, for [good reasons](https://medium.com/built-to-adapt/the-three-r-s-of-enterprise-security-rotate-repave-and-repair-f64f6d6ba29d#.d9y5v2om1)...

ricochet2200:
@thewoolleyman It is important to note that PaaS is not always an option for those that must host their own stuff.  If you are required to deploy on other people's hardware this becomes even more problematic as many times you will not have access to the internet.  For complex installations, containers aren't always a feasible solution.

justin808:
If we get several people that want this, and the PR is good, I'll merge.

thewoolleyman:
@justin808 I still think we should require a linter to fail in CI if newer syntax is used to consider this mergeable.  Otherwise, we'll just have a continuous stream of issues like this as people inadvertently use newer syntax on master and in pull requests.

justin808:
I'm closing. If several people want Ruby 2.0, please comment.

austinfox:
I'm using JRuby 1.7 and am unable to upgrade to JRuby 9000 (company won't allow it currently). It would be great if this issue could be re-opened

justin808:
@austinfox Your use case sounds valid. If you can make a PR that fixes this, I'll merge it. Otherwise, we [ShakaCode](http://www.shakacode.com) can contract with your firm and get it done.

Issue #403, opened 2016-04-25
Replacing URI.parse with Addressable


lucke84:
It seems the new `rails_context` method uses `uri = URI.parse(request.original_url)`, which sometimes fails to parse URIs with non-ASCII characters (see **react_on_rails_helper.rb**, line 361)

Here's an example of a real-life(-ish) URI that should be supported but fails on our ReactOnRails powered website:
```
http://mywebsite.com/mypage?myparam=méridien
```
(note the `é` character in `myparam`'s value), returning an error like:
```
URI::InvalidURIError: URI must be ascii only "http://mywebsite.com/mypage?myparam=mm\u{e9}ridien"
```

The [addressable gem](https://rubygems.org/gems/addressable/versions/2.4.0) seems to be a valid alternative (which might become the official replacement for URI in the future). Unless I'm missing something here, is there a possibility to switch to using `addressable`? 

Feedback of what I might be doing wrong (e.g. not encoding properly?) are more than welcome as well.

jbhatab:
@lucke84 interesting. Would you mind making a PR for this?

lucke84:
See https://github.com/shakacode/react_on_rails/pull/405 :)

jbhatab:
@lucke84 looks good. just check out Justins comments, I agree with both of them. Besides that, awesome work.

lucke84:
Happy to help :) The comment have been fixed already, it'd be nice to :shipit: .

justin808:
Merged.

PR #405, opened 2016-04-26
Replace URI parse with Addressable gem

lucke84:
PR related to issue https://github.com/shakacode/react_on_rails/issues/403

Also, adds the fragment of the requested URI as a attribute of the Rails context.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/405)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file. Items under
 
 Contributors: please follow the recommendations outlined at [keepachangelog.com](http://keepachangelog.com/). Please use the existing headings and styling as a guide, and add a link for the version diff at the bottom of the file. Also, please update the `Unreleased` link to compare to the latest release version.
 ## [Unreleased]
+##### Changed
+- Replace URI with Addressable gem. See [#405](https://github.com/shakacode/react_on_rails/pull/405) by [lucke84]
 
 ## [5.2.0] - 2016-04-08
 ##### Added
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gemspec
 # They must be defined here because of the way Travis CI works, in that it will only
 # bundle install from a single Gemfile. Therefore, all gems that we will need for any dummy/example
 # app have to be manually added to this file.
-gem "addressable"
 gem "bootstrap-sass"
 gem "coffee-rails", "~> 4.1.0"
 gem "jbuilder", "~> 2.0"
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gemspec
 # They must be defined here because of the way Travis CI works, in that it will only
 # bundle install from a single Gemfile. Therefore, all gems that we will need for any dummy/example
 # app have to be manually added to this file.
+gem "addressable"
 gem "bootstrap-sass"
 gem "coffee-rails", "~> 4.1.0"
 gem "jbuilder", "~> 2.0"
--- a/README.md
+++ b/README.md
@@ -213,7 +213,6 @@ The `railsContext` has: (see implementation in file react_on_rails_helper.rb for
     port: uri.port, 
     pathname: uri.path, # /posts
     search: uri.query, # id=30&limit=5
-    fragment: uri.fragment, # time=1305298413
 
     # Locale settings
     i18nLocale: I18n.locale,
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ The `railsContext` has: (see implementation in file react_on_rails_helper.rb for
     port: uri.port, 
     pathname: uri.path, # /posts
     search: uri.query, # id=30&limit=5
+    fragment: uri.fragment, # time=1305298413
 
     # Locale settings
     i18nLocale: I18n.locale,
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -346,13 +346,12 @@ ReactOnRails.setStore('#{store_name}', store);
       result = {
         # URL settings
         href: request.original_url,
-        location: "#{uri.path}#{uri.query.present? ? "?#{uri.query}" : ''}#{uri.fragment.present? ? "##{uri.fragment}" : ''}",
+        location: "#{uri.path}#{uri.query.present? ? "?#{uri.query}" : ''}",
         scheme: uri.scheme, # http
         host: uri.host, # foo.com
         port: uri.port,
         pathname: uri.path, # /posts
         search: uri.query, # id=30&limit=5
-        fragment: uri.fragment, # time=1305298413
 
         # Locale settings
         i18nLocale: I18n.locale,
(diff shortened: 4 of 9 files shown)


justin808:
Couple comments.

---

Reviewed 2 of 2 files at r1.

---

*[Gemfile, line 10 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGNvFyR1JYDm7FsdvfW:-KGNvFySU0xN4WJs3P8y:-1051848398) ([raw file](https://github.com/shakacode/react_on_rails/blob/332081ec3c45c40cbd46f132172ed2965c962403/Gemfile#L10)):*
I believe we need to modify the gemspec for a peer dependency.

Did you try to make a new app and see if it works?

We certainly don't want apps that use ReactOnRails to have include this gem.

---

*[app/helpers/react_on_rails_helper.rb, line 342 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGNv437F80fi1Hvej4x:-KGNv437F80fi1Hvej4y:786823491) ([raw file](https://github.com/shakacode/react_on_rails/blob/332081ec3c45c40cbd46f132172ed2965c962403/app/helpers/react_on_rails_helper.rb#L342)):*
We need a clear comment in the code as to why we are not using the standard URI.

What problem does this solve. 

Otherwise, this will be confusing to somebody later on.

We can refer to the github issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-:-KGNvVoF_mGFDbCEVS1_:1120263039)*

justin808:
@lucke84 Have you looked into why CI is failing? Maybe rebase on top of master and push to start CI again.

lucke84:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/helpers/react_on_rails_helper.rb, line 342 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGNv437F80fi1Hvej4x:-KGc4-2ra-lWDBl9sNB4:-499146009) ([raw file](https://github.com/shakacode/react_on_rails/blob/332081ec3c45c40cbd46f132172ed2965c962403/app/helpers/react_on_rails_helper.rb#L342)):*
Done.

---

*[Gemfile, line 10 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGNvFyR1JYDm7FsdvfW:-KGc45KAk2adeGo4wAd0:-499146009) ([raw file](https://github.com/shakacode/react_on_rails/blob/332081ec3c45c40cbd46f132172ed2965c962403/Gemfile#L10)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405)*

jbhatab:
Reviewed 2 of 3 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405)*

jbhatab:
@justin808 looks good to me if you want to merge it in.

justin808:
Another comment on the addition of the "hash".

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[app/helpers/react_on_rails_helper.rb, line 355 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGcxBqNpNZgoC7w_nSe:-KGcxBqNpNZgoC7w_nSf:-275309772) ([raw file](https://github.com/shakacode/react_on_rails/blob/a6fdbccb6a42677e9d623a7b983eb6a10e313b35/app/helpers/react_on_rails_helper.rb#L355)):*
This is a nice addition, actually superb.

however, this needs to be:

1. Match Location api for browser: https://developer.mozilla.org/en-US/docs/Web/API/Location (use Location.hash)
2. Update docs.
3. Provide test of this addition (see related ones).
4. Update PR description
5. Update CHANGELOG with this one. I'll revise the release and date. I'm going to do this with all PR requests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-:-KGcxeWN_bp7XyOziTi6:-2041398981)*

lucke84:
Thanks for the comments, I'd applied some changes (see specific comments).

---

Review status: 1 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/helpers/react_on_rails_helper.rb, line 355 \[r3\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-KGcxBqNpNZgoC7w_nSe:-KGh7AVCJkPOvvdgw0Eg:-1803604324) ([raw file](https://github.com/shakacode/react_on_rails/blob/a6fdbccb6a42677e9d623a7b983eb6a10e313b35/app/helpers/react_on_rails_helper.rb#L355)):*
I've updated the docs (README and CHANGELOG) and added tests. The PR description already includes the fact that the fragment is included, we can maybe specify it better in the changelog (should I do it or will you?).

I'm not 100% sure what you meant with point 1, do you want the attribute to be called "hash" instead of "fragment"?

BTW, the way we populate the "href" attribute is misleading, as we use request.original_url which does not include the hash/fragment. I added it to the "location" attribute but I'm not sure it's good to leave "href" like that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-:-KGh7tv8Tr7pIwYXS7P6:-1016674818)*

lucke84:
I removed the fragment part at all: Addressable is able to recognize it but the fragment is never sent to the server so there's no point in try to read it.

justin808:
:lgtm: 

---

Reviewed 1 of 3 files at r2, 1 of 2 files at r4, 1 of 2 files at r5, 1 of 4 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/405#-:-KGk2rSpBoib3ocS8ocb:980077423)*

justin808:
Thanks @lucke84. Nice job!

lucke84:
Hey @justin808, I was wondering if there's a simple way for you guys to release a new version of the 5.x.y series to include this change. According to the changelog, this change has been added to the 6.x.y series but I don't think you should be forced to upgrade in order to get it (it doesn't seem to be related whatsoever to the breaking changes of version 6, looks to me more like a minor/patch release). What do you think?

justin808:
@lucke84 Most of the changes concern the generator and a tiny bit with the running of tests, as well as some additional bits like yours.

The v6 change is because we have a couple additional config params for the test running.

I'll send you a PR of me updating the https://github.com/shakacode/react-webpack-rails-tutorial/ and if you still feel that this is too major let me know. Give me a few hours.

lucke84:
Sure, take your time 👍  I was just pointing out that those bits and pieces which are not part of bumping a major should have probably been released as minor/patch, hopefully useful suggestion for the next time :)

justin808:
@lucke84 We're following full semantic versioning being pushed in the JS community. In that, **ANY** deprecation or addition of something that needs to change requires a major version bump.

PR #411, opened 2016-05-04
Change div tags to script tags for props

martyphee:
WIP - Changed div's to script tags.

---

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/411)


justin808:
@martyphee Any update on this one?

martyphee:
Sorry, no.  I'll be working on it again this week.  I've been in London
with my family for two weeks and haven't had time to work on it.

jbhatab:
@martyphee let me know if you wanna chat about this at all too.

martyphee:
Yes, do you have any time tomorrow?

justin808:
Please refer to:

* https://github.com/shakacode/react_on_rails/pull/366
* https://github.com/shakacode/react_on_rails/pull/113

To merge, we need:

1. Ensure this works just fine with both server and client rendering.
2. Ensure this works flawlessly with Rails 5 and Turbolinks 5.
3. Ensure that we've properly addressed all possible sources of injected scripts.
4. Ensure that the performance gain is worth any extra complexity.

Take a look at the home page for https://www.producthunt.com/

Seems that ProductHunt uses this strategy of a script.



They probably have done tons of optimization.

Interesting that the embedded store values seem to be a string, rather than an object.

```js
    window.initialState = "[\"^ \",\"annotations\",[\"~#iM\",[69195,[\"~#iR\",[\"^ \",\"n\",\"AnnotationRecord\",\"v\",[\"^ \",\"type\",\"top\",\"postId\",null,\"categoryId\",1,\"userCount\",0,\"userIds\",[]]]],69203,[\"^2\",[\"^ \",\"n\",\"AnnotationRecord\",\"v\",[\"^ \",\"^3\",\"top
```

So maybe the strategy is to convert to an HTML JSON string, and then JSON parse the string?

justin808:
@martyphee Any update?

justin808:
@martyphee How's this one going? Any preliminary performance numbers?

martyphee:
No, sorry.  I haven't had time to work on it.​  I'll try to spend time this
week before I go on vacation.  Just got back from UK.

justin808:
@martyphee Keep us posted. Hopefully you'll have some perf numbers to share!

justin808:
@martyphee How's this one going?

martyphee:
Sorry for the delay. I'm on vacation right now. No plans to head over seas
for a while so I hope to revisit not week.

justin808:
Hi @martyphee I'm about to push out 6.1 this weekend. Would you like to try to get this in?

martyphee:
I have two failing tests yet.

```

**Comment by justin808** [comment]
@martyphee Please address these in the PR description and the final commit. https://github.com/shakacode/react_on_rails/pull/411#issuecomment-231896070

**Comment by justin808** [comment]
@martyphee Here's a security reference on the XSS topic: https://rorsecurity.info/cross-site-scripting-xss-rails

**Comment by justin808** [comment]
@martyphee Any new info on this one?

**Comment by justin808** [comment]
@martyphee How's this one going?

**Comment by justin808** [comment]
@martyphee Any more details on this one?

**Comment by justin808** [comment]
@martyphee Are you still interested in this fix?

File: app/helpers/react_on_rails_helper.rb [hunk 1]
```diff
@@ -222,31 +226,31 @@ def server_render_js(js_expression, options = {})
...
                          "",
                          class: "js-react-on-rails-store",
-                         style: ReactOnRails.configuration.skip_display_none ? nil : "display:none",
-                         data: redux_store_data)
-    prepend_render_rails_context(result)
+                         style: nil,
+                         data: redux_store_data) do
+      # redux_store_data
+      "var #{redux_store_data[:store_name].tr('-', '_')} = #{Yajl.dump(redux_store_data[:props])};".html_safe
+    end
```
kapowaz, line 251: Generating a JavaScript string from Ruby is a pretty bad idea in my experience. For one thing the generated identifier is in the global namespace, but it’s also potentially very fragile.

My personal preference for getting values from Ruby into JS is to provide them as `data-` attributes on some element, then pull them out with JS… but that seems to be pretty much what you were doing already. Is there some reason why you wanted to use a `<script>` tag rather than a `<div>`?

lededje: Global namespace shouldn't be a problem. Attempt to make it as unique as possible though. Prepend it with a bunch of underscores or something similar. I'd advise against using a data attribute though: You might accidentally include an [ambiguous ampersand](https://html.spec.whatwg.org/multipage/syntax.html#syntax-ambiguous-ampersand) which is [not allowed.](https://html.spec.whatwg.org/multipage/syntax.html#syntax-attribute-value)

martyphee:
@justin808 Yes, we still want to do this to try speeding up the page rendering.  I'll work on it this week.

justin808:
@martyphee How's this one going?

martyphee:
I started to do some work on it last night.  I'll try to get thru it
tonight.  In Edinburgh right now.

app/helpers/react_on_rails_helper.rb
```diff
@@ -102,11 +104,13 @@ def react_component(component_name, raw_options = {})
...
-      content_tag(:div,
-                  "",
+      content_tag(:script,
                   class: "js-react-on-rails-component",
-                  style: options.style,
-                  data: options.data)
+                  style: nil,
+                  data: options.data) do
+        props = Yajl.dump(options.props.is_a?(String) ? JSON.parse(options.props) : options.props)
+        "var #{options.dom_id.tr('-', '_')} = #{props};".html_safe
```
squadette, line 112: I would suggest not playing games with global Javascript variables but instead use

```
<script type="application/json" id="...">{ "json": "encoded-data" }</script>
```

Also, it would really help development if you use human-readable JSON in non-development environment, a-la:

```
  def json_safe_and_pretty(something)
    if Rails.env.development?
      json_escape(JSON.pretty_generate(something.as_json))
    else
      json_escape(something.to_json)
    end
  end
```

app/helpers/react_on_rails_helper.rb
squadette: @kapowaz a possible reason behind this is that initial props can be quite big, e.g. 100Kb.  It is really awkward and unreadable to cram it to HTML attribute, given that every quote is replaced with `&quot;`.

app/helpers/react_on_rails_helper.rb
squadette: It also looks like JSON `<script>` tag also solves the Turbolinks issue, because there is no need to execute any `<script>` tag, you just have to replace `el.getAttribute()` with (IIRC) `.textContent`.  Also, no `eval()`!

martyphee:
I haven't had much time to look at this yet.

justin808:
Hi @martyphee Any new developments on this one?

justin808:
@martyphee Is this one worth doing? @squadette any interest in working on this one?

justin808:
See #660.

justin808:
Deferring to #660.

PR #425, opened 2016-05-18
fix too early unmount

szyablitsky:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/425)

### DIFF
--- a/node_package/src/clientStartup.js
+++ b/node_package/src/clientStartup.js
@@ -145,10 +145,10 @@ export function clientStartup(context) {
     } else {
       if (turbolinksVersion5()) {
         debugTurbolinks(
-          'USING TURBOLINKS 5: document added event listeners turbolinks:before-cache and ' +
+          'USING TURBOLINKS 5: document added event listeners turbolinks:before-render and ' +
           'turbolinks:load.'
         );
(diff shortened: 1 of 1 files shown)


justin808:
@szyablitsky This is super awesome! Would you like to create a failing without, passing with test in the `spec/dummy` app? If not, I can try. I will merge your commit into a branch on my machine and then I'll push up a new PR. If you can write a test, that would be AMAZING! Thanks so much!

szyablitsky:
@justin808 testing this can be very tricky because of Capybara's waiting behavior. https://github.com/jnicklas/capybara#asynchronous-javascript-ajax-and-friends
Also rerendering actually happens every time. But depending on cache state time between `turbolinks:before-cache` and `turbolinks:before-render` can be zero (in which case rerendering is not visible) or it can be big enough to see the empty space in place of unmounted component.

justin808:
@szyablitsky At the least, we can put in an example that confirms this manually, in `spec/dummy`

I'm also going to push an RC with this fix and test it in my branch for the https://github.com/shakacode/react-webpack-rails-tutorial/.

justin808:
I merged this change here: 

https://github.com/shakacode/react_on_rails/commit/40553fc58976370ea23d08ce69511b869c4268d9

We should still have an example of this in `spec/dummy`.

Issue #437, opened 2016-05-30
`data-dom-id` collision with (fragment) caching


michaelbaudino:
We encountered a problem where `data-dom-id` of components on the same page are not unique.

This is particularly obvious when components are stored in cache fragments and can be ordered differently depending on parameters. For exemple, think about search results where each result is a cache fragment containing a component rendered using `react_component` (the fragment will contain the `data-dom-id` auto-incremented based on the rank this result had in the page **when rendered the first time**).

For the record, if some people have the same problem, the workaround we found is to explicitely define `data-dom-id` using the `id` parameter of the `react_component` method [as documented here](https://github.com/shakacode/react_on_rails#react_component).

The point of this PR is that the default `data-dom-id` should actually not be auto-incremented based on the rank of the component in the page, but rather by an UUID or (even better, but not sure if possible) by a MD5/SHA hash of its content.

PS: I saw a few existing issues related to fragment caching, but it was more about cache invalidation, not that kind of collision.

justin808:
@michaelbaudino Thanks for reporting this. We should get a fix out ASAP. If you have any other feedback, please let me know. A PR would definitely be appreciated.

justin808:
@michaelbaudino We probably don't actually need the auto-generated DOM ID for anything other than integration tests (what do we need this for?):

Let's have a config setting:

`dom_id_generator`: Values:

* false: (default) which means none
* UUID (don't really know about MD5)
* Allow the value to be a proc, so you can have a custom function, and we provide an example of this that gives the old behavior, which could be used just for tests, so that there's a super simple upgrade path for those apps depending on this.

@jbhatab @thewoolleyman @alexfedoseev @robwise Any comments?

Feedback is welcome.

We'll need a major version bump for this.

In the very short term, we need a doc PR to address this. We should have a doc file for Fragment Caching tips. (any volunteers?: call it `/docs/additional-reading/fragment-caching.md`)

@michaelbaudino Please share your tips on how you set up your cache key with react_on_rails. I'd like to consider eventually making a helper for this.

michaelbaudino:
You mean, kinda like #438 ? :smile:

justin808:
Yes, for anybody reading this. See #438. Big thanks to @michaelbaudino for fixing this!

justin808:
I want to get this one merged. The only thing holding me back is getting consensus on doing a major version bump. I think it's prudent to do a major version bump so somebody doing updates will look at the CHANGELOG.md.

[...]

PR #438, opened 2016-05-31
Fix `dom_id` collisions by using a UUID rather than an autoincremented index

michaelbaudino:
This PR ensures that all DOM nodes ids are unique when multiple instances of a component co-exist on the same page.

It also allows fragment caching of generated HTML (since the unicity does not depend on other components on the page).
    
Solves #437

A few thoughts following discussion in #437 (numbered, to ease followups):

1. If I understand correctly, `dom_id` is not only used in tests, but also to attach React components to DOM nodes (you're probably using `ReactDOM.render` or equivalent)
2. I did not implement the proc version, sorry :disappointed: (but I'm not sure it would really be useful, tbh)
3. I'd consider my implementation backward-compatible (since it only changes internal behaviour: I consider auto-generated ids to be internal) so I don't think a major bump is needed (at least, if you're using semantic versioning)... OK, actually, I see 1 way of considering generated HTML to be broken: if some users are looking for ids using an hard-coded value (like all the integration tests :cry:) or a number-only regexp like `/\d+/`...
4. Regarding fragment caching, we simply use a unique identifier as `id` parameter for `react_component` (in our case an `ActiveRecord` id, but it could as well be a UUID) and then wraps the call to `react_component` in a `cache(["v1", I18n.locale, some_dependencies...])` block
5. I'm not sure this is how you expected me to make the integration tests pass, let me know if it needs to be changed (maybe we should not explicitly define `id` for `react_component` calls that have a display-only purpose?)

I hope this helps.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/438)


justin808:
@michaelbaudino This is **REALLY OUTSTANDING**. 

We need to update the `/CHANGELOG.md` per prior examples **AND** explain what you might need to do if you did something like wrote tests that depended on the auto-generated component id.

I'm leaning to agree with you that we'll do 6.0.x for this. However, I'm pretty sure that semantic versioning fanatics would 
suggest that we bump to 7.0.0 for this.

I'm hoping that this only affects tests, and that with a good description in the CHANGELOG.md and this PR as a reference, then the change is minor enough not to cause pain.

Also, please squash to one commit.

Thanks again!

<details><summary><i>Previously, coveralls wrote…</i></summary>
> 
> [Coverage Status](https://coveralls.io/builds/6395429)
> 
> Coverage decreased (-0.1%) to 82.305% when pulling **866cc440539b4a5dab4333ec24e66cca1715e90c on michaelbaudino:fix-dom_id-collisions** into **c7b9209ccdc7476c0e182977c84f1e78ebb4280b on shakacode:master**.
> 
</details>


---

Reviewed 21 of 21 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/438#-:-KJ5Q-x-VUUMLhAedtS2:-1400960779)*

michaelbaudino:
OK @justin808 I squashed those commits and added a CHANGELOG update (quite clear, hopefully) and the version bump.

Let me know if there's something else to do :+1:

justin808:
I'm going to merge this. I might change to v7 however because there is work to be done, and I do believe that a 6.0.X change would suggest no chance of breaking tests.

<details><summary><i>Previously, coveralls wrote…</i></summary>
> 
> [Coverage Status](https://coveralls.io/builds/6412255)
> 
> Coverage decreased (-0.1%) to 82.305% when pulling **ea3c7b345bc69312e5e23361f56645de3bc45102 on michaelbaudino:fix-dom_id-collisions** into **c7b9209ccdc7476c0e182977c84f1e78ebb4280b on shakacode:master**.
> 
</details>


---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/438#-:-KJIC0Xp-QF6VSI-6lGj:2115631544)*

PR #464, opened 2016-06-28
fixed ReferenceError: navigator is not defined during server_rendering

ramusus:
I've got this error using react-check package and it's components. It uses [navigator Object](http://www.w3schools.com/jsref/obj_navigator.asp) inside

Here is related issue https://github.com/luqin/react-icheck/issues/25

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/464)

### DIFF
--- a/lib/react_on_rails/server_rendering_pool/exec.rb
+++ b/lib/react_on_rails/server_rendering_pool/exec.rb
@@ -136,6 +136,9 @@ function setTimeout() {
 function clearTimeout() {
   #{undefined_for_exec_js_logging('clearTimeout')}
 }
+
+navigator = {};
+
           JS
         end
 


justin808:
@ramusus @thewoolleyman @jbhatab My question on this PR is whether or not we want to stub everything missing. Possibly some components should not work for server rendering.

My other concern is that once we stub this, we no longer know that components are depending on this.

Incidentally, you can know if you are server rendering from the railsContext that is passed into components when being rendered.

justin808:
See https://github.com/luqin/react-icheck/issues/25. That component doesn't support server rendering. So I'm not yet convinced we should just start stubbing things. For one, if we stub the object, and then the component fails because some property is not there, then what? @ramusus

ramusus:
ok, agree. I solved my issue in another way, so I'm closing this PR for now

PR #476, opened 2016-07-13
Replace symlinks with copy files

alleycat-at-git:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/476)


justin808:
See #479

Issue #495, opened 2016-07-29
Port on localhost is 5000 not 3000 for me


cubadomingo:
Hey I just finished installing the gem and I noticed in the installation guide it mentions to visit localhost:3000/hello_world
<img width="918" alt="screen shot 2016-07-29 at 9 56 19 am" src="https://cloud.githubusercontent.com/assets/15992549/17250939/8506458c-5574-11e6-8354-facef3ae4188.png">
It was actually localhost:5000/hello_world for me.

Everything is still working properly but I'm not too sure if port 3000 isn't used anymore or if it could be either 3000 or 5000 and just depends on the setup. Regardless I was wondering if this was worthy enough to add to the README?

HintikkaKimmo:
Same here there is nothing starting at port :3000

cubadomingo:
@HintikkaKimmo Wait so you're saying that localhost:3000 does work for you? If so do you mind telling me what version of Rails are you running?

HintikkaKimmo:
Just corrected my comment so rails starts at port 5000 not 3000 as mentioned on the tutorial and CLI input

cubadomingo:
Okay so I did some digging around and even though the config/puma.rb file has this line:

`port        ENV.fetch("PORT") { 3000 }`

[Here](https://devcenter.heroku.com/articles/heroku-local#run-your-app-locally-using-foreman) it states:

> Your web process loads on port 5000 because this is what Foreman provides as a default in the $PORT env var

Considering we both have Rails 5 which uses Puma instead of WEBrick as the web dev server, that might be why it ports to :5000 instead of :3000. Maybe others who are using different version could chime in. Anyways I think it might be nice to add this in the README.md.

cubadomingo:
Update: Make sure you're on the latest version of the gem, they have it configured in the procfile to run from port 3000.

profsmallpine:
@cubadomingo Thanks mate.

PR #496, opened 2016-07-30
Updates README.md

cubadomingo:
Added the option to visit localhost:5000 after starting foreman, considering port 3000 doesn't work for some. In reference to #495

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/496)

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ We're definitely not doing that. With react_on_rails, webpack is mainly generati
   foreman start -f Procfile.dev
   ```
 
-7. Visit [localhost:3000/hello_world](http://localhost:3000/hello_world)
+7. Visit [localhost:3000/hello_world](http://localhost:3000/hello_world) or [localhost:5000/hello_world](http://localhost:5000/hello_world) if the former doesn't work.
 
 ### Installation Summary
 


README.md
```diff
@@ -137,7 +137,7 @@ We're definitely not doing that. With react_on_rails, webpack is mainly generati
   foreman start -f Procfile.dev
   ```
 
-7. Visit [localhost:3000/hello_world](http://localhost:3000/hello_world)
+7. Visit [localhost:3000/hello_world](http://localhost:3000/hello_world) or [localhost:5000/hello_world](http://localhost:5000/hello_world) if the former doesn't work.
```
justin808, line 140: Not sure we want to put that in. The generator is supposed to set the port, maybe.

This is really a puma configuration issue, and the port is listed lin the output.

@jbhatab would know about this one.

jbhatab, line 140: @justin808 we moved this to 3000 only. https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/templates/base/base/Procfile.dev.tt

You're right.

cubadomingo, line 140: Just looke at my Procfile and it seems to be missing the '-p 3000'. I'll add that in. Any ideas why it might not have generated that properly for me? @jbhatab

jbhatab, line 140: @cubadomingo You probably ran the generator before we fixed this? Hmmm. not sure exactly. Are you on the latest version?

cubadomingo, line 140: Had a typo on my gemfile 😅. Ran Bundle update and it's updated now. Thanks for the help guys!

jbhatab, line 140: 👍

PR #497, opened 2016-07-30
Remove docker from CI tests .travis.yml

justin808:
Docker was previously added for linting with CI tests.

It has not been used for many months, as all linting dependencies are loaded via npm and rubygrems directly.

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/497)


justin808:
@dylangrafmyre Please review this PR. Should we remove the docker-compose.yml file? I removed docker from the `rake` ci task a long time ago because we need to lock linting versions in the package.json file.

dylangrafmyre:
@justin808 I would ask @robwise. If no one is using Docker for this project, you can remove the compose file and `Dockerfile_tests`. Also if you are not using Selenium but Poltergeist, you can remove the other dependencies like the `xvfb` server.

I really don't know where this project stands.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/497#-:-KNwpTiptjly10ASzYSE:398325002)*

robwise:
I think we were only using the container for the linting, so since we lock that down via gemfile/package.json we should be able to delete this without a problem

PR #515, opened 2016-08-10
A proposal to update the "rubocop"

AlexKVal:
```
lib/react_on_rails/server_rendering_pool.rb:19:7: C: Style/MethodMissing: When using method_missing, define respond_to_missing? and fall back on super.
  def method_missing(sym, *args, &block) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

lib/react_on_rails/utils.rb:14:7: C: Style/NumericPredicate: Use $CHILD_STATUS.exitstatus.zero? instead of $CHILD_STATUS.exitstatus == 0.
  $CHILD_STATUS.exitstatus == 0
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

--
`TravisCI` fails but it is not connected to this PR.
Maybe I can figure it out where is the offender.

Local `rake` command show all green now.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/515)

### DIFF
--- a/lib/react_on_rails/server_rendering_pool.rb
+++ b/lib/react_on_rails/server_rendering_pool.rb
@@ -16,6 +16,7 @@ module ReactOnRails
         end
       end
 
+      # rubocop:disable Style/MethodMissing
       def method_missing(sym, *args, &block)
         pool.send sym, *args, &block
       end
--- a/lib/react_on_rails/utils.rb
+++ b/lib/react_on_rails/utils.rb
@@ -11,6 +11,7 @@ module ReactOnRails
     end
 
     def self.last_process_completed_successfully?
+      # rubocop:disable Style/NumericPredicate
       $CHILD_STATUS.exitstatus == 0
     end
 
--- a/spec/dummy/Gemfile.lock
+++ b/spec/dummy/Gemfile.lock
@@ -209,7 +209,7 @@ GEM
       rspec-mocks (~> 3.4.0)
       rspec-support (~> 3.4.0)
     rspec-support (3.4.1)
-    rubocop (0.40.0)
+    rubocop (0.42.0)
       parser (>= 2.3.1.0, < 3.0)
       powerpack (~> 0.1)
       rainbow (>= 1.99.1, < 3.0)


justin808:
@AlexKVal This is odd. Maybe you're using a different rubocop version?

AlexKVal:
```
› rubocop -v
0.42.0
```
It is the latest one, released 17 days ago, [https://rubygems.org/gems/rubocop/versions/0.42.0](https://rubygems.org/gems/rubocop/versions/0.42.0) because `Gemfiles` in this project don't enforce any particular version.

And it has these new rules
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodMissing
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericPredicate

I've tried `$CHILD_STATUS.exitstatus.zero?` but it has no `.zero?` method at that point,
hence I'm disabling the rule.

--
There are only three places where `rubocop` is mentioned in Gemfiles and both are w/o a particular version number:
`gem "rubocop", require: false` [here at spec/dummy/Gemfile#L60](https://github.com/shakacode/react_on_rails/blob/07a6e481875f618057a2878e8546b097010b674d/spec/dummy/Gemfile#L60)

And `gem "rubocop", require: false` [here at Gemfile#L17](https://github.com/shakacode/react_on_rails/blob/6976a0b1a879455b2d8f5f257da115634654eaf1/Gemfile#L17)

The only version is `rubocop (0.40.0)` [here at spec/dummy/Gemfile.lock#L212](https://github.com/shakacode/react_on_rails/blob/07a6e481875f618057a2878e8546b097010b674d/spec/dummy/Gemfile.lock#L212)

--
I use `rvm`. I've created a standalone `gemset` especially for clean setup for the project.

--
Once we were fixing the `eslint`, `babel-eslint`, and `eslint-plugin-babel` versions in the `React-Bootstrap` project because of the similar case.
From time to time new PRs were failing with linting errors because of updates with adding new rules in those projects.
Then we decided to use the custom set of rules `eslint-config-airbnb`.

I am not (yet) familiar with `rubocop`'s custom sets of rules and probably the simplest
way to prevent future `TravisCI` failures is to fix the `rubocop`'s version in Gemfiles.

On the other hand looking at the frequency of `https://github.com/bbatsov/rubocop/releases`
it should not be an issue at all and as a side bonus `TravisCI` always will hint us about new `rubocop`'s rules 👍

AlexKVal:
Oh. I see now 😄 
This is the fixing of the version `rubocop (0.40.0)` [here at spec/dummy/Gemfile.lock#L212](https://github.com/shakacode/react_on_rails/blob/07a6e481875f618057a2878e8546b097010b674d/spec/dummy/Gemfile.lock#L212)
Probably I've run `bundle install` in a different order, because `TravisCI`
now fails with these new rules.
rubocop 0 40

--
Then I rename the PR to `a proposal to upgrade rubocop` 😄

justin808:
Good finds!  We should fix the version of rubocop globally to the current. Linter versions should always be fixed at the project level.

justin808:
@samnang What gems here should we lock? https://github.com/shakacode/react_on_rails/blob/master/Gemfile

samnang:
@justin808 I feel most of thems in that files can be upgraded because those are for development, but we don't want to upgrade Rails to 5.0 yet :)

justin808:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/515#-:-KPMZ_fbEzt4eefdgwXl:bnfp4nl)*

PR #485, opened 2016-07-21
Add hook to handleError.

squaresurf:
Why:

* It is hard to send exceptions to an exception tracker (e.g. Rollbar)
  when they're all caught and sent to console.error.

This change addresses the need by:

* Update handleError to check for the existence of a logException
  function on the window, then pass the exception to it if it exists.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/485)

PR #521, opened 2016-08-17
Pass-through JS error to client

dzirtusss:
Pass-through simple version of #485

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/521)


dzirtusss:
@justin808 @robwise Please look - this is the most minimalist intrusion into error I can imagine. We just update the message and rethrow same error with all same parameters, incl. stacktrace

justin808:
Looking good.

@robwise Take a look. 

@samnang @robwise I don't see any worthwhile way to test this. I think we can skip a new test for this change.

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2.

---

*[node_package/src/handleError.js, line 38 \[r1\]](https://reviewable.io:443/reviews/shakacode/react_on_rails/521#-KPRW-rh_gQ8QV90iLqj:-KPRW-rijqkbRYgyDLZ_:ba0qqz8) ([raw file](https://github.com/shakacode/react_on_rails/blob/048c8d8522fbc46930bc68d7cf508c1b7ed7db72/node_package/src/handleError.js#L38)):*
> ```JavaScript
> }
> 
> export default (options) => {
> ```

const handleError = (option) => {
// blah blah
}

export default handleError;

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/521#-:-KPRWL-rkkGV6g2KPhHr:b-t0jgta)*

robwise:
I think this is a good compromise. 

Also, if we are going to be modifying original error messages like this, we should tell the user who is doing it. So `message` should be like `ReactOnRails encountered an error while rendering component: ${name}. Original message: ${e.message}`

Or something

justin808:
@dzirtusss Let's apply Rob's suggestion and call it good!

dzirtusss:
Done. What about `handleGeneratorFunctionIssue`? Keep?

justin808:
:lgtm:

(Yes keep)

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/521#-:-KPXBYQxF4Wq3fsht_qm:b3zksus)*

justin808:
I'll merge once CI passes!

justin808:
@dzirtusss Can you please update the CHANGELOG.md.

squaresurf:
:clap:

PR #536, opened 2016-08-29
update gem version suggested in README

Aeon:
the README instructs the user to install v5 of the the gem, but v6 is already available. When I followed the readme, I was very surprised when I ran npm upgrade and immediately got a new version of the npm package, which made me wonder if I was missing updates in the ruby gem as well.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/536)


justin808:
Thanks @Aeon!

PR #573, opened 2016-10-22
Update turbolinks.md

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/573)

### DIFF
--- a/docs/additional-reading/turbolinks.md
+++ b/docs/additional-reading/turbolinks.md
@@ -2,15 +2,23 @@
 
 * See [Turbolinks on Github](https://github.com/rails/turbolinks)
 * Currently support 2.5.x of Turbolinks and 5.0.0 of Turbolinks 5.
-* Turbolinks is currently included only via the Rails gem and the Rails manifest file rather than NPM. [Turbolinks Issue #658 ](https://github.com/rails/turbolinks/issues/658) discusses this.
+* You may include Turbolinks either via npm (recommended) or via the gem.
 
 ## Why Turbolinks?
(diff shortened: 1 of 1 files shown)


justin808:
@Judahmeek?

justin808:
@robwise, @alexfedoseev, @dzirtusss Please comment.

alexfedoseev:
@justin808 LGTM

Judahmeek:
Nice! Reading https://github.com/turbolinks/turbolinks-classic/issues/658 was pretty awesome.

dzirtusss:
@justin808 @alexfedoseev @Judahmeek @robwise LGTM, and just for further info, some issues I had with Turbolinks5 worth mentioning:
- CFRS tokens need thorough check with Turbolinks5. Turbolinks5 do change `head` element by JavaScript (not only `body`) on page change with correct csrf meta tag, but it was something with the way it was taken from head when several windows were opened, that our specs were not all passing. I didn't look details however, may be it is app code related, not library code. Anyway it may need additional check because there is cfrs helper in ReactOnRails and it need to work with Turbolinks5.
- Turbolinks5 send requests without `Accept: */*` in header, only exact like `Accept: text/html` which makes Rails behave a bit specifically compared to normal and do mime-parsing (which is skipped by desing when Rails see `*/*`). In my case it was triggering :mjml parser not found error, as example. For some more details on Rails and `*/*` can read http://blog.bigbinary.com/2010/11/23/mime-type-resolution-in-rails.html
- As noted above there should be no names conflict and all components need to be loaded and registered with first page load

PR #581, opened 2016-10-27
Modify register API to accept renderer functions

jtibbertsma:
See [Issue #477](https://github.com/shakacode/react_on_rails/issues/477)

---

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/581)


justin808:
@jtibbertsma Very intriguing! 

1. Can you modify the example in spec/dummy to demonstrate this? Or make `spec/dummy2`?
2. Alternately a PR to https://github.com/shakacode/react-webpack-rails-tutorial/ would be great to demonstrate this. The PR can refer to your fork.

@robwise @alexfedoseev We'll have to give this a close look in the coming weeks. However, please stay focused on F&G v2! We can potentially leverage this when we make a big push for performance after v2 is completed!

jtibbertsma:
@justin808 Sure, I'll work on a pr for react-webpack-rails-tutorial this weekend.

justin808:
@jtibbertsma I'm looking forward to seeing the changes!

justin808:
@jtibbertsma Any update?

jtibbertsma:
@justin808 I've been a bit busy this week. I should have some time to work on it today.

jtibbertsma:
https://github.com/shakacode/react-webpack-rails-tutorial/pull/337

justin808:
@jtibbertsma Would it make sense to have a simple example in `/spec/dummy` along with some simple tests of this. It's critical to keep really good test coverage to avoid regressions.

jtibbertsma:
@justin808 Sure, I'll work on it this week

justin808:
@jtibbertsma Keep me posted!

jtibbertsma:
@justin808 I added a simple integration test that shows that the renderer function gets called correctly.

justin808:
@jtibbertsma **REALLY AWESOME**. Let's address the comments!

---

Reviewed 8 of 12 files at r1, 8 of 8 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[CHANGELOG.md, line 8 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWffJ6PDaYtLufCy0fu:-KWffJ6PDaYtLufCy0fv:b-ovfgi7) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/CHANGELOG.md#L8)):*
> ```Markdown
> ## [Unreleased]
> ##### Added
> - New API registerRenderer which allows a user to manually render their app to the DOM by [jtibbertsma](https://github.com/jtibbertsma)
> ```

reference the PR, as done below for other changes

you can use the release 6.2.0

and change the bottom part showing the diff

---

*[README.md, line 456 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfzOKhpXH4oRmwUBx3:-KWfzOKhpXH4oRmwUBx4:b-abmgdp) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/README.md#L456)):*
> ```Markdown
> 
> 1. [React on Rails docs for react-router](docs/additional-reading/react-router.md)
> 1. Examples in [spec/dummy/app/views/react_router](spec/dummy/app/views/react_router) and follow to the JavaScript code in the [spec/dummy/client/app/startup/ServerRouterApp.jsx](spec/dummy/client/app/startup/ServerRouterApp.jsx).
> ```

we should mention the `registerRenderer`  here, or maybe in the docs for react-router.

---

*[README.md, line 476 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfzX2hI1K7oKjxAbaC:-KWfzX2hI1K7oKjxAbaD:b-kl19oh) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/README.md#L476)):*
> ```Markdown
>   + [Turbolinks](docs/additional-reading/turbolinks.md)
> 
> + **Javascript**
> ```

we need to refer to the page on code splitting here

---

*[docs/additional-reading/code-splitting.md, line 15 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfxrQHY1g-hu81SNgc:-KWfxrQHY1g-hu81SNgd:b-8ipbnw) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L15)):*
> ```Markdown
>  (client) 
> 
> Different markup is generated on the client than on the server. Why does this happen? When you register a component with `ReactOnRails.register`, react on rails will render the component as soon as the page loads. However, react-router renders a comment while waiting for the code chunk to be fetched from the server. This means that react will tear all of the server rendered code out of the DOM, and then rerender it a moment later once the code chunk arrives from the server, defeating most of the purpose of server rendering.
> ```

I'm a bit confused about the interaction between deferred rendering and react-router. Is the idea that a page is bookmarked, and react-router's deferred rendering counts as a first render, so that the server render and what's in react-router's deferred renders should be the same?

It would be great to have some graphic or other explanation of this.

What's not obvious is that we can server render what the deferred rendering will be.

---

*[docs/additional-reading/code-splitting.md, line 26 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfhS9KpuBoCbLNl8rJ:-KWfhS9L1KGHufMzF6fL:bgbcu99) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L26)):*
> ```Markdown
> Here's an example of how you might use this in practice:
> 
> page.html.erb
> ```

Maybe use `####` to format the file names.

---

*[docs/additional-reading/code-splitting.md, line 42 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWg1xeyaU6sdPgsjVJ0:-KWg1xezcCbhcod_sRD0:b-xzn0gb) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L42)):*
> ```Markdown
> ReactOnRails.registerStore({applicationStore});
> ReactOnRails.register({NavigationApp});
> ReactOnRails.registerRenderer({RouterApp});
> ```

We need to be very clear that RouterApp will handle calling ReactDOM to create the react element, attached to a DOM node.


Should we be showing some `serverRegistration.js` page?

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfhu1Juo0cn7ut8vnM:-KWfhu1Juo0cn7ut8vnN:b5w2sak) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*
> ```Markdown
> ### Caveats
> 
> If you're going to try to do code splitting with server rendered routes, it's important that you have seperate webpack configurations for client and server. The code splitting happens for the client, but the server should one big file.
> ```

`should one big file`

maybe:

`should be one file containing all the JavaScript code`

---

*[docs/api/javascript-api.md, line 64 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfyvFaewdrl3ZvfWk3:-KWfyvFaewdrl3ZvfWk4:bbjg8zy) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/api/javascript-api.md#L64)):*
> ```Markdown
>    * Allows registration of renderers. A renderer is a function that accept three args:
>    * props, railsContext, and domNodeId, and is responsible for rendering a component
>    * to the DOM. Not available on the server. For one possible use case see:
> ```

We might be able to throw an error if this is called on the server since we'll know we're server rendering with the rails context.

---

*[node_package/src/ReactOnRails.js, line 114 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWg02OgPZXAERHR08zY:-KWg02OgPZXAERHR08zZ:bshox4f) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/node_package/src/ReactOnRails.js#L114)):*
> ```JavaScript
>     ComponentRegistry.registerRenderer(renderers);
>   },
> ```

Please move this next to the definition immediately below `registerStore`

---

*[spec/dummy/client/app/startup/ManualRenderAppRenderer.jsx, line 12 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWg3hVJjt3gT__VrIbi:-KWg3hVK5ZTLNzL_Z36l:bsoij57) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/spec/dummy/client/app/startup/ManualRenderAppRenderer.jsx#L12)):*
> ```JSX
>   );
> 
>   ReactDOM.render(reactElement, document.getElementById(domNodeId));
> ```

I feel that we need the example to show react-router.

We use react-router in the dummy apps.

Another consideration is support for react-router v4.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KWg3wOzzWxHDqnaQnYK:b-s95lwj)*

alexfedoseev:
I have one comment below.

And also one more comment on the global API: I'm not a big fan of owning key deps (as `ReactDOM`) w/o strong reason (is there any?), so leaving it in user land by default is IMO a better design decision, as we had already few issues w/ it.

---

Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWlkQweSroR6kHCdgMQ:-KWlkQweSroR6kHCdgMR:b-8c5kkk) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*
> ```Markdown
> ### Caveats
> 
> If you're going to try to do code splitting with server rendered routes, it's important that you have seperate webpack configurations for client and server. The code splitting happens for the client, but the server should one big file.
> ```

I don't think the main point here is different webpack configs (as client/server configs are usually different regardless of whether or not code splitting is used), but the fact that it requires different `routes` files for the client and the server (w/ same routes, but different components resolvers), which is most uncommon & untransparent part here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KWlgqZs1q_6cMrqUCm2:b-d2srgx)*

jtibbertsma:
Review status: 13 of 18 files reviewed at latest revision, 12 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 18 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfyRX0MArujnkeiwPJ:-KWpXXuwtVr95OwpVmEB:b-se8gwn) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L18)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> I'm a bit confused about the interaction between deferred rendering and react-router. Is the idea that a page is bookmarked, and react-router's deferred rendering counts as a first render, so that the server render and what's in react-router's deferred renders should be the same?
> 
> It would be great to have some graphic or other explanation of this.
> 
> What's not obvious is that we can server render what the deferred rendering will be.
</details>
The idea is that `match` from react-router is async; it fetches the component using the `getComponent` method that the user provides with the route definition, and then passes the props to the callback that are needed to do the complete render. Then we do the first render inside of the callback, so that the first render is the same as the server render.

The server render matches the deferred render because the server bundle is a single file, and so it doesn't need to wait for anything to be fetched.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: 13 of 18 files reviewed at latest revision, 12 unresolved discussions.

---

*[docs/api/javascript-api.md, line 64 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfyvFaewdrl3ZvfWk3:-KWpYzi2Ywy2yScceO4v:b-i71hl3) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/api/javascript-api.md#L64)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> We might be able to throw an error if this is called on the server since we'll know we're server rendering with the rails context.
</details>
There is an error thrown if it's used on the server, but only if the corresponding `react_component` is called in a view. See serverRenderReactComponent.js and serverRenderReactComponent.test.js.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: 13 of 18 files reviewed at latest revision, 12 unresolved discussions.

---

*[spec/dummy/client/app/startup/ManualRenderAppRenderer.jsx, line 12 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWg3hVJjt3gT__VrIbi:-KWpZV0Z-cCDHsOyngwK:b-i1cm0b) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/spec/dummy/client/app/startup/ManualRenderAppRenderer.jsx#L12)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> I feel that we need the example to show react-router.
> 
> We use react-router in the dummy apps.
> 
> Another consideration is support for react-router v4.
</details>
I'll expand the example a bit to use react-router and have a code split route, and I'll add some test cases. I should have to do it on Saturday.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

justin808:
Looking good!

---

Reviewed 5 of 5 files at r6.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 18 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfyRX0MArujnkeiwPJ:-KWqKwwWel2Cg5LYcYxE:blc285r) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L18)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary>
> The idea is that `match` from react-router is async; it fetches the component using the `getComponent` method that the user provides with the route definition, and then passes the props to the callback that are needed to do the complete render. Then we do the first render inside of the callback, so that the first render is the same as the server render.
> 
> The server render matches the deferred render because the server bundle is a single file, and so it doesn't need to wait for anything to be fetched.
</details>
Consider putting this explanation in your MD file here! the key is that the first render is done by react-router, so it must match the server.

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWlkQweSroR6kHCdgMQ:-KWqM-c0flcKSpCqmkQc:b-lex6ll) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*
<details><summary><i>Previously, alexfedoseev (Alex Fedoseev) wrote…</i></summary>
> I don't think the main point here is different webpack configs (as client/server configs are usually different regardless of whether or not code splitting is used), but the fact that it requires different `routes` files for the client and the server (w/ same routes, but different components resolvers), which is most uncommon & untransparent part here.
</details>
@jtibbertsma Please address @alexfedoseev. 

---

*[docs/additional-reading/code-splitting.md, line 58 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWqLaOb1zbTYEFv2FiE:-KWqLaOc1kc2T7XjaKKB:b-8dtzy4) ([raw file](https://github.com/shakacode/react_on_rails/blob/174a3eaaaf77005b8785220cae70ed44873b33f5/docs/additional-reading/code-splitting.md#L58)):*
> ```Markdown
> });
> ```
> Note that you should not use `registerRenderer` on the server. Instead, use `register` like normal. For an example of how to set up an app for server rendering, see the [react router docs](react-router.md).
> ```

We'll know if we're server rendering. Could we allow the same registration code for client and server, and just ensure that we don't do the deferred render thing on the server?

Needing this to be different for the server rendering seems like a big **gotcha**.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KWqLKDmnxTH7LmHiaZJ:br2sbr6)*

justin808:
@jtibbertsma I'd like to get this in the next release, maybe this weekend!

jtibbertsma:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWlkQweSroR6kHCdgMQ:-KWv8IM9D_9QbWJoUqxB:b-h4hohh) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> @jtibbertsma Please address @alexfedoseev. 
</details>
I did amend code-splitting.md to mention the seperate route files. Honestly, I don't really know if you need two route files, that's just the way I did it. Maybe there's another way to get it to work. The key is that the server bundle needs to be a single file, not code-split.

---

*[docs/additional-reading/code-splitting.md, line 58 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWqLaOb1zbTYEFv2FiE:-KWv9b2OX-0RRlHOIKI-:b8rem5y) ([raw file](https://github.com/shakacode/react_on_rails/blob/174a3eaaaf77005b8785220cae70ed44873b33f5/docs/additional-reading/code-splitting.md#L58)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> We'll know if we're server rendering. Could we allow the same registration code for client and server, and just ensure that we don't do the deferred render thing on the server?
> 
> Needing this to be different for the server rendering seems like a big **gotcha**.
</details>
I don't know if it's a good idea to allow the same registration code on the server. The user provided function is calls `ReactDOM.render` and references `document ` in order to get the DOM node to render into. I suppose that I could rework it so that instead of passing `domNodeId` into the user provided function, it would pass a callback that the user would call when they've created a react element. But I like it better the way it is now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: 17 of 18 files reviewed at latest revision, 5 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 18 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWfyRX0MArujnkeiwPJ:-KWvD97mb487QQamW1YY:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L18)):*
<details><summary><i>Previously, justin808 (Justin Gordon) wrote…</i></summary>
> Consider putting this explanation in your MD file here! the key is that the first render is done by react-router, so it must match the server.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

justin808:
Review status: 17 of 18 files reviewed at latest revision, 5 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 58 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWqLaOb1zbTYEFv2FiE:-KWvDD-lH5Diip7_phft:b-rwtzql) ([raw file](https://github.com/shakacode/react_on_rails/blob/174a3eaaaf77005b8785220cae70ed44873b33f5/docs/additional-reading/code-splitting.md#L58)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary>
> I don't know if it's a good idea to allow the same registration code on the server. The user provided function is calls `ReactDOM.render` and references `document ` in order to get the DOM node to render into. I suppose that I could rework it so that instead of passing `domNodeId` into the user provided function, it would pass a callback that the user would call when they've created a react element. But I like it better the way it is now.
</details>
@alexfedoseev @robwise @alleycat-at-git Any opinion?

This is an important feature that we'll probably use.

The key question for me is if we should have separate registerRenderer and register APIs.

Since this is an API change, I want to make sure we get this right.

Possibly we'll implement this as a beta release to get some community feedback?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: 12 of 25 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[spec/dummy/spec/features/integration_spec.rb, line 170 at r8](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWzHQ7zDUy_44GkV8ri:-KWzHQ7zDUy_44GkV8rj:bjg2g9w) ([raw file](https://github.com/shakacode/react_on_rails/blob/fa49c6bb7e5fced2bd3b49f5bda4fe9d269c5f2f/spec/dummy/spec/features/integration_spec.rb#L170)):*
> ```Ruby
> end
> 
> # feature "Code Splitting w/ Server Rendering", :js do
> ```

I want to write a test that shows that there is no client/server checksum mismatch when loading the `/deferred_render_with_server_rendering/async_page` route. When React detects a mismatch, it logs an error and then rerenders everything. So I can't test for this by looking for something on the page. I was thinking I could check if any errors were logged, but there doesn't seem to be a way to do this with poltergeist. Any ideas about how I could test for this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: 9 of 25 files reviewed at latest revision, 6 unresolved discussions.

---

*[spec/dummy/spec/features/integration_spec.rb, line 170 at r8](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWzHQ7zDUy_44GkV8ri:-KXCZbz2R55EIAfBdBYn:b38kho9) ([raw file](https://github.com/shakacode/react_on_rails/blob/fa49c6bb7e5fced2bd3b49f5bda4fe9d269c5f2f/spec/dummy/spec/features/integration_spec.rb#L170)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary>
> I want to write a test that shows that there is no client/server checksum mismatch when loading the `/deferred_render_with_server_rendering/async_page` route. When React detects a mismatch, it logs an error and then rerenders everything. So I can't test for this by looking for something on the page. I was thinking I could check if any errors were logged, but there doesn't seem to be a way to do this with poltergeist. Any ideas about how I could test for this?
</details>
I figured out a way to do this. Just check the root element for the data-react-checksum attribute. If it's there, there wasn't a checksum mismatch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
@justin808 I've been thinking about this, and I've decided that there's actually no need to have a new API for this. Instead of having `registerRenderer`, `register` could be modified so that it checks the number of arguments of the generator function. If the function takes three arguments, it would be considered a renderer. This is cleaner, because there's no need for a new API, but we would still be able to control the call to `ReactDOM.render` if we want.

justin808:
@jtibbertsma I look forward to seeing what you come up with!

justin808:
:lgtm:

@jtibbertsma This is looking amazing. Keep me posted when you want more review.

---

Reviewed 8 of 12 files at r8, 1 of 3 files at r9, 11 of 14 files at r12, 1 of 1 files at r13, 2 of 2 files at r14.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 22 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXtKEgelanYLI_Ckei:-KXXtKEhz3437NuHTdzW:b-dysycv) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L22)):*
> ```Markdown
> ### The solution
> 
> To prevent this, you have to wait until the code chunk is fetched before doing the initial render on the client side. To accomplish this, react on rails allows you to register a renderer. This works just like registering a generator function, except that the function you pass takes three arguments: `renderer(props, railsContext, domNodeId)`, and is responsible for calling `ReactDOM.render` to render the component to the DOM. React on rails will automatically detect when a generator function takes three arguments, and will not call `ReactDOM.render`, instead allowing you to control the initial render yourself.
> ```

Cool! I just hope this is not to subtle.

---

*[docs/additional-reading/code-splitting.md, line 51 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXtw9VJE8eGMlVIwvL:-KXXtw9VJE8eGMlVIwvM:bg4rzdy) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L51)):*
> ```Markdown
> import ReactOnRails from 'react-on-rails';
> import NavigationApp from './NavigationApp';
> import RouterApp from './RouterAppServer';
> ```

I'd put a comment above this line and the file above, to highlight the difference!

---

*[docs/additional-reading/code-splitting.md, line 60 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXu-H5NHycRZfgwZ4v:-KXXu-H5NHycRZfgwZ4w:b-fpxeep) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L60)):*
> ```Markdown
> });
> ```
> Note that you should not register a renderer on the server, since there won't be a domNodeId when we're server rendering. For an example of how to set up an app for server rendering, see the [react router docs](react-router.md).
> ```

and comment here on how RouterApp is two different imports given client vs server.

---

*[docs/additional-reading/code-splitting.md, line 101 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXuVh9Dy-ZABmRk7lY:-KXXuVh9Dy-ZABmRk7lZ:b-lmwn9t) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L101)):*
> ```Markdown
> The idea is that match from react-router is async; it fetches the component using the getComponent method that you provide with the route definition, and then passes the props to the callback that are needed to do the complete render. Then we do the first render inside of the callback, so that the first render is the same as the server render.
> 
> The server render matches the deferred render because the server bundle is a single file, and so it doesn't need to wait for anything to be fetched.
> ```

How does this fit into React Router v4? 

https://github.com/ReactTraining/react-router/tree/v4

I'm OK making this feature only work with v4.

Or we should make the docs work with v3 and v4. 

CC: @alexfedoseev @robwise 

---

*[docs/additional-reading/code-splitting.md, line 107 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXv7IjsHJ26VKbPLSs:-KXXv7IjsHJ26VKbPLSt:bgiyhtr) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L107)):*
> ```Markdown
> If you're going to try to do code splitting with server rendered routes, you'll probably need to use seperate route definitions for client and server to prevent code splitting from happening for the server bundle. The server bundle should be one file containing all the JavaScript code.
> 
> The reason is we do server rendering with ExecJS, which is not capable of doing anything asynchronous. It would be impossible to asyncronously fetch a code chunk while server rendering. See [this issue](https://github.com/shakacode/react_on_rails/issues/477) for a discussion.
> ```

FYI -- we're investigating alternatives to execJS. Possibly, @jtibbertsma, we could get you involved.  

https://github.com/shakacode/react_on_rails/issues/625

---

*[docs/additional-reading/code-splitting.md, line 109 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXztmRr75BsmlOvXYq:-KXXztmSwg7yNFd6Vyr7:b-v6bty6) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L109)):*
> ```Markdown
> The reason is we do server rendering with ExecJS, which is not capable of doing anything asynchronous. It would be impossible to asyncronously fetch a code chunk while server rendering. See [this issue](https://github.com/shakacode/react_on_rails/issues/477) for a discussion.
> 
> Also, do not attempt to register a renderer on the server. Instead, register either a generator function or a component. If you register a renderer in the server bundle, you'll get an error when react on rails tries to server render the component.
> ```

also point out to have different entry files for the server and client webpack configs

---

*[docs/additional-reading/code-splitting.md, line 123 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-37ASNjvSsnAC0DE:-KXY-37ASNjvSsnAC0DF:b-nscjhr) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L123)):*
> ```Markdown
> ```
> 
> This causes Webpack to prepend the code chunk filename with `/assets/` in the request url. The react on rails sets up the webpack config to put webpack bundles in `app/assets/javascripts/webpack`, and modifies `config/initializers/assets.rb` so that rails detects the bundles. This means that when we prepend the request url with `/assets/`, rails will know what webpack is asking for.
> ```

Suppose this is not done. How hard is it to debug this?

Maybe show the error here so that a google search will find this page.

---

*[node_package/src/serverRenderReactComponent.js, line 18 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-cOWvNbilzSOTpa0:-KXY-cOWvNbilzSOTpa1:b-j59jym) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/src/serverRenderReactComponent.js#L18)):*
> ```JavaScript
>     const componentObj = ComponentRegistry.get(name);
>     if (componentObj.isRenderer) {
>       throw new Error(`Detected a renderer while server rendering component '${name}'`);
> ```

Let's make the error a bit more explicit, even including a doc link.

---

*[node_package/tests/ComponentRegistry.test.js, line 59 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY00dqxVfTRKXBlr8q:-KXY00dqxVfTRKXBlr8r:b-m7i9m0) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/tests/ComponentRegistry.test.js#L59)):*
> ```JavaScript
>   const expected = { name: 'C4', component: C4, generatorFunction: true, isRenderer: true };
>   assert.deepEqual(actual, expected,
>     'ComponentRegistry registers and retrieves renderers');
> ```

How about a test with 2 params?

---

*[spec/dummy/client/app/startup/DeferredRenderAppRenderer.jsx, line 3 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY0OFxHonn1xzO-cdz:-KXY0OFxHonn1xzO-ce-:b-lw9el) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/app/startup/DeferredRenderAppRenderer.jsx#L3)):*
> ```JSX
> import React from 'react';
> import ReactDOM from 'react-dom';
> import { match, Router, browserHistory } from 'react-router';
> ```

how will this change with https://github.com/ReactTraining/react-router/tree/v4 ?

---

*[spec/dummy/client/app/startup/DeferredRenderAppServer.jsx, line 38 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY0ag1E9PVxGTMVGDn:-KXY0ag1E9PVxGTMVGDo:b-izxnfj) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/app/startup/DeferredRenderAppServer.jsx#L38)):*
> ```JSX
> };
> 
> export default DeferredRenderAppServer;
> ```

Let's have some reference to these example files in the doc MD file.

---

*[spec/dummy/spec/features/integration_spec.rb, line 176 at r8](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY1E8id1yDqzZm6Ium:-KXY1E8id1yDqzZm6Iun:b-30ms2t) ([raw file](https://github.com/shakacode/react_on_rails/blob/fa49c6bb7e5fced2bd3b49f5bda4fe9d269c5f2f/spec/dummy/spec/features/integration_spec.rb#L176)):*
> ```Ruby
> #     expect(page).to_not have_errors
> #   end
> # end
> ```

we should uncomment the test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KXY1MdHfpN2yZT1qUsv:b7vq2ey)*

spec/dummy/client/webpack.client.rails.build.config.js
```diff
@@ -12,6 +12,7 @@ const devBuild = process.env.NODE_ENV !== 'production';
 config.output = {
   filename: '[name]-bundle.js',
   path: '../app/assets/webpack',
+  publicPath: '/assets/',
```
justin808, line 15: @jtibbertsma Should we put the files separately for code splitting? i would think there would be another change to some webpack file.

jtibbertsma:
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 22 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXtKEgelanYLI_Ckei:-KXn--w2Wh_h2TcqQZEk:b-tz3tmc) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L22)):*

Cool! I just hope this is not to subtle.
</blockquote></details>

I like this way better than having a separate API.

---

*[docs/additional-reading/code-splitting.md, line 51 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXtw9VJE8eGMlVIwvL:-KXmFZjYZmKo8n2fCnZf:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L51)):*

I'd put a comment above this line and the file above, to highlight the difference!
</blockquote></details>

Done.

---

*[docs/additional-reading/code-splitting.md, line 60 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXu-H5NHycRZfgwZ4v:-KXmFwNH2fwOpI88lCiG:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L60)):*

and comment here on how RouterApp is two different imports given client vs server.
</blockquote></details>

Done.

---

*[docs/additional-reading/code-splitting.md, line 101 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXuVh9Dy-ZABmRk7lY:-KXmG1G1Vi_QTmC-zq_9:bvxatw9) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L101)):*

How does this fit into React Router v4? 

https://github.com/ReactTraining/react-router/tree/v4

I'm OK making this feature only work with v4.

Or we should make the docs work with v3 and v4. 

CC: @alexfedoseev @robwise 
</blockquote></details>

It's really annoying when libraries completely change their APIs. I suppose that all the docs will have to be updated for v4. I don't know how code splitting will work with v4 yet. I'll do some more research on it eventually. Anyway, it's still in alpha, so I'm not going to worry too much about this right now.

---

*[docs/additional-reading/code-splitting.md, line 107 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXv7IjsHJ26VKbPLSs:-KXmGzkoAK9tQPewiYYU:bb6u032) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L107)):*

FYI -- we're investigating alternatives to execJS. Possibly, @jtibbertsma, we could get you involved.  

https://github.com/shakacode/react_on_rails/issues/625
</blockquote></details>

Cool

---

*[docs/additional-reading/code-splitting.md, line 109 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXztmRr75BsmlOvXYq:-KXmHNQIIuDZqMJBIpbo:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L109)):*

also point out to have different entry files for the server and client webpack configs
</blockquote></details>

Done.

---

*[docs/additional-reading/code-splitting.md, line 123 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-37ASNjvSsnAC0DE:-KXmHQWpsD1SQ4FpSIN_:b-uon2r7) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L123)):*

Suppose this is not done. How hard is it to debug this?

Maybe show the error here so that a google search will find this page.
</blockquote></details>

If you don't put `/assets/` as the public path, webpack will send the request to `/{filename}`, and Rails will handle the request, since NGINX or whatever won't see it as a request for an asset. The exact error you'll get depends on how your Rails router is set up. You'll probably just get a 404 error, but since you're probably using react-router, you might have some sort of catch all route. It depends on what your catch all route does.

Another consideration is how you do error handling with code splitting. With webpack v1, there's really no way to implement error handling, but in v2 you can implement error handling by calling `.catch` on the Promise returned by `System .import`. Since we're using v1, I didn't mention error handling.

I went ahead and added part of this explanation and a possible error message that you might get.

---

*[node_package/src/serverRenderReactComponent.js, line 18 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-cOWvNbilzSOTpa0:-KXmcfuaneWUwVBUueyJ:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/src/serverRenderReactComponent.js#L18)):*

Let's make the error a bit more explicit, even including a doc link.
</blockquote></details>

Done.

---

*[node_package/tests/ComponentRegistry.test.js, line 59 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY00dqxVfTRKXBlr8q:-KXn-Fa-qn4kUL3ulyAG:bz4pvo4) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/tests/ComponentRegistry.test.js#L59)):*

How about a test with 2 params?
</blockquote></details>

Do you have something in mind? I can't think of anything to test here other than checking that isRenderer key is set to true.

---

*[spec/dummy/client/webpack.client.rails.build.config.js, line 15 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWzGq4pA43jIx4qFwbT-r14-15:-KXmlW1LB-cPyfWxx1A1:b-8w2e51) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/webpack.client.rails.build.config.js#L15)):*

@jtibbertsma Should we put the files separately for code splitting? i would think there would be another change to some webpack file.
</blockquote></details>

Do you mean that the code chunks should go to a different folder than app-bundle and vendor-bundle?

---

*[spec/dummy/client/app/startup/DeferredRenderAppRenderer.jsx, line 3 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY0OFxHonn1xzO-cdz:-KXmkBFO-B1gJJZo4ym_:b-8h74sm) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/app/startup/DeferredRenderAppRenderer.jsx#L3)):*

how will this change with https://github.com/ReactTraining/react-router/tree/v4 ?
</blockquote></details>

I'll have to look into it.

---

*[spec/dummy/client/app/startup/DeferredRenderAppServer.jsx, line 38 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY0ag1E9PVxGTMVGDn:-KXmzTVRpILqXOfGfUHE:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/app/startup/DeferredRenderAppServer.jsx#L38)):*

Let's have some reference to these example files in the doc MD file.
</blockquote></details>

Done.

---

*[spec/dummy/spec/features/integration_spec.rb, line 176 at r8](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY1E8id1yDqzZm6Ium:-KXmmHXKDu0QY19KshAR:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/fa49c6bb7e5fced2bd3b49f5bda4fe9d269c5f2f/spec/dummy/spec/features/integration_spec.rb#L176)):*

we should uncomment the test
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

justin808:
:lgtm:

This is super close to being ready.

@jtibbertsma please see my comments and let me know if you want me to merge this. This is really an impressive PR!

---

Reviewed 3 of 3 files at r15.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWlkQweSroR6kHCdgMQ:-KXnd70pA4wuh0YgkOqt:bnt19mv) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

I did amend code-splitting.md to mention the seperate route files. Honestly, I don't really know if you need two route files, that's just the way I did it. Maybe there's another way to get it to work. The key is that the server bundle needs to be a single file, not code-split.
</blockquote></details>

@jtibbertsma We really want the spec/dummy example to show code splitting on the client side. Why? This will show the usefulness of this technique!

---

*[docs/additional-reading/code-splitting.md, line 101 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXXuVh9Dy-ZABmRk7lY:-KXncpHnW6tZAsFvBI-7:b-djj4ja) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L101)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

It's really annoying when libraries completely change their APIs. I suppose that all the docs will have to be updated for v4. I don't know how code splitting will work with v4 yet. I'll do some more research on it eventually. Anyway, it's still in alpha, so I'm not going to worry too much about this right now.
</blockquote></details>

@jtibbertsma The changes are really great and worth the churn. Yes, we will eventually update the docs.  I think the technique is the same.

---

*[docs/additional-reading/code-splitting.md, line 123 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-37ASNjvSsnAC0DE:-KXndS_RhGWLQA-1oVUb:bbew3jh) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L123)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

If you don't put `/assets/` as the public path, webpack will send the request to `/{filename}`, and Rails will handle the request, since NGINX or whatever won't see it as a request for an asset. The exact error you'll get depends on how your Rails router is set up. You'll probably just get a 404 error, but since you're probably using react-router, you might have some sort of catch all route. It depends on what your catch all route does.

Another consideration is how you do error handling with code splitting. With webpack v1, there's really no way to implement error handling, but in v2 you can implement error handling by calling `.catch` on the Promise returned by `System .import`. Since we're using v1, I didn't mention error handling.

I went ahead and added part of this explanation and a possible error message that you might get.
</blockquote></details>

Webpack v2 is close enough to being published that it's worth considering for the documentation.

---

*[node_package/src/serverRenderReactComponent.js, line 20 at r15](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXndixShbnU8fj9og29:-KXndixShbnU8fj9og2A:b500rwb) ([raw file](https://github.com/shakacode/react_on_rails/blob/15c0618ceef696171009eb67790c0c1a4a916bbc/node_package/src/serverRenderReactComponent.js#L20)):*
> ```JavaScript
>       throw new Error(`\
> Detected a renderer while server rendering component '${name}'. \
> See https://github.com/shakacode/react_on_rails#renderer-functions`);
> ```

is this better than the code-spliting.md doc?

---

*[node_package/tests/ComponentRegistry.test.js, line 59 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY00dqxVfTRKXBlr8q:-KXndoyQzu5xbGFdrMKr:b-6nz7uu) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/tests/ComponentRegistry.test.js#L59)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

Do you have something in mind? I can't think of anything to test here other than checking that isRenderer key is set to true.
</blockquote></details>

that's exactly what I was thinking of, except, isRenderer should be **false**

---

*[spec/dummy/client/webpack.client.rails.build.config.js, line 15 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWzGq4pA43jIx4qFwbT-r14-15:-KXndw2DuCHbd4jQZUZa:b3t2abi) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/webpack.client.rails.build.config.js#L15)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

Do you mean that the code chunks should go to a different folder than app-bundle and vendor-bundle?
</blockquote></details>

Not a separate folder, but we could have app-bundle-router, and app-bundle-route-b.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KXneK5CxfAc6uTKRosY:b-5b98qp)*

jtibbertsma:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 85 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWlkQweSroR6kHCdgMQ:-KXnhFHuI1CWLuDOBRLc:bwr1aoj) ([raw file](https://github.com/shakacode/react_on_rails/blob/6bcf72f07c42144c795ba83e6fe31c7a18ceab5b/docs/additional-reading/code-splitting.md#L85)):*

@jtibbertsma We really want the spec/dummy example to show code splitting on the client side. Why? This will show the usefulness of this technique!
</blockquote></details>

I did put a code splitting example in spec/dummy.

---

*[node_package/src/serverRenderReactComponent.js, line 20 at r15](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXndixShbnU8fj9og29:-KXnjBM7KC8b0lmnx4HL:b1cd6lz) ([raw file](https://github.com/shakacode/react_on_rails/blob/15c0618ceef696171009eb67790c0c1a4a916bbc/node_package/src/serverRenderReactComponent.js#L20)):*

is this better than the code-spliting.md doc?
</blockquote></details>

Yeah. The information there is more relevant to the error message, and it links to code-splitting.md

---

*[node_package/tests/ComponentRegistry.test.js, line 59 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY00dqxVfTRKXBlr8q:-KXnkLjdazKoSc1sksqa:b-ljw7vt) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/tests/ComponentRegistry.test.js#L59)):*

that's exactly what I was thinking of, except, isRenderer should be **false**
</blockquote></details>

Ok, I see. I'll add another test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[docs/additional-reading/code-splitting.md, line 123 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY-37ASNjvSsnAC0DE:-KXqpY2DlCMGd8uawmCF:bt5q42) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/docs/additional-reading/code-splitting.md#L123)):*

Webpack v2 is close enough to being published that it's worth considering for the documentation.
</blockquote></details>

Ok, I added a comment about error handling.

---

*[node_package/tests/ComponentRegistry.test.js, line 59 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KXY00dqxVfTRKXBlr8q:-KXquEWunyAcgwluNR8S:b-rcfh87) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/node_package/tests/ComponentRegistry.test.js#L59)):*
<details><summary><i>Previously, jtibbertsma (Joseph Tibbertsma) wrote…</i></summary><blockquote>

Ok, I see. I'll add another test
</blockquote></details>

I added another test case.

---

*[spec/dummy/client/webpack.client.rails.build.config.js, line 15 at r14](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-KWzGq4pA43jIx4qFwbT-r14-15:-KXqtXUoEuwzr3Bt8I-a:bi2n9d3) ([raw file](https://github.com/shakacode/react_on_rails/blob/c9a3e17c63e101bfaac19d9be6850877906762f7/spec/dummy/client/webpack.client.rails.build.config.js#L15)):*

Not a separate folder, but we could have app-bundle-router, and app-bundle-route-b.
</blockquote></details>

I think this is fine how it is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
@justin808 Ready to merge? Should I rebase and squash the commits?

justin808:
@jtibbertsma If you'd like to create a nice squashed commit with an awesome, that would be super!

justin808:
Reviewed 2 of 2 files at r16.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581)*

jtibbertsma:
@justin808 Done

justin808:
:lgtm:

---

Reviewed 6 of 6 files at r17.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/581#-:-KXrkpfiV5dHfh6YEe33:bnfp4nl)*

PR #593, opened 2016-11-07
Fix broken URL for gitbooks

VagishVela:
The URL is broken on https://shakacode.gitbooks.io/react-on-rails/content/docs/additional-reading/hot-reloading-rails-development.html since it is looking relatively and not at the repo.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/593)


VagishVela:
@justin808 Noticed a broken URL on gitbooks so made it absolute on here.

justin808:
Thanks @VagishVela!

PR #612, opened 2016-11-20
Remove Context calling React.Component constructor

justin808:
The React.Component takes one param in a call to super in the
constructor.


robwise:
lgtm

PR #689, opened 2017-01-18
Added property renderedHtml to return gen func

justin808:
This is to support React Router v4, which needs the result of calling
renderToString.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/689)

### DIFF
--- a/spec/dummy/spec/features/integration_spec.rb
+++ b/spec/dummy/spec/features/integration_spec.rb
@@ -176,6 +176,15 @@ feature "Code Splitting w/ Server Rendering", :js do
   end
 end
 
+feature "renderedHtml from generator function", :js do
+  subject { page }
+  background { visit "/rendered_html" }
+  scenario "renderedHtml should not have any errors" do
+    expect(subject).to have_text "Props: {\"hello\":\"world\"}"
+    expect(subject.html).to include("[SERVER] RENDERED RenderedHtml to dom node with id")
+  end
+end
+
 shared_examples "React Component Shared Store" do |url|
   subject { page }
   background { visit url }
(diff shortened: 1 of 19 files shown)


justin808:
@robwise @alexfedoseev can I get a quick review on the JS parts?
@mapreal19 @samnang please take a quick peek at the Ruby code.

spec/dummy/spec/features/integration_spec.rb
```diff
@@ -176,6 +176,15 @@ def change_text_expect_dom_selector(dom_selector)
   end
 end
 
+feature "renderedHtml from generator function", :js do
+  subject { page }
```
samnang, line 180: Do you prefer to call it `subject` instead of `page`? To me, page is more explicit and specific in this context.

spec/dummy/spec/features/integration_spec.rb
```diff
@@ -176,6 +176,15 @@ def change_text_expect_dom_selector(dom_selector)
...
+  background { visit "/rendered_html" }
+  scenario "renderedHtml should not have any errors" do
+    expect(subject).to have_text "Props: {\"hello\":\"world\"}"
+    expect(subject.html).to include("[SERVER] RENDERED RenderedHtml to dom node with id")
```
samnang, line 184: Why do we have to check with the whole string here instead of putting assertion on what we want to verify each part.

mapreal19:
Reviewed 6 of 19 files at r1.
Review status: 6 of 18 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/689)*

spec/dummy/spec/features/integration_spec.rb
justin808: I'm being consistent with other tests in this file.

spec/dummy/spec/features/integration_spec.rb
justin808: Because this is in a script tag.

alexfedoseev:
@justin808 I tried only client part of `react-router@4` yet. Looking at docs for server rendering, api seems kinda strange, I don't get why 404 case requires second render. But anyway, talking about integration w/ `react_on_rails`— it LGTM.

PR #690, opened 2017-01-18
Fix incorrect "this" references of Node.js SSR

nostophilia:
The default Node.js server file created by the generator throw errors of `undefined`. Arrow functions don't bind `this`, so it's not referencing the `handler` object as expected but the global object likely. So something like this won't work and throw an error.
```js
Handler.prototype.initialize = () => {
  console.log(`Processing ${this.queue.length} pending requests`);
  ......
}
```
We can either use ES6 class or write something like `Handler.prototype.handle = function (connection) ...`. I'd prefer the former one.
 
(BTW, I'm using Node.js as backend because I'm experimenting server-side rendering with `Relay` or `Apollo` and `ExecJS` doesn't seem to support this. But with the `Node` solution, I don't like the fact that there seems to be some unnecessary communications between `Rails` and `Node` due to the way those GraphQL client SSR works. I wish there can be a way to **hydrate** Apollo store from `Rails` like how you did with `Redux` SSR. Maybe I will update you later on another issue after some researching)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/690)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,13 @@ All notable changes to this project's source code will be documented in this fil
 Contributors: please follow the recommendations outlined at [keepachangelog.com](http://keepachangelog.com/). Please use the existing headings and styling as a guide, and add a link for the version diff at the bottom of the file. Also, please update the `Unreleased` link to compare to the latest release version.
 
 ## [Unreleased]
-*Please add entries here for your pull requests.*
+### Fixed
+- Fix incorrect "this" references of Node.js SSR [#690] by [nostophilia](https://github.com/nostophilia)
 
 ## [6.4.0] - 2017-1-12
 
 ### Possible Breaking Change
-- Since foreman is no longer a dependency of the React on Rails gem, please run `gem install foreman`. If you are using rvm, you may wish to run `rvm @global do gem install foreman` to install foreman for all your gemsets. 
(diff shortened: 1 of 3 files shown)


justin808:
@alleycat-at-git Please review

alleycat-at-git:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/690#-:-Kb9qzkCHaDARuM3zqrY:bnfp4nl)*

justin808:
@nostophilia Thanks! Please add the appropriate changelog entry and I'll merge this.

justin808:
slight request

---

Reviewed 1 of 1 files at r2.

---

*[CHANGELOG.md, line 7 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/690#-Kbmz0r9KSMvBJl_eb9Q:-Kbmz0r9KSMvBJl_eb9R:bpcagee) ([raw file](https://github.com/shakacode/react_on_rails/blob/53af84126bb1a7c10a89849625938e84775f344e/CHANGELOG.md#L7)):*
> ```Markdown
> 
> ## [Unreleased]
> *Please add entries here for your pull requests.*
> ```

Please don't remove this line!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/690#-:-Kbmz4IVemNryZvEM6T_:b-urvmq1)*

PR #695, opened 2017-01-26
Alexey/render deferred

justin808:
The purpose of this change is to allow an option on a react_component to be rendered with an explicit call rather than automatically when page loaded.

We thought this one was needed for rendering React on Rails components created on the view with react_component to be rendered by another library asynchronously, but now we're not sure this is required.

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/695)

### DIFF
--- a/lib/react_on_rails/react_component/options.rb
+++ b/lib/react_on_rails/react_component/options.rb
@@ -33,6 +33,10 @@ module ReactOnRails
         retrieve_key(:prerender)
       end
 
+      def render_deferred
+        options[:render_deferred] || false
+      end
+
       def trace
         retrieve_key(:trace)
       end
@@ -50,7 +54,8 @@ module ReactOnRails
           component_name: name,
           props: props,
           trace: trace,
-          dom_id: dom_id
+          dom_id: dom_id,
+          render_deferred: render_deferred
         }
       end
 
--- a/lib/react_on_rails/version_checker.rb
+++ b/lib/react_on_rails/version_checker.rb
@@ -68,7 +68,7 @@ module ReactOnRails
       end
 
       def relative_path?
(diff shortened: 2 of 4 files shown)


justin808:
LGTM with one tiny change.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/version_checker.rb, line 71 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/695#-KbRuEddNT-yovg1INsq:-KbRuEddNT-yovg1INsr:b-39soph) ([raw file](https://github.com/shakacode/react_on_rails/blob/3f84aa854f018f9bdc269b9382075ab48b6e2a4f/lib/react_on_rails/version_checker.rb#L71)):*
> ```Ruby
> 
>       def relative_path?
>         raw.match(%r{(\.\.|\Afile:///)}).present?
> ```

why did this change?

---

*[node_package/src/ReactOnRails.js, line 89 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/695#-KbRuQ0WqyIyHqKn22M2:-KbRuQ0WqyIyHqKn22M3:blfvdl4) ([raw file](https://github.com/shakacode/react_on_rails/blob/41f15abb8440be01f19b49db970de8b027b19e51/node_package/src/ReactOnRails.js#L89)):*
> ```JavaScript
> 
>   /**
>    * Render react_components with option data-renderdeferred: true. These components are not
> ```

Should be a dash.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/695#-:-KbRuZWBGXD_kD9eJA1B:b-7hp25x)*

justin808:
@alleycat-at-git should I close this one?

PR #698, opened 2017-01-27
Remove reference to heroku-buildpack-multi

xavriley:
`heroku buildpacks:add ...` will support any custom buildpacks - there's no use case for using `heroku-buildpack-multi` unless you are relying on a `.buildpacks` file to be present in the repo (which some older projects do).

In addition, https://github.com/ddollar/heroku-buildpack-multi is deprecated and is no longer being maintained. There is a maintained fork of the project at https://github.com/heroku/heroku-buildpack-multi but it isn't necessary to point people to this.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/698)


justin808:
Thanks @xavriley

PR #710, opened 2017-02-09
update i18n doc for .travis.yml

JasonYCHuang:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/710)

### DIFF
--- a/docs/basics/i18n.md
+++ b/docs/basics/i18n.md
@@ -57,7 +57,7 @@ You can refer to [react-webpack-rails-tutorial](https://github.com/shakacode/rea
   end
   ```
 
-5. In React, you need to initialize `react-intl`, and set parameters for it.
+4. In React, you need to initialize `react-intl`, and set parameters for it.
 
   ```js
   ...
@@ -88,3 +88,5 @@ You can refer to [react-webpack-rails-tutorial](https://github.com/shakacode/rea
     { formatMessage(defaultMessages.yourLocaleKeyInCamelCase) }
   )
   ```
+
+If the test are executed in Travis CI, make sure that `rake` is executed __before__ `npm run build` in `.travis.yml`. `npm run build` depends on `translations.js` & `default.js`, and they are generated or updated by `rake` or `rails server`.


justin808:
See comment.

---

Reviewed 1 of 1 files at r1.

---

*[docs/basics/i18n.md, line 92 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/710#-KcaTbXseGrgp_30yxxx:-KcaTbXvwsPNS5uN0_aC:b-mqvsjq) ([raw file](https://github.com/shakacode/react_on_rails/blob/39923b3b52c8dc8c53e8a0a944b7b52557c4fd7e/docs/basics/i18n.md#L92)):*
> ```Markdown
>   ```
> 
> If the test are executed in Travis CI, make sure that `rake` is executed __before__ `npm run build` in `.travis.yml`. `npm run build` depends on `translations.js` & `default.js`, and they are generated or updated by `rake` or `rails server`.
> ```

1. Not everybody uses Travis for CI
2. When tests run, they ensure that the the timestamp on the generated files are newer than the source files, or else the build script runs.
3. We should probably check the i18n files right before building the webpack files. This would cover the case for CI as well, so we would not need this.
4. In the case of the Rails server, it's fine to ensure that the i18n is built before the server starts, and the people just have to know to restart the server to pick up changes. Otherwise, we might want a specific rake task that can run`ReactOnRails::LocalesToJs.new`

The bottom line is that this small comment, right after the clearly numbered sections, is a bit confusing. We should "just make this work".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/710#-:-KcaUa49ntazdcgdRabW:btk743l)*

JasonYCHuang:
Agree with your comment.

The best way is to make sure i18n files are generated before building the webpack files.
I18n files are generated by rails, and I have no idea how to execute rake files from javascript files.
Could you give me some hint?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/710#-:-KccV9m5l9KDN_H4cD3l:b-rodevw)*

justin808:
@JasonYCHuang Here is where we configure the test helper to check webpack files are ready:
* https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/test_helper.rb#L34

The best way to ensure that the localization files are built before we build the webpack files is to have the npm scripts that build the webpack files first call some rake task. If you depend on the environment, that will probably run the code you put at server startup. However, I'm not sure that we always want the code at server startup. You'd know better.

Here's a rake task that just depends on server startup info:

```ruby
desc "Pick a random user as the winner"
task :winner => :environment do
  puts "Yes -- we did something. Describe it!"
end
```
http://stackoverflow.com/q/7044714/1009332

justin808:
@JasonYCHuang Should we close this PR? Per #717

JasonYCHuang:
@justin808 Yes, please close this PR.

PR #720, opened 2017-02-16
Update tutorial.md

gurbraj:
added:
gem install foreman             # download and install Foreman

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/720)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -33,6 +33,7 @@ rvm use 2.3.1 --default         # use it and make it default
 rvm list                        # check
 
 gem install rails               # download and install latest stable Rails
+gem install foreman             # download and install Foreman
 ```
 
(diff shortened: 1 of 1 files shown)


justin808:
Thanks @gurbraj!

gurbraj:
Thanks for a cool gem! Excited to try it out.

When I run the server with: 

foreman start -f Procfile.dev

I can not connect to it from the client (my browser).

However, it works when I do the usual: 
bin/rails s

I'm trying to integrate react onto an already made rails app.

Any ideas why? And can it become a problem down the line if I just use 'bin/rails s'?

Thanks!

justin808:
@gurbraj 

`export PORT=3000`

if you want to use localhost:3000

port 5000 is the default for foreman.

PR #734, opened 2017-02-23
Fix spec failing due to duplicate component

hrishimittal:
On running `rake`, the following spec fails:

```
  1) Pages/Index All in one page Simple Component Without Redux changes name in message according to input
     Failure/Error:
       within(dom_selector) do
         find("input").set new_text
         within("h3") do
           is_expected.to have_content new_text
         end
       end

     Capybara::Ambiguous:
       Ambiguous match, found 2 elements matching css "div#HelloWorld-react-component-4"
     Shared Example Group: "React Component" called from ./spec/features/integration_spec.rb:47
```

The reason is a duplicate component with the id `HelloWorld-react-component-4` in `pages/index.html.erb`.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/734)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Contributors: please follow the recommendations outlined at [keepachangelog.com]
 ### Fixed
 - Upgrade Rails to 4.2.8 to fix security vulnerabilities in 4.2.5. [#735](https://github.com/shakacode/react_on_rails/pull/735) by [hrishimittal](https://github.com/hrishimittal).
 
+### Fixed
+- Fix spec failing due to duplicate component. [#734](https://github.com/shakacode/react_on_rails/pull/734) by [hrishimittal](https://github.com/hrishimittal).
 
 ## [6.6.0] - 2017-02-18
 ### Added
--- a/spec/dummy/app/views/pages/index.html.erb
+++ b/spec/dummy/app/views/pages/index.html.erb
@@ -79,10 +79,10 @@ This page demonstrates a few things the other pages do not show:
 
 <h1>Simple Component Without Redux</h1>
 <pre>
-  <%%= react_component("HelloWorld", props: @app_props_hello, prerender: false, trace: true, id: "HelloWorld-react-component-4") %>
+  <%%= react_component("HelloWorld", props: @app_props_hello, prerender: false, trace: true, id: "HelloWorld-react-component-5") %>
   <%%= react_component("HelloES5", props: @app_props_hello, prerender: false, trace: true, id: "HelloES5-react-component-5") %>
 </pre>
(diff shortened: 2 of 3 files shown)


hrishimittal:
Travis failed because of a timeout. Can someone please re-run it? - https://travis-ci.org/shakacode/react_on_rails/jobs/204543891#L3795

justin808:
This change does not look valid. Can we close this PR?

---

Reviewed 2 of 2 files at r1.

---

*[spec/dummy/app/views/pages/index.html.erb, line 77 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/734#-KdgvlXQ0AfZ32oZCJQS:-KdgvlXQ0AfZ32oZCJQT:bukm6bd) ([raw file](https://github.com/shakacode/react_on_rails/blob/3106b4b8f8fb643d0dc02cdca22e3d53f2cf2cce/spec/dummy/app/views/pages/index.html.erb#L77)):*
> ```erb
>   <%%= react_component("HelloES5", props: @app_props_hello, prerender: false, trace: true, id: "HelloES5-react-component-5") %>
> </pre>
> <%= react_component("RenderedHtml", prerender: true, trace: true, id: "HelloWorld-react-component-4") %>
> ```

you essentially took out the running of this test of generating some html on the server.

the code directly above is in a pre block and it's not rendered per the `<%%`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/734#-:-Kdgw9OSrUdf4zhYyJON:bba3qtz)*

wanyamaman:
I was also looking at this.

Thoughts: I suspect the solution is to increase the `id` number and write corresponding tests for it. There might also be a missing component since the `pre` tag shows two components but the actual code has one.

hrishimittal:
@justin808 I understand the pre block code is not rendered. But look at this line - 

https://github.com/shakacode/react_on_rails/pull/734/files#diff-13675373e6eaa2b6e91995c3ec8974f7R84

The same id is used again, which is why the spec was failing.

Removing the RenderedHtml line didn't break any test, so that suggests maybe a test is missing?

As @wanyamaman suggested, we probably need another id and corresponding tests.

justin808:
@hrishimittal The ids shouldn't conflict. Can you put back the removed line and just change around the ids?

justin808:
> Removing the RenderedHtml line didn't break any test, so that suggests maybe a test is missing?
and yes! looks that way!

hrishimittal:
@justin808 I've restored the deleted line and changed ids to fix the spec. If you can give me a LGTM, I'll squash and rebase.

justin808:
@hrishimittal Please rebase on top of master and push force. Thanks!

PR #761, opened 2017-03-14
Add missing listen gem

wanyamaman:
The `listen` gem is needed for rails 5 apps when using the file watcher [here.](https://github.com/shakacode/react_on_rails/blob/f42e8f47af7a6b7e0f7b55d4563b236a86aba4a3/spec/dummy/config/environments/development.rb#L53)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/761)


justin808:
Big thanks @wanyamaman!

PR #764, opened 2017-03-19
Fix turbolinks mount event

szyablitsky:
This change fixes "The node you're attempting to unmount was rendered by another copy of React." error when navigating to cached page using Turbolinks.

More info can be found at https://github.com/renchap/webpacker-react/pull/14#issuecomment-282500093 and https://sevos.io/2017/02/27/turbolinks-lifecycle-explained.html

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/764)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Contributors: please follow the recommendations outlined at [keepachangelog.com]
 
 ## [Unreleased]
 *Please add entries here for your pull requests.*
+
+### Fixed
+- Fixed error "The node you're attempting to unmount was rendered by another copy of React." [#706](https://github.com/shakacode/react_on_rails/issues/706) when navigating to cached page using Turbolinks  [#763](https://github.com/shakacode/react_on_rails/pull/763) by [szyablitsky](https://github.com/szyablitsky).
+
 ## [6.8.0] - 2017-03-06
 ## Added
 - Converted to Webpack v2 for generators, tests, and all example code. [#742](https://github.com/shakacode/react_on_rails/pull/742) by [justin808](https://github.com/justin808).
@@ -29,7 +33,7 @@ Commenting out this line addresses the issue:
 
 ```
 config.i18n_dir = Rails.root.join("client", "app", "libs", "i18n")
-``` 
+```
 
 ### Added
 - Allow using rake task to generate javascript locale files. The test helper automatically creates the localization files when needed. [#717](https://github.com/shakacode/react_on_rails/pull/717) by [JasonYCHuang](https://github.com/JasonYCHuang).
--- a/docs/additional-reading/turbolinks.md
+++ b/docs/additional-reading/turbolinks.md
@@ -94,7 +94,7 @@ TURBO: reactOnRailsPageLoaded
 
 Turbolinks 5:
 ```
-TURBO: WITH TURBOLINKS 5: document turbolinks:before-render and turbolinks:load handlers installed. (program)
+TURBO: WITH TURBOLINKS 5: document turbolinks:before-render and turbolinks:render handlers installed. (program)
 TURBO: reactOnRailsPageLoaded
 ```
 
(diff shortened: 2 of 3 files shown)


justin808:
@szyablitsky Thanks! Any details on how you resolved this issue?

---

Reviewed 3 of 3 files at r1.

---

*[node_package/src/clientStartup.js, line 201 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-Kf_-EWQzJZ-pwOen_at:-Kf_-EWQzJZ-pwOen_au:b-61ppja) ([raw file](https://github.com/shakacode/react_on_rails/blob/8ea86471f95127d144e397e2ac58087e90e41920/node_package/src/clientStartup.js#L201)):*
> ```JavaScript
>       );
>     }
>     reactOnRailsPageLoaded();
> ```

@szyablitsky Are we sure that the event handlers on `turbolinks:render` or `page:change` won't  causes a double render?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-:-Kf_-cTr_NECD6S9Arbi:b-htpcva)*

szyablitsky:
I included a link to the article where this issue explained in full details.
In short, when user visits cached page, Turbolinks first render cached version, then (when page is loaded) render loaded version, so we have two `turbolinks:before-render` events and one `turbolinks:load` event. Second `turbolinks:before-render` tries to unmount components but they are already unmounted. So we use `turbolinks:render` which is called  after each `turbolinks:before-render`. And we should mount components on page load (`DOMContentLoaded`), because Turbolinks do not fire `turbolinks:render` on first page render.

---

---

*[node_package/src/clientStartup.js, line 201 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-Kf_-EWQzJZ-pwOen_at:-Kf_8SzhMmOUNkU6FZ6P:b-nm7nw6) ([raw file](https://github.com/shakacode/react_on_rails/blob/8ea86471f95127d144e397e2ac58087e90e41920/node_package/src/clientStartup.js#L201)):*

@szyablitsky Are we sure that the event handlers on `turbolinks:render` or `page:change` won't  causes a double render?
</blockquote></details>

My bad. Forgot about Turbolinks 2. Will fix now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-:-Kf_9-GKRYziPMOWaRMO:bkmtrvn)*

justin808:
Looks good. 100% sure we can ship, @szyablitsky? 

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[node_package/src/clientStartup.js, line 201 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-Kf_-EWQzJZ-pwOen_at:-Kf_Kpatgpn9aimnOHSU:b-zfwjs2) ([raw file](https://github.com/shakacode/react_on_rails/blob/8ea86471f95127d144e397e2ac58087e90e41920/node_package/src/clientStartup.js#L201)):*
<details><summary><i>Previously, szyablitsky (Sergey Zyablitsky) wrote…</i></summary><blockquote>

My bad. Forgot about Turbolinks 2. Will fix now.
</blockquote></details>

Let me know when you're sure we're good to go and I'll release this!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/764#-:-Kf_L2XHYwFECBvwWO66:b59jutt)*

szyablitsky:
We need to rebuild on CI because it's timed out on bundle install.

justin808:
As soon as I can get this to pass CI, I'll release.

justin808:
error:
```
Installing rails-html-sanitizer 1.0.3
Installing capybara 2.12.1
Gem::RemoteFetcher::UnknownHostError: timed out
(https://rubygems.org/gems/pry-rescue-1.4.5.gem)
An error occurred while installing pry-rescue (1.4.5), and Bundler
cannot continue.
Make sure that `gem install pry-rescue -v '1.4.5'` succeeds before bundling.
The command "bundle install" failed 3 times.
The command "travis_retry bundle install" failed and exited with 5 during .
Your build has been stopped.
```

justin808:
I'll restart

justin808:
Thanks @szyablitsky

PR #765, opened 2017-03-19
Add image handling example

justin808:
Demonstrates using both background images in CSS and img tags. Path of images can be relative or absolute.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/765)


justin808:
## 2 problems:

1. Not passing CI
2. starting the app by `cd spec/dummy && foreman start -f Procfile.hot` gives missing images per image below:

justin808:
I restarted CI. That might be an intermittent failure.

Cheremukhin23:
@justin808 
1)It seems like the first problem is really intermittent and travis should be restarted one more time.
2)Error encountered when webpack configured wrong and might be connected with dev server inner configuration
3) The main reason of images lack is disability of css at all (you can notice that labels also have no background images)
I've added style-loader to webpack config and it works now
image

justin808:
Finished in #768.

PR #778, opened 2017-03-29
Fix issue with release script

isolo:
Instead of gem-release's **release** command, I use **build** and **release:rubygems_push** from bundler/gem_tasks. 
Default release command of bundler consists of [the four following commands](https://github.com/bundler/bundler/blob/master/lib/bundler/gem_helper.rb#L55): 
- **build**
- **release:guard_clean**
- **release:source_control_push**
- **release:rubygem_push**

We already have implementation of "**release:guard_clean**" (checks if there is uncommitted changes) and "**release:source_control_push**" (pushes tags to vcs). I thought we have more flexibility with the current implementation, so I just replaced '**release**' command of gem-release.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/778)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project's source code will be documented in this fil
 Contributors: please follow the recommendations outlined at [keepachangelog.com](http://keepachangelog.com/). Please use the existing headings and styling as a guide, and add a link for the version diff at the bottom of the file. Also, please update the `Unreleased` link to compare to the latest release version.
 
 ## [Unreleased]
-*Please add entries here for your pull requests.*
+## Fixed
+- Fixed error in the release script [#767](https://github.com/shakacode/react_on_rails/issues/767) by [isolo](https://github.com/isolo).
 
 ## [6.8.2] - 2017-03-24
 ## Fixed
--- a/Rakefile
+++ b/Rakefile
@@ -12,3 +12,6 @@ task default: tasks
 
 desc "All actions but no examples, good for local developer run."
 task all_but_examples: ["run_rspec:all_but_examples", "lint"]
+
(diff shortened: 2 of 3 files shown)


justin808:
Hey! this looks great!

I had made this comment in the other issue:

I found the issue. @isolo Let me know if you're quite sure we should go with your fix as is.

One thing to consider is maintenance. What happens as the `gem-release` evolves? And React on Rails. What's the most bulletproof?

Thanks again!

Justin


----------

```
find . -name "*.gemspec"
./react_on_rails.gemspec
./spec/dummy/client/node_modules/react-on-rails/react_on_rails.gemspec
```

That comes from: https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/client/package.json#L36

```
    "react-on-rails": "file:../../..",
```


@isolo Any chance that you can take a look at the rake task for the release and delete that extra gemspec? Or maybe delete all gemspecs except the root one? I think the properly crafted find/rm command right here could do the trick.

https://github.com/shakacode/react_on_rails/blob/master/rakelib/release.rake#L29

justin808:
Looks super clean. I'll try out later today. :lgtm: 

---

Reviewed 1 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/778#-:-KgRAMrZ6NdWIVFiyj-R:b-uguctz)*

justin808:
We should remove the gen release dependency if not needed.

justin808:
And add entry to the changelog.

isolo:
I've added an entry to the changelog. As for gem-release, right now we have a dependency on it because we decided to use 'gem release' and 'gem bump'.

justin808:
LGTM! Awesome!

PR #794, opened 2017-04-06
add additional steps that will allow yarn the permissions to write to…

josephecombs:
… /home/webapp during asset compilation

Asset compilation will still fail after adding this ebextension in the way mentioned in this blog post:

http://tech.nickserra.com/2016/11/15/npm-install-error-eacces-permission-denied-mkdir-homewebapp/

the additional .ebextensions commands create a directory with the correct permissions to allow yarn to write to it later.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/794)


justin808:
Thanks @josephecombs !

PR #811, opened 2017-04-19
Daniel/webpacker integration

kaizencodes:
- Updated generators, templates and tests
- Updated dummy app (wip: css not loading)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/811)


justin808:
Great start!!!!

---

Reviewed 26 of 26 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js, line 15 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4X4q7zdPzVOuFmvtO:-Ki4X4q7zdPzVOuFmvtP:b-x7a4df) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js#L15)):*
> ```JavaScript
>   paths,
>   publicPath,
> };
> ```

@kaizencodes Should we put this in the node library?

so we don't have to include this with the generator?

I think so.

If this is in the library, we can have the exports come from a function call, so that if the configuration place is different, it can be set, or if the webpack config is in a different place.

Essentially, have this be the default config path, but allow overriding it.

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml, line 19 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4XbkfltH6H3M8vYyb:-Ki4XbkfltH6H3M8vYyc:bbgb8vx) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml#L19)):*
> ```YAML
> production:
>   <<: *default
>   assets: webpack/production
> ```

this is **AWESOME**

cc: @robwise, @alexfedoseev, @samnang  

---

*[spec/dummy/client/webpack.client.base.config.js, line 20 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4Y1NMLF6yyoBoZ2ij:-Ki4Y1NMLF6yyoBoZ2ik:b-cms1mn) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/client/webpack.client.base.config.js#L20)):*
> ```JavaScript
> } catch (ex) {
>   console.error(ex);
>   console.log('Make sure the client build (client.base.build or client.rails.build) creates a manifest in:', manifestPath);
> ```

this line is maybe too wide, definitely we want to try to fi lines under 100 chars wide

---

*[spec/dummy/client/webpack.client.base.config.js, line 39 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4XuHipQ7qgjaYCrmz:-Ki4XuHipQ7qgjaYCrn-:b-ey9pzd) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/client/webpack.client.base.config.js#L39)):*
> ```JavaScript
>     //   paths.source,
>     //   paths.node_modules,
>     // ],
> ```

remove comment?

---

*[spec/dummy/client/webpackConfigLoader.js, line 22 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4YMRp4lwBcY1gYymV:-Ki4YMRqSJSxmqKZq_5X:b-f9n0g1) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/client/webpackConfigLoader.js#L22)):*
> ```JavaScript
>   paths,
>   publicPath,
> };
> ```

this could come from the node file, just like the generator file

so that would DRY up things

---

*[spec/dummy/config/initializers/react_on_rails.rb, line 100 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4YT_tz3AbqQmN3YT7:-Ki4YT_tz3AbqQmN3YT8:b-hjz0vs) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/config/initializers/react_on_rails.rb#L100)):*
> ```Ruby
>   # For any asset matching this regex, a file is copied to the correct path to have a digest.
>   # To disable creating digested assets, set this parameter to nil.
>   config.symlink_non_digested_assets_regex = nil
> ```

that needs to be the default for new projects --> did that change for the generator?

In fact, we should consider removing this from generated file, and the default is nil.

I think for here, it's OK to specify it. However, that will not confirm that the default is nil, possibly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-Ki4YqWYo5kYDlw7L0tr:bs5k9hm)*

robwise:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml, line 2 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki5jI1pqTd9VN47Xk6h:-Ki5jI1pqTd9VN47Xk6i:byysxjo) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml#L2)):*
> ```YAML
> default: &default
>   config: client/webpack
> ```

I still don't get how I set it to build two configs (server-rendering and client). I've said this over and over and no one ever responds. Why does webpacker even need to know about my webpack config?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml, line 2 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki5jI1pqTd9VN47Xk6h:-Ki5kfv5BhUyid0GdaNP:b-lmlpjz) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml#L2)):*

I still don't get how I set it to build two configs (server-rendering and client). I've said this over and over and no one ever responds. Why does webpacker even need to know about my webpack config?
</blockquote></details>

@robwise It needs to know where the manifest file is, to include the assets. Why wouldn't it work for 2 configs? Does the output differ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

robwise:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml, line 2 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki5jI1pqTd9VN47Xk6h:-Ki5l_IjUtI2vOX8dRVe:bbcp8xo) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml#L2)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

@robwise It needs to know where the manifest file is, to include the assets. Why wouldn't it work for 2 configs? Does the output differ?
</blockquote></details>

@kaizencodes It wouldn't work for two configs because it only takes a single string as an argument and not an array, so how do I put two configs there? We have a server bundle config and a client bundles config for server rendering. We also have a DLL config for development. Which one of the three do I put here? What happens to the other two?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml, line 2 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki5jI1pqTd9VN47Xk6h:-Ki5oUICu7JnYXQkZu7e:b-tnuosg) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/config/webpack/paths.yml#L2)):*

@kaizencodes It wouldn't work for two configs because it only takes a single string as an argument and not an array, so how do I put two configs there? We have a server bundle config and a client bundles config for server rendering. We also have a DLL config for development. Which one of the three do I put here? What happens to the other two?
</blockquote></details>

@robwise Okay, now I get where you are getting at. First of this was meant to show a dir path where all the configs are located. But we don't use this 'config' key anymore (Should have deleted it already). The procfiles load the appropriate configs, each loads this shared file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
@kaizencodes Let's delete any unnecessary configs, and also make it clear in the comments of the config files that webpacker_lite **differs** from webpacker in that you have to provide your own config.

kaizencodes:
Shouldn't it be better to place this info in the webpacker_lite readme?

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js, line 15 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4X4q7zdPzVOuFmvtO:-Ki90fDIJSi2xNlpybDA:bi6r5nk) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js#L15)):*

@kaizencodes Should we put this in the node library?

so we don't have to include this with the generator?

I think so.

If this is in the library, we can have the exports come from a function call, so that if the configuration place is different, it can be set, or if the webpack config is in a different place.

Essentially, have this be the default config path, but allow overriding it.
</blockquote></details>

I'm not sure how this would fit in the node package.

---

*[spec/dummy/config/initializers/react_on_rails.rb, line 100 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4YT_tz3AbqQmN3YT7:-Ki8zr9-Eeh3AaAN1iOc:bp4hnvb) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/config/initializers/react_on_rails.rb#L100)):*

that needs to be the default for new projects --> did that change for the generator?

In fact, we should consider removing this from generated file, and the default is nil.

I think for here, it's OK to specify it. However, that will not confirm that the default is nil, possibly.
</blockquote></details>

nil is the default. However, This is from the template:
```
config.symlink_non_digested_assets_regex = /\.(png|jpg|jpeg|gif|tiff|woff|ttf|eot|svg|map)/
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-Ki8xN9PSxB9CKLQBqy6:buj28vf)*

justin808:
Couple questions

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[spec/dummy/client/webpack.client.base.config.js, line 19 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiEYgfo09UNGBlJV-g-:-KiEYgfo09UNGBlJV-g0:b-y759w0) ([raw file](https://github.com/shakacode/react_on_rails/blob/e41d0918da8e8ea0341f543f9de22c11cd1bd54c/spec/dummy/client/webpack.client.base.config.js#L19)):*
> ```JavaScript
>   sharedManifest = require(manifestPath);
> } catch (ex) {
>   console.log('No manifest found. It will be created at:', manifestPath);
> ```

I'm not clear on this code or the error message. Please explain. What is the sharedManifest variable for? Seems to be some sort of caching thing.

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 22 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiEZdBb-7g6nK6B_0UQ:-KiEZdBb-7g6nK6B_0UR:bo6pvpm) ([raw file](https://github.com/shakacode/react_on_rails/blob/e41d0918da8e8ea0341f543f9de22c11cd1bd54c/spec/dummy/client/webpack.server.rails.build.config.js#L22)):*
> ```JavaScript
> } catch (ex) {
>   console.error(ex);
>   console.log('No manifest found. It will be created at:', manifestPath);
> ```

is this error ever expected, as in the first time this is run?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-KiEZhhh-56xqtVSoN1Y:bdrz0gi)*

kaizencodes:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[spec/dummy/client/webpack.client.base.config.js, line 19 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiEYgfo09UNGBlJV-g-:-KiE_Sg2CzPxtM1LnS5y:b-k10oqk) ([raw file](https://github.com/shakacode/react_on_rails/blob/e41d0918da8e8ea0341f543f9de22c11cd1bd54c/spec/dummy/client/webpack.client.base.config.js#L19)):*

I'm not clear on this code or the error message. Please explain. What is the sharedManifest variable for? Seems to be some sort of caching thing.
</blockquote></details>

Yes, it is caching. When we use 2 configs like client and server, each will write to the manifest. However, if we do not include the hash created by the first, the second will overwrite it. 

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 22 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiEZdBb-7g6nK6B_0UQ:-KiEa-BhAgm59hCmj-kw:b-l4e3gg) ([raw file](https://github.com/shakacode/react_on_rails/blob/e41d0918da8e8ea0341f543f9de22c11cd1bd54c/spec/dummy/client/webpack.server.rails.build.config.js#L22)):*

is this error ever expected, as in the first time this is run?
</blockquote></details>

yes, it occurs with the first config always.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js, line 15 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4X4q7zdPzVOuFmvtO:-KiEanaSa2FyqS1p-7G7:blihmpt) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js#L15)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

I'm not sure how this would fit in the node package.
</blockquote></details>

This could be a non-default import in the node package.

import webpackConfigLoader from 'react-on-rails/webpackConfigLoader'

---

*[spec/dummy/client/webpack.client.base.config.js, line 19 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiEYgfo09UNGBlJV-g-:-KiEb0SuSG7zQXGi4hXL:blqm6pf) ([raw file](https://github.com/shakacode/react_on_rails/blob/e41d0918da8e8ea0341f543f9de22c11cd1bd54c/spec/dummy/client/webpack.client.base.config.js#L19)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

Yes, it is caching. When we use 2 configs like client and server, each will write to the manifest. However, if we do not include the hash created by the first, the second will overwrite it. 
</blockquote></details>

Maybe the message should only be printed if in a debugging mode? Maybe the message should be clear that this prints the first time the config is run?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
some comments 

---

Reviewed 12 of 12 files at r3.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpack.config.js, line 7 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJhpEDRvkyNeRfARgq:-KiJhpEDRvkyNeRfARgr:b-tx0f7h) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/lib/generators/react_on_rails/templates/base/base/client/webpack.config.js#L7)):*
> ```JavaScript
> const webpack = require('webpack');
> const { resolve } = require('path');
> const ManifestPlugin = require('webpack-manifest-plugin'); // eslint-disable-line import/no-unresolved
> ```

This seems odd to need the eslint disable here. Any idea on why?

---

*[lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js, line 15 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Ki4X4q7zdPzVOuFmvtO:-KiJlFMTFdgwEA_uAk7O:b-i1csp9) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/lib/generators/react_on_rails/templates/base/base/client/webpackConfigLoader.js#L15)):*

This could be a non-default import in the node package.

import webpackConfigLoader from 'react-on-rails/webpackConfigLoader'
</blockquote></details>

@robwise what do you think about putting this in the react-on-rails package?

The one issue is that we'd have to pass in a param of the location of the config/webpack/paths.yml file.

---

*[spec/dummy/app/views/layouts/application.html.erb, line 24 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJjG6tNxCSaBYI3_AW:-KiJjG6tNxCSaBYI3_AX:bgikuq4) ([raw file](https://github.com/shakacode/react_on_rails/blob/85d6f6fc6f6011513efdca057d98f314cf68e34c/spec/dummy/app/views/layouts/application.html.erb#L24)):*
> ```erb
>   
>   <%= javascript_pack_tag('vendor',
>                           'data-turbolinks-track': true) %>
> ```

why are we removing the turbo tag?

---

*[spec/dummy/app/views/layouts/application.html.erb, line 20 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJizF1DtbCyaHmuNhT:-KiJizF1DtbCyaHmuNhU:bauc1aq) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/app/views/layouts/application.html.erb#L20)):*
> ```erb
>   
> 
>   
> ```

should this comment be here

---

*[spec/dummy/client/webpack.client.base.config.js, line 37 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkPqFuWPmfS2S-zl6:-KiJkPqGUBdG0sU-xcMG:bura290) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/client/webpack.client.base.config.js#L37)):*
> ```JavaScript
>     alias: {
>       images: join(process.cwd(), 'app', 'assets', 'images'),
>       'react-on-rails': resolve(__dirname, '..', '..', '..'),
> ```

this seems odd b/c we also have react-on-rails listed in the package.json

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 39 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkXF1JaM-oZVdVYGK:-KiJkXF1JaM-oZVdVYGL:bura290) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/client/webpack.server.rails.build.config.js#L39)):*
> ```JavaScript
>     alias: {
>       images: join(process.cwd(), 'app', 'assets', 'images'),
>       'react-on-rails': resolve(__dirname, '..', '..', '..'),
> ```

this seems odd b/c we also have react-on-rails listed in the package.json

---

*[spec/dummy/config/initializers/assets.rb, line 14 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkgv92Ap594yHO_Il:-KiJkgv92Ap594yHO_Im:bxr066j) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/initializers/assets.rb#L14)):*
> ```Ruby
> 
> # Add folder with webpack generated assets to assets.paths
> Rails.application.config.assets.paths << Rails.root.join("public", "webpack", Rails.env)
> ```

very nice!

this needs to be well documented!

is this in the generator as well?

it should be!

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiJkoexQKUTvGPay4uZ:b-n5z31h) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*
> ```YAML
> development:
>   <<: *default
>   enabled: false
> ```

so you're turning off hot reloading?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-KiJlt4T6J4cwH9hWN0S:b-b9q1iz)*

kaizencodes:
Review status: 24 of 29 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpack.config.js, line 7 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJhpEDRvkyNeRfARgq:-KiLaIV2uTM7ProAP49S:bp5ds81) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/lib/generators/react_on_rails/templates/base/base/client/webpack.config.js#L7)):*

This seems odd to need the eslint disable here. Any idea on why?
</blockquote></details>

eslint couldn't resolve it. Which is understandable given we don't have the package in the generator.

---

*[spec/dummy/client/webpack.client.base.config.js, line 37 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkPqFuWPmfS2S-zl6:-KiLYXSz_96F2NOnlMd0:bmpwf5c) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/client/webpack.client.base.config.js#L37)):*

this seems odd b/c we also have react-on-rails listed in the package.json
</blockquote></details>

yes, it is needed when we require it locally.

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 39 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkXF1JaM-oZVdVYGK:-KiLZ7bj_NjD_tpw1NG8:bmpwf5c) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/client/webpack.server.rails.build.config.js#L39)):*

this seems odd b/c we also have react-on-rails listed in the package.json
</blockquote></details>

yes, it is needed when we require it locally.

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiL_ElKB0G14EbIFjHo:b-u752w4) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*

so you're turning off hot reloading?
</blockquote></details>

worked with static last time, hot loading should be the default?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[lib/generators/react_on_rails/base_generator.rb, line 62 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAMnjcJvSrEPqIkA_:-KiPAMnjcJvSrEPqIkAa:b-9xss4t) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/base_generator.rb#L62)):*
> ```Ruby
>            Procfile.dev
>            package.json
>            client/package.json).each { |file| template("#{base_path}#{file}.tt", file, force: true) }
> ```

I think this is going to blow away any existing application.html.erb.

The generator will be used on existing projects.

How about we make a new layout and set that in the controller for hello_world?

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiPAbMeHrpaGs8TPNmB:bkg21uz) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*
> ```Ruby
>           'extensions: [\'.js\', \'.jsx\'],
>             alias: {
>               \'react-on-rails\': resolve(__dirname, \'..\', \'..\', \'..\', \'..\'),
> ```

I don't get why we need an alias for this.

The generators will generate code that depend on the release of react-on-rails corresponding to the gem version.

---

*[lib/generators/react_on_rails/templates/base/base/app/views/layouts/application.html.erb.tt, line 10 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAsukRiQEJigYtk2f:-KiPAsukRiQEJigYtk2g:b-jmdthl) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/templates/base/base/app/views/layouts/application.html.erb.tt#L10)):*
> ```
>     <%%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
>     <%%= javascript_pack_tag 'main' %>
>   </head>
> ```

This file can be renamed so as not to conflict with any existing application.html.erb.

---

*[lib/generators/react_on_rails/templates/base/base/client/webpack.config.js, line 7 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJhpEDRvkyNeRfARgq:-KiPAz_WhGQe7CdaXvoz:bxqhle7) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/lib/generators/react_on_rails/templates/base/base/client/webpack.config.js#L7)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

eslint couldn't resolve it. Which is understandable given we don't have the package in the generator.
</blockquote></details>

hmmm....we don't want the generated code to have that unnecessary disable. I think we can add the package globally /package.json

---

*[spec/dummy/client/webpack.client.base.config.js, line 37 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkPqFuWPmfS2S-zl6:-KiPBYMIDMeNx7KtAFf2:bipf33k) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/client/webpack.client.base.config.js#L37)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

yes, it is needed when we require it locally.
</blockquote></details>

@kaizencodes But we didn't change this for 7.0, so we should not be adding this.

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiPBgEUaOePQCmp08Uq:bygj0dh) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

worked with static last time, hot loading should be the default?
</blockquote></details>

I'm really not sure on this one.

Can this setting be overriden with an env value? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiPD6qpTTr6FIjDR9aY:bn8iyve) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*

I don't get why we need an alias for this.

The generators will generate code that depend on the release of react-on-rails corresponding to the gem version.
</blockquote></details>

I don't get it either, yet without it react-on-rails fails to resolve.

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiQiyq0q7YOD8vXHh-j:b9thmlm) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*

I'm really not sure on this one.

Can this setting be overriden with an env value? 
</blockquote></details>

We could. Should we?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiQpLZ5AsoD59ht96E-:by6zkk3) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

I don't get it either, yet without it react-on-rails fails to resolve.
</blockquote></details>

this is what I get on master:
```
ERROR in ./app/bundles/HelloWorld/startup/registration.jsx
Module not found: Error: Can't resolve 'react-on-rails' in '/home/daniel/rails/react_on_rails/gen-examples/examples/redux-server-rendering/client/app/bundles/HelloWorld/startup'
 @ ./app/bundles/HelloWorld/startup/registration.jsx 3:20-45
 @ multi es5-shim/es5-shim es5-shim/es5-sham babel-polyfill ./app/bundles/HelloWorld/startup/registration
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
Review status: 27 of 32 files reviewed at latest revision, 17 unresolved discussions.

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiQwTZCmQSCpdayrRwm:bat8vf6) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

We could. Should we?
</blockquote></details>

We should, this caused the travis error. I've set up the config loader to overwrite this value when REACT_ON_RAILS_ENV is HOT. Now the question comes: Do we need this? Is there a scenario where we need the dev server but not hot reload?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

kaizencodes:
rebased onto master

---

Review status: 27 of 32 files reviewed at latest revision, 17 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-KiQxm-JW1RWugoqUBub:bg6itf)*

justin808:
more comments

---

Reviewed 6 of 6 files at r5.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[.eslintignore, line 15 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR3H_ZkNW5tQMYgqjw:-KiR3H__AJhZUYMmxKyc:b-brvmz) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/.eslintignore#L15)):*
> ```
> **/coverage/**
> **/cable.js
> gen-examples/**/webpack.config.js
> ```

I don't think we should do this. It's good to lint these.

Did you try installing at the top level, the missing dependency?

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiR44g72orCMayR18CO:buhbec2) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

this is what I get on master:
```
ERROR in ./app/bundles/HelloWorld/startup/registration.jsx
Module not found: Error: Can't resolve 'react-on-rails' in '/home/daniel/rails/react_on_rails/gen-examples/examples/redux-server-rendering/client/app/bundles/HelloWorld/startup'
 @ ./app/bundles/HelloWorld/startup/registration.jsx 3:20-45
 @ multi es5-shim/es5-shim es5-shim/es5-sham babel-polyfill ./app/bundles/HelloWorld/startup/registration
```
</blockquote></details>

this is odd

We definitely can't put this in.

For one, the generator won't work outside of within the REact on Rails source!

---

*[lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb, line 2 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR4HsvXymZQ5IFPZSa:-KiR4HsvXymZQ5IFPZSb:b-uooqxh) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb#L2)):*
> ```Ruby
> class HelloWorldController < ApplicationController
>   layout "hello_world"
> ```

I think we should rename all the "hello_world" to "hello_react_on_rails"

---

*[lib/generators/react_on_rails/templates/base/base/client/webpack.config.js, line 7 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJhpEDRvkyNeRfARgq:-KiR4VJS4cY3VvgAO_6_:b44st8) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/lib/generators/react_on_rails/templates/base/base/client/webpack.config.js#L7)):*

hmmm....we don't want the generated code to have that unnecessary disable. I think we can add the package globally /package.json
</blockquote></details>

hey, package should be in the generated package.json. Maybe that's the issue. And not a dev, but a regular dependency.

---

*[spec/dummy/client/webpackConfigLoader.js, line 10 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR4reOsVpCZ2foCWR3:-KiR4reOsVpCZ2foCWR4:b5a2ivx) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/spec/dummy/client/webpackConfigLoader.js#L10)):*
> ```JavaScript
> 
> const devServerConfig = join(configPath, 'development.server.yml');
> const devServer = safeLoad(readFileSync(devServerConfig, 'utf8')).development;
> ```

let's put a comment in the yml file about this env value

and I still want this whole thing in the node package

---

*[spec/dummy/client/webpackConfigLoader.js, line 14 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR4p_ek_U5sqb5HDE0:-KiR4p_ek_U5sqb5HDE1:bnq95dc) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/spec/dummy/client/webpackConfigLoader.js#L14)):*
> ```JavaScript
> if (env.REACT_ON_RAILS_ENV == "HOT") {
>   devServer.enabled = true
> }
> ```

excellent!

---

*[spec/dummy/config/webpack/development.server.yml, line 8 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiJkoexQKUTvGPay4uY:-KiR51ddUNy7axxTO4dO:b-d9f3yi) ([raw file](https://github.com/shakacode/react_on_rails/blob/ed05bafe2cb004be4772253bfad8053303f61fe0/spec/dummy/config/webpack/development.server.yml#L8)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

We should, this caused the travis error. I've set up the config loader to overwrite this value when REACT_ON_RAILS_ENV is HOT. Now the question comes: Do we need this? Is there a scenario where we need the dev server but not hot reload?
</blockquote></details>

hot == dev server

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-KiR55pR4R9AnTquw5rF:b-uh9cw6)*

kaizencodes:
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb, line 2 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR4HsvXymZQ5IFPZSa:-KiW0rE3NlHWUEH3-3ZX:by4mspp) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb#L2)):*

I think we should rename all the "hello_world" to "hello_react_on_rails"
</blockquote></details>

Should we? it's all over the place

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
getting there! see comments

---

Reviewed 19 of 19 files at r6.
Review status: all files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[package.json, line 44 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYergks0kL1ROckKjE:-KiYergks0kL1ROckKjF:b-jb2q4g) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/package.json#L44)):*
> ```JSON
>   },
>   "browser": {
>     "fs": false
> ```

what is this part for?

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiYez-4h64WBJq5KPDr:b-rpi7x1) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*

this is odd

We definitely can't put this in.

For one, the generator won't work outside of within the REact on Rails source!
</blockquote></details>

did you figure this out? maybe I can help today.

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 55 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgyBN1DQToIFTwwLd:-KiYgyBN1DQToIFTwwLe:bxpf1o4) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/lib/generators/react_on_rails/dev_tests_generator.rb#L55)):*
> ```Ruby
>       #                  old_contents.gsub(sentinel, FALLBACK_OPTION_FOR_NODE_MODULES)
>       #   File.open(config, "w+") { |f| f.puts new_contents }
>       # end
> ```

remove comments?

---

*[lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb, line 2 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiR4HsvXymZQ5IFPZSa:-KiYf0qqMKpiTFLgGAkY:byjlvt7) ([raw file](https://github.com/shakacode/react_on_rails/blob/9c3926f1c365e73ae0985003528cb898b37b5b3f/lib/generators/react_on_rails/templates/base/base/app/controllers/hello_world_controller.rb#L2)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

Should we? it's all over the place
</blockquote></details>

actually, this could be a separate PR -- keep this the same for now

---

*[lib/generators/react_on_rails/templates/base/base/client/webpack.config.js, line 8 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYf93mqWxdqzI71n2X:-KiYf93mqWxdqzI71n2Y:b-3smx0v) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/lib/generators/react_on_rails/templates/base/base/client/webpack.config.js#L8)):*
> ```JavaScript
> const { resolve } = require('path');
> const ManifestPlugin = require('webpack-manifest-plugin');
> const webpackConfigLoader = require('react-on-rails').default.webpackConfigLoader; // eslint-disable-line import/no-unresolved
> ```

any way we can get this to pass without the eslint issue? I think maybe we can list react-on-rails in the top level package.json, and use a relative file ref


and we need to change this

instead the require needs to be like

`const configLoader = require('react-on-rails/configLoader');`

rather than the default

---

*[lib/generators/react_on_rails/templates/base/base/config/webpack/development.server.yml, line 8 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYfOrPlCVcg-yzQH11:-KiYfOrPlCVcg-yzQH12:bhnrogv) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/lib/generators/react_on_rails/templates/base/base/config/webpack/development.server.yml#L8)):*
> ```YAML
> development:
>   <<: *default
>   enabled: false
> ```

we should put a comment here to make hot reloading the default,  you can change this line to true

something like that, that is clear and short

---

*[node_package/src/ReactOnRails.js, line 11 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYfxUdu2m-FYp1FvMn:-KiYfxUdu2m-FYp1FvMo:b-t2pmiw) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/node_package/src/ReactOnRails.js#L11)):*
> ```JavaScript
> import createReactElement from './createReactElement';
> import Authenticity from './Authenticity';
> import webpackConfigLoader from './webpackConfigLoader';
> ```

we need to remove this

instead the require needs to be like

`require('react-on-rails/configLoader')`

rather than the default

---

*[node_package/src/ReactOnRails.js, line 227 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYfuJ7xaXrtJj1YWU4:-KiYfuJ8uGaMFqFfOy5C:b-sejz2j) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/node_package/src/ReactOnRails.js#L227)):*
> ```JavaScript
>   },
> 
>   webpackConfigLoader(configPath) {
> ```

this should not go in the default react on rails package

---

*[node_package/src/webpackConfigLoader.js, line 8 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgEKudFBuIb3QHo02:-KiYgEKudFBuIb3QHo03:bqdfo5o) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/node_package/src/webpackConfigLoader.js#L8)):*
> ```JavaScript
> export default {
>   configLoader(configPath) {
>     // const configPath = resolve('..', 'config', 'webpack');
> ```

remove comments

```js
const configLoader = (configPath) => {

  return something;
} 

export default configLoader;
``` 
 

---

*[spec/dummy/client/webpack.client.base.config.js, line 40 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgkHNPEy3KX6LGH8r:-KiYgkHNPEy3KX6LGH8s:b-ey9pxx) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/spec/dummy/client/webpack.client.base.config.js#L40)):*
> ```JavaScript
>     alias: {
>       images: join(process.cwd(), 'app', 'assets', 'images'),
>       // 'react-on-rails': resolve(__dirname, '..', '..', '..'),
> ```

remove comments

---

*[spec/dummy/client/webpack.client.rails.hot.config.js, line 11 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYglmkrQnpr7BobqIB:-KiYglmkrQnpr7BobqIC:b-4bi8v1) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/spec/dummy/client/webpack.client.rails.hot.config.js#L11)):*
> ```JavaScript
> 
> const config = require('./webpack.client.base.config');
> const webpackConfigLoader = require('react-on-rails').default.webpackConfigLoader;
> ```

see comments above

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 40 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgnrtCVWJmklbgopM:-KiYgnrtCVWJmklbgopN:b-ey9pxx) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/spec/dummy/client/webpack.server.rails.build.config.js#L40)):*
> ```JavaScript
>     alias: {
>       images: join(process.cwd(), 'app', 'assets', 'images'),
>       // 'react-on-rails': resolve(__dirname, '..', '..', '..'),
> ```

remove comments

---

*[spec/react_on_rails/generators/dev_tests_generator_spec.rb, line 47 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgqB6ETRUdFYRwgJU:-KiYgqB6ETRUdFYRwgJV:b-hfpv1m) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/spec/react_on_rails/generators/dev_tests_generator_spec.rb#L47)):*
> ```Ruby
>     #     assert_match(expected, contents)
>     #   end
>     # end
> ```

remove?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-KiYh-VPXUV4fbkcniUH:b-o6km3h)*

kaizencodes:
Review status: 27 of 38 files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[package.json, line 44 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYergks0kL1ROckKjE:-KiYz2HFkmbPg9fXKN85:b-bte2ho) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/package.json#L44)):*

what is this part for?
</blockquote></details>

couldn't resolve fs that is in the configLoader

---

*[lib/generators/react_on_rails/dev_tests_generator.rb, line 56 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiPAbMeHrpaGs8TPNmA:-KiYzQgylSJuH_adJVnA:b-3mmtq1) ([raw file](https://github.com/shakacode/react_on_rails/blob/1abd20a032890c9f618483b7866ac4bf793ff0a1/lib/generators/react_on_rails/dev_tests_generator.rb#L56)):*

did you figure this out? maybe I can help today.
</blockquote></details>

yes I did, needed to run npm i on the react-on-rails package. which is odd as it seems the rake tasks should do that.

---

*[spec/react_on_rails/generators/dev_tests_generator_spec.rb, line 47 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYgqB6ETRUdFYRwgJU:-KiYyokavnwgUJqIa51j:b-q1qf5p) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/spec/react_on_rails/generators/dev_tests_generator_spec.rb#L47)):*

remove?
</blockquote></details>

I wasn't sure if we need this somewhere or not

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
Good progress. Definitely some more to do!

---

Reviewed 1 of 19 files at r6, 11 of 11 files at r7.
Review status: all files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[package.json, line 44 at r6](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiYergks0kL1ROckKjE:-KiazBON-T9lpaHrISnZ:b-mkgrkz) ([raw file](https://github.com/shakacode/react_on_rails/blob/2c99a5f86e8fd2a3092b3b9e98c2a7ac06cde6ca/package.json#L44)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

couldn't resolve fs that is in the configLoader
</blockquote></details>

yes -- this should be removed, so that we're not loading the config file in the browser.

---

*[rakelib/run_rspec.rake, line 125 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-Kiay0gBBxbggmCqLLFk:-Kiay0gBBxbggmCqLLFl:b-8zpedo) ([raw file](https://github.com/shakacode/react_on_rails/blob/0b18d2b76b3176663875d34190906ceb06a36bf3/rakelib/run_rspec.rake#L125)):*
> ```Ruby
> def gen_assets(dir)
>   path = calc_path(dir)
>   sh_in_dir(path.realpath, "yarn run build:test")
> ```

this might be OK for a bit

we need to investigate if the test helper can do this automatically

---

*[spec/dummy/client/server-rails-hot.js, line 22 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiayFzI_kcogl2CGWdx:-KiayFzI_kcogl2CGWdy:b-5ofwtf) ([raw file](https://github.com/shakacode/react_on_rails/blob/0b18d2b76b3176663875d34190906ceb06a36bf3/spec/dummy/client/server-rails-hot.js#L22)):*
> ```JavaScript
> import webpackConfig from './webpack.client.rails.hot.config';
> 
> const webpackConfigLoader = require('react-on-rails').default.webpackConfigLoader;
> ```

this should be a relative require

---

*[spec/dummy/client/webpack.server.rails.build.config.js, line 11 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiayjercwKRb46nNiqf:-KiayjercwKRb46nNiqg:bym9j7n) ([raw file](https://github.com/shakacode/react_on_rails/blob/0b18d2b76b3176663875d34190906ceb06a36bf3/spec/dummy/client/webpack.server.rails.build.config.js#L11)):*
> const { env, paths, publicPath } = webpackConfigLoader(resolve('..', 'config', 'webpack'));

```js
const configPath = resolve('..', 'config', 'webpack');
const { env, paths, publicPath } = webpackConfigLoader(configPath);
```

---

*[spec/dummy/config/initializers/react_on_rails.rb, line 21 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiaytviXLWjFOBtuFg_:-KiaytviXLWjFOBtuFga:b-u0cgjz) ([raw file](https://github.com/shakacode/react_on_rails/blob/0b18d2b76b3176663875d34190906ceb06a36bf3/spec/dummy/config/initializers/react_on_rails.rb#L21)):*
> ```Ruby
> 
>   # Define the files we need to check for webpack compilation when running tests.
>   config.webpack_generated_files = %w(app-bundle.js vendor-bundle.js server-bundle.js)
> ```

we should not be removing the css files! those should be generated

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-:-Kiaz4dGjyVt0OzERVoP:b1jyiie)*

justin808:
Review status: all files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[spec/dummy/config/initializers/react_on_rails.rb, line 21 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KiaytviXLWjFOBtuFg_:-Kiin5Ez5dOpYngog7qS:b-z9h3en) ([raw file](https://github.com/shakacode/react_on_rails/blob/0b18d2b76b3176663875d34190906ceb06a36bf3/spec/dummy/config/initializers/react_on_rails.rb#L21)):*

we should not be removing the css files! those should be generated
</blockquote></details>

Actually, we didn't have them there before.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
I grabbed your latest commit and rebased my branch on master.

PENDING:
* work on the PR: https://github.com/shakacode/react-webpack-rails-tutorial/pull/387/files#r114082691
* doc describing changes in pr 387 for upgrading linked in the CHANGELOG and README
* Update tutorial.md file in /docs.

justin808:
Review status: 21 of 48 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[docs/basics/migrating-to-webpacker.md, line 38 at r12](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KjMWkUe-DS2xJUw0oVl:-KjMWkUe-DS2xJUw0oVm:b-4v7k9e) ([raw file](https://github.com/shakacode/react_on_rails/blob/4688123b97e612b21ba8087e35103b356b1b9299/docs/basics/migrating-to-webpacker.md#L38)):*
> ```Markdown
> 4. Add the manifest
> [plugin](https://github.com/danethurber/webpack-manifest-plugin) to your webpack config
>   ```
> ```

you never closed this

if you looked at your file as rendered in markdown, this is obvious.

---

*[docs/basics/migrating-to-webpacker.md, line 44 at r12](https://reviewable.io:443/reviews/shakacode/react_on_rails/811#-KjMWYxgN60xl2A4ddCm:-KjMWYxgN60xl2A4ddCn:b-fi9yww) ([raw file](https://github.com/shakacode/react_on_rails/blob/4688123b97e612b21ba8087e35103b356b1b9299/docs/basics/migrating-to-webpacker.md#L44)):*
> ```Markdown
>   Where `publicPath` is the webpack output directory.
> 
> 5. (optional) use the react on rails configLoader
> ```

```js
const webpackConfigLoader = require('react-on-rails/webpackConfigLoader');
const configPath = resolve('..', 'config', 'webpack');
const { paths, publicPath } = webpackConfigLoader(configPath);
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/811)*

justin808:
This PR is still open b/c of the doc commit. @kaizencodes Please move that to a new PR.

justin808:
Closing since this was merged in #822 and we have #835 open.

Issue #782, opened 2017-03-30
[Rails 3.2][v6.9.0] Parsing error when fetching ReactOnRails context from script tag


Ynote:
Hi,

I updated the gem to the version `6.9.0` so it can be compatible with the node module (6.9.0). 
I'm using it with Rails 3.2.

Unexpectedly, I got this parsing error: 
```
Uncaught SyntaxError: Unexpected token i in JSON at position 18
    at JSON.parse (<anonymous>)
    at parseRailsContext (app.js:23226)
    at reactOnRailsPageLoaded (app.js:23235)
    at HTMLDocument.<anonymous> (app.js:23289)
```

This is the output I have in the `script` tag generated by the helper `prepend_render_rails_context`:
```html
<script id="js-react-on-rails-context" type="application/json">
{
  inMailer: false,
  i18nLocale: fr,
  i18nDefaultLocale: fr,
  href: http://localhost:3000/,
  location: /,
  scheme: http,
  host: localhost,
  port: 3000,
  pathname: /,
  search: null,
  httpAcceptLanguage: fr,en;q=0.8,en-US;q=0.6,zh-CN;q=0.4,zh;q=0.2,tr;q=0.2,nl;q=0.2,it;q=0.2,de;q=0.2,es;q=0.2,
  serverSide: false
}
</script>
```

The generated output doesn't seem to be formatted as json so it cannot be parsed correctly by the javascript. Is there something that I should have done in my configuration?

Thank you very much for your help (and your great great project)!

justin808:
@cheremukhin23 It seems the changes don't work for Rails 3.x.

@Ynote Any chance that you can try to submit a PR to fix this?

@Ynote can you downgrade to 6.8.x. I suspect that Rails 3.2 is not compatible with the latest changes.

1. This [call creates the content tag for the context](https://github.com/shakacode/react_on_rails/pull/775/files#diff-741e57aaaefab9559b7892657ec8c75eR241)
2. Here is the [call that formats the content](https://github.com/shakacode/react_on_rails/pull/775/files#diff-741e57aaaefab9559b7892657ec8c75eR225)

https://github.com/cheremukhin23/react_on_rails/blob/82adca11e08651b91f384b1ca23ca498eda9e3d2/app/helpers/react_on_rails_helper.rb#L225
```ruby
  def json_safe_and_pretty(something)
    if Rails.env.development?
      ERB::Util.json_escape(JSON.pretty_generate(something.as_json))
    else
[...]

PR #787, opened 2017-03-31
Improve json conversion with tests and support for older Rails 3.x

cheremukhin23:
- Add error handling when unable to parse json
- Add tests to ensure json_safe_and_pretty handles hashes and strings

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/787)

PR #760, opened 2017-03-11
Use <script type="application/json"> for props and store

squadette:
See discussion in https://github.com/shakacode/react_on_rails/pull/660

Build passing: https://travis-ci.org/squadette/react_on_rails/builds/210095090

Thank you @eliaslopezgt for initial implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/760)

PR #812, opened 2017-04-19
[Extends #787] - Improve json conversion with tests and support for Rails older than 4.1

Ynote:
- Closes https://github.com/shakacode/react_on_rails/issues/782.
- Updates https://github.com/shakacode/react_on_rails/pull/787 with some tests.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/812)


spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
```diff
@@ -10,24 +10,71 @@
...
+
+    shared_examples :escaped_json do
+      it "returns a well-formatted json with escaped characters" do
+        expect(helper.json_safe_and_pretty(special_characters_hash))
+          .to eq(special_characters_sanitized_hash)
+      end
+    end
+
+    context 'with Rails 3' do
+      before { allow(Rails).to receive(:version).and_return(3) }
```
Ynote: I'm not really sure that mocking the version is the better way to do. 

Could it be a good idea to configure Travis to lauch the tests with differents Rails version? I'm just raising the question, I know this imply to extend the tests duration and to change the Travis configuration.

Ynote:
It seems I cannot fetch the upstream to be up-to-date with the `master` branch.

justin808:
Please resolve the conflicts a add a test with a Rails version of 10. We can't easily run CI on rails 32. If you want to submit a PR on that, that would be awesome. Contact me if you want advice on how to do that.

justin808:
A couple comments. I'd like to see the escaping part refactored to it's own class for better testing.

CC: @cheremukhin23 @ynote @squadette 

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---

*[CHANGELOG.md, line 9 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki6rj1vdsayXWHmmBr5:-Ki6rj1vdsayXWHmmBr6:b-3k5vbw) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/CHANGELOG.md#L9)):*
> ```Markdown
> 
> ### Improved
> - Improve json conversion with tests and support for older Rails 3.x. [#787](https://github.com/shakacode/react_on_rails/pull/787) by [cheremukhin23](https://github.com/cheremukhin23).
> ```

@ynote you can add you as an author as well ;-)

---

*[app/helpers/react_on_rails_helper.rb, line 229 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki6rqBDSg9tHGfFoen-:-Ki6rqBDSg9tHGfFoen0:b-bfqe8n) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/app/helpers/react_on_rails_helper.rb#L229)):*
> ```Ruby
>   def json_safe_and_pretty(hash_or_string)
>     unless hash_or_string.class.in?([Hash, String])
>       raise "#{hash_or_string.class} is unsupported argument class for this method"
> ```

list what types are supported in the error message

---

*[app/helpers/react_on_rails_helper.rb, line 243 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki6rz4RvN-2862TEXqV:-Ki6rz4RvN-2862TEXqW:bc3mqi8) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/app/helpers/react_on_rails_helper.rb#L243)):*
> ```Ruby
> 
>   def rails_version_less_than(version)
>     Gem::Version.new(Rails.version) <= Gem::Version.new(version)
> ```

need tests for this this

main issue is that we originally had string comparisons and "10.0" is < "4"

In order to best test private methods like this, I'd like to see this escaping functionality moved to a different ruby class. 

That will work best with @squadette's idea to pretty print in some cases. See #760.  

---

*[spec/dummy/spec/helpers/react_on_rails_helper_spec.rb, line 59 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki5fB0HTPOzEcf_x1pb-r2-59:-Ki6smk3xlmZN3V9W65_:b-fapm29) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb#L59)):*
<details><summary><i>Previously, Ynote (Little Cheung) wrote…</i></summary><blockquote>

I'm not really sure that mocking the version is the better way to do. 

Could it be a good idea to configure Travis to lauch the tests with differents Rails version? I'm just raising the question, I know this imply to extend the tests duration and to change the Travis configuration.
</blockquote></details>

Yes, but it would take (see #814):

1. /spec/dummy-rails3 (like spec/dummy, but with Rails 3)
2. pass env value on travis runs for different ruby versions and ONLY run rails 3 tests (spec/dummy-rails3) and unit tests) for the old ruby version. Don't run /spec/dummy and the generator tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-Ki6u6pAPtRPXIPXmNjC:bpro41o)*

Ynote:
Review status: 0 of 7 files reviewed at latest revision, 4 unresolved discussions.

---

*[app/helpers/react_on_rails_helper.rb, line 243 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki6rz4RvN-2862TEXqV:-KiB-v-v3euS1k4-0Tjr:b-uryjzf) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/app/helpers/react_on_rails_helper.rb#L243)):*

need tests for this this

main issue is that we originally had string comparisons and "10.0" is < "4"

In order to best test private methods like this, I'd like to see this escaping functionality moved to a different ruby class. 

That will work best with @squadette's idea to pretty print in some cases. See #760.  
</blockquote></details>

I moved this method into ReactOnRails::Util as it could be used elsewhere too. Is it ok for you?

---

*[spec/dummy/spec/helpers/react_on_rails_helper_spec.rb, line 59 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-Ki5fB0HTPOzEcf_x1pb-r2-59:-KiB0M5cRZ7UQtGAC7HS:b-3bqwdy) ([raw file](https://github.com/shakacode/react_on_rails/blob/c521f1599d789053990ab9e7449ff6f21d4755ed/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb#L59)):*

Yes, but it would take (see #814):

1. /spec/dummy-rails3 (like spec/dummy, but with Rails 3)
2. pass env value on travis runs for different ruby versions and ONLY run rails 3 tests (spec/dummy-rails3) and unit tests) for the old ruby version. Don't run /spec/dummy and the generator tests.
</blockquote></details>

Ok, I'll have a look at this later. I'll try to find some time to work on it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812)*

justin808:
Wow! @ynote **GREAT JOB!** 

This is really **OUTSTANDING**!

:lgtm_strong: 

---

Reviewed 7 of 7 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiER1ljqK3SmXUihEUr:b-cbau3b)*

justin808:
@Ynote Please be sure that your PR passes CI.

You have many linter errors. Many will be fixed by running `rubocop -a`.

justin808:
@Ynote Do you need any help from me to get this one finished up? I want to do one last 6.x release before I merge in the webpacker changes and go to 7.0.0-beta.0

justin808:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/json_output.rb, line 13 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiJpTBSsC82NiEMsjF6:-KiJpTBSsC82NiEMsjF7:b-stas03) ([raw file](https://github.com/shakacode/react_on_rails/blob/21d502a65a0f2cdfe5ecffd3fe9ec94bfe31503a/lib/react_on_rails/json_output.rb#L13)):*
> ```Ruby
> 
>     def escaped
>       return escaped_without_erb_utils if Utils::rails_version_less_than("4")
> ```

We need to confirm this is less than 4 and not less than 4.2, per #815  and #816 by @EdmundLeex 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812)*

Ynote:
Hi, 
I'm sorry, I wasn't available yesterday. I'm gonna have a look at these errors and fix it now.

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/json_output.rb, line 13 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiJpTBSsC82NiEMsjF6:-KiKKsZubmS9SmxT1ay2:b-7mn9sw) ([raw file](https://github.com/shakacode/react_on_rails/blob/21d502a65a0f2cdfe5ecffd3fe9ec94bfe31503a/lib/react_on_rails/json_output.rb#L13)):*

We need to confirm this is less than 4 and not less than 4.2, per #815  and #816 by @EdmundLeex 
</blockquote></details>

Ok, I'm gonna change the tested version number.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiKKhMppxVequI0ybAP:bkp7hkm)*

Ynote:
Sorry I didn't run rubocop before -_-" It is fixed now! Hope it will be ok for your release.

---

Review status: 3 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/json_output.rb, line 13 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiJpTBSsC82NiEMsjF6:-KiKRt7OM53DDPMv-m4V:b-yn58jh) ([raw file](https://github.com/shakacode/react_on_rails/blob/21d502a65a0f2cdfe5ecffd3fe9ec94bfe31503a/lib/react_on_rails/json_output.rb#L13)):*
<details><summary><i>Previously, Ynote (Little Cheung) wrote…</i></summary><blockquote>

Ok, I'm gonna change the tested version number.
</blockquote></details>

After readinghttps://github.com/shakacode/react_on_rails/issues/815 and https://github.com/shakacode/react_on_rails/pull/816, I suppose what you meant was to confirm this is less than 4.2 and not 4 :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiKS7lTjPA-dDOVs9ni:b47ky77)*

EdmundLeex:
I have a couple of maybe naive questions. 
- Why not just reimplement the `ERB::Util::json_escape` method since its a fairly simple one instead of checking the Rails version? (this seems adds unnecessary performance overhead, mainly in doing `Gem.new(...)`)
- Maybe a class method instead of a new `JsonOutput` class and instantiate an instance every time? Ruby object instantiation is costly and since there is not state changes, seems creating an object is just adding performance overhead. (I brought this up because we are at a stage where performance really matter)

This method is called in every request and maybe multiple times in each request. Think we might want to pay more attention to performance.

justin808:
@EdmundLeex:
> Why not just reimplement the ERB::Util::json_escape method since its a fairly simple one instead of checking the Rails version? (this seems adds unnecessary performance overhead, mainly in doing Gem.new(...))

Suppose an XSS is discovered. We want to leverage what's in the Rails code.

@Ynote, @EdmundLeex:  can we do some small memoization, maybe at the class level to address @EdmundLeex's concerns. I agree with him that anything in the rendering cycle should be fairly efficient. That being said. I bet if this were profiled, there would be ZERO difference if we optimized this.

**I need to get this merged and shipped this weekend, as I'm about to go beta for 7.0.0**

justin808:
Review status: 3 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/json_output.rb, line 13 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiJpTBSsC82NiEMsjF6:-KiMQ0ooG-0NZmrcFGG_:b-ifjh40) ([raw file](https://github.com/shakacode/react_on_rails/blob/21d502a65a0f2cdfe5ecffd3fe9ec94bfe31503a/lib/react_on_rails/json_output.rb#L13)):*
<details><summary><i>Previously, Ynote (Little Cheung) wrote…</i></summary><blockquote>

After readinghttps://github.com/shakacode/react_on_rails/issues/815 and https://github.com/shakacode/react_on_rails/pull/816, I suppose what you meant was to confirm this is less than 4.2 and not 4 :)
</blockquote></details>

I'm not 100% sure on the exact version to compare. This sounds reasonable!

---

*[lib/react_on_rails/json_output.rb, line 19 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiMPYfhYuBBYgznCoJO:-KiMPYfiGFbk6ieG7jWb:bvrx7nl) ([raw file](https://github.com/shakacode/react_on_rails/blob/460eb372fabd0f345350832897a699403cdd5c04/lib/react_on_rails/json_output.rb#L19)):*
> ```Ruby
> 
>     def escaped
>       return escaped_without_erb_utils if Utils.rails_version_less_than("4.2")
> ```

@ynote Let's add a method that encapsulates the 4.2 check and memoize the result.

We also can make the method static, since this class is so simple.

I believe that would address @edmundleex concerns. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812)*

justin808:
Reviewed 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812)*

EdmundLeex:
FYI: 4.1.1 was the first version that corrected this issue.

Ynote:
Thanks @EdmundLeex for your concern about performance!

@justin808, I added the memoization and follow @EdmundLeex's advice using the test on Rails 4.1.1.

---

Review status: 3 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiMoL9awMMCb1KArmgW:bc3pmc8)*

lib/react_on_rails/utils.rb
```diff
@@ -25,7 +25,14 @@ def self.running_on_windows?
     end
 
     def self.rails_version_less_than(version)
-      Gem::Version.new(Rails.version) < Gem::Version.new(version)
+      @rails_version_less_than ||= {}
+      @rails_version_less_than[version] = begin
+        Gem::Version.new(Rails.version) < Gem::Version.new(version)
+      end
```
EdmundLeex: hmm. I believe this block runs every time instead of using the memoized result?

app/helpers/react_on_rails_helper.rb
```diff
@@ -224,6 +225,17 @@ def server_render_js(js_expression, options = {})
...
 
+  def json_safe_and_pretty(hash_or_string)
+    unless hash_or_string.class.in?([Hash, String])
+      raise "#{__method__} only accepts String or Hash as argument "\
+            "(#{hash_or_string.class} given)."
+    end
+
+    json_value = hash_or_string.is_a?(String) ? hash_or_string : hash_or_string.to_json
+
+    ReactOnRails::JsonOutput.new(json_value).escaped
```
EdmundLeex: This is where the performance might suffer because a new `JsonOutput` object is instantiated each time. Rather, `JsonOutput.escape_json(...)`, with `escape_json` being a class method can avoid this instantiation.

Ynote: @EdmundLeex You are right. I'm gonna update it.

lib/react_on_rails/utils.rb
Ynote: True. I'm gonna fix it, thanks for the review. Can you tell me why my test passes and so, is not correct? https://github.com/shakacode/react_on_rails/pull/812/files#diff-ad2d9ab6a9a22bf3f00405aba4094cbaR38

spec/react_on_rails/utils_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        it { expect(subject).to eq(false) }
+      end
+
+      context "called twice" do
+        before do
+          allow(Rails).to receive(:version).and_return("4.2")
+        end
+
+        it "should memoize the result" do
+          2.times { subject }
```
EdmundLeex: this is because `subject` is memoized by rspec.

lib/react_on_rails/utils.rb
EdmundLeex: this is because `subject` is memoized by rspec.

justin808:
:lgtm:

:lgtm_strong: 

---

Reviewed 3 of 3 files at r6, 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 39 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiOIJ90I65R-dZ3iXQ-:-KiOIJ90I65R-dZ3iXQ0:b-87qx5e) ([raw file](https://github.com/shakacode/react_on_rails/blob/f5feb91b7f00eefc20b31234e448876833153790/lib/react_on_rails/utils.rb#L39)):*
> ```Ruby
>     end
> 
>     def self.rails_version_less_than_4_1_1
> ```

was the bug fixed in 4.1.1? (rather than 4.2?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiOISo39FMebUAx52kd:b-e3pzy9)*

justin808:
@ynote Please confirm that we should merge and release.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiOJ902YuYEqM5etE0F:b-lgk8ii)*

Ynote:
I just tested the PR again with my stack. Everything is ok for me. I think we can merge and release.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 39 at r7](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-KiOIJ90I65R-dZ3iXQ-:-KiOXIq05sI37xcKWW2Z:bn6mi56) ([raw file](https://github.com/shakacode/react_on_rails/blob/f5feb91b7f00eefc20b31234e448876833153790/lib/react_on_rails/utils.rb#L39)):*

was the bug fixed in 4.1.1? (rather than 4.2?)
</blockquote></details>

This is the PR that fixes the eating quotes issue and the issue with `\u2028` and `\u2029`special characters: https://github.com/rails/rails/pull/13073. It was merged in Rails 4.1. So, I think it is ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/812#-:-KiObfJ6WwSvqzLl_rB1:b7zm1t)*

justin808:
Thank you @Ynote!

Issue #819, opened 2017-04-24
Performance decrease after 6.9 due to JSON.parse of already json string


dzirtusss:
There is some performance decrease for RoR versions after 6.9 which looks like this (from our internal server-side benchmarks):
```
----------------
 Benchmarking master VS test-ror
----------------
visitor_test1              79.751ms VS 92.823ms, +16.39% BAD !!!
visitor_test2              433.685ms VS 493.804ms, +13.86% BAD !!!
visitor_test3              123.157ms VS 134.358ms, +9.09% BAD !!!
user_test1                 131.78ms VS 144.51ms, +9.66% BAD !!!
user_test2                 1038.557ms VS 1128.468ms, +8.65% BAD !!!
user_test3                 162.683ms VS 174.69ms, +7.38% BAD !!!
```
The reason for it is JSON.parsing of the already json string to hash (and later back), which should be generally just a copy-paste operation
```
props.is_a?(String) ? JSON.parse(ERB::Util.json_escape(props)) : props
```
https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/react_component/options.rb#L17

justin808:
Reverting the line here:

https://github.com/shakacode/react_on_rails/commit/7ccd603683af377880d4e16f56752ac7587c38b7#diff-ab00ff0a0648717a06431522f7eb6861R16

justin808:
Fixed by #821 and shipped in 7.0.0.

PR #821, opened 2017-04-24
Raise if versions differ and fix performance regression for string props

justin808:
## Raise on Gem and Node Version Mismatch

Because many installations did not notice issues when versions differed, we're now going to force upgrading both the gem and node package at the same time.

## Fix Performance Issue on String Props

6.9 introduced code that would parse the props into a Hash
unnecessarily. Any customers with significant String props took a huge
performance hit from this on page rendering.

Fix is to remove the unnecessary parsing.

I believe the parsing was done so that pretty printing of the JSON could
be done. If we introduce that feature in the future, it will:

* Be configurable on whether or not to use it.
* Possibly never run in production.

See #819.

CC: @cheremukhin23 @robwise @dylangrafmyre @squadette 

Thanks @dzirtusss for finding this! AWESOME!!!

See https://github.com/shakacode/react_on_rails/commit/7ccd603683af377880d4e16f56752ac7587c38b7#diff-ab00ff0a0648717a06431522f7eb6861R16

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/821)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 # Change Log
-All notable changes to this project's source code will be documented in this file. Items under `Unreleased` is upcoming features that will be out in next version.
+All notable changes to this project's source code will be documented in this file. Items under `Unreleased` is upcoming features that will be out in next version. NOTE: major versions of the npm module and the gem must be kept in sync.
 
 Contributors: please follow the recommendations outlined at [keepachangelog.com](http://keepachangelog.com/). Please use the existing headings and styling as a guide, and add a link for the version diff at the bottom of the file. Also, please update the `Unreleased` link to compare to the latest release version.
 
 ## [Unreleased]
+## [7.0.0] - 2017-04-25
+### Changed
+- Any version differences in gem and node package for React on Rails throw an error [#821](https://github.com/shakacode/react_on_rails/pull/821) by [justin808](https://github.com/justin808)
+
+### Fixed
+- Fixes serious performance regression when using String props for rendering. [#821](https://github.com/shakacode/react_on_rails/pull/821) by [justin808](https://github.com/justin808)
+
 ## [6.10.1] - 2017-04-23
 ### Fixed
 - Improve json conversion with tests and support for older Rails 3.x. [#787](https://github.com/shakacode/react_on_rails/pull/787) by [cheremukhin23](https://github.com/cheremukhin23) and [Ynote](https://github.com/Ynote).
@@ -534,7 +541,8 @@ Best done with Object destructing:
 ##### Fixed
 - Fix several generator related issues.
 
-[Unreleased]: https://github.com/shakacode/react_on_rails/compare/6.10.1...master
+[Unreleased]: https://github.com/shakacode/react_on_rails/compare/7.0.0...master
+[7.0.0]: https://github.com/shakacode/react_on_rails/compare/6.10.1...7.0.0
 [6.10.1]: https://github.com/shakacode/react_on_rails/compare/6.10.0...6.10.1
 [6.10.0]: https://github.com/shakacode/react_on_rails/compare/6.9.3...6.10.0
 [6.9.3]: https://github.com/shakacode/react_on_rails/compare/6.9.1...6.9.3
--- a/lib/react_on_rails/version_checker.rb
+++ b/lib/react_on_rails/version_checker.rb
@@ -2,44 +2,49 @@ module ReactOnRails
   # Responsible for checking versions of rubygem versus yarn node package
   # against each otherat runtime.
   class VersionChecker
-    attr_reader :node_package_version, :logger
-    MAJOR_VERSION_REGEX = /(\d+)\.?/
+    attr_reader :node_package_version
+    MAJOR_MINOR_PATCH_VERSION_REGEX = /(\d+)\.(\d+)\.(\d+)/
 
     def self.build
-      new(NodePackageVersion.build, Rails.logger)
+      new(NodePackageVersion.build)
     end
 
-    def initialize(node_package_version, logger)
-      @logger = logger
+    def initialize(node_package_version)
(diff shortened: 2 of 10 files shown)


justin808:
FYI: Some of the tests need fixing. I'll get to this later today. If anybody can fix these sooner, I'll take your PR. Just make a fork and a new branch.

justin808:
If anybody wants to tackle fixing the te unit tests on this one, I'd love to get the help.

justin808:
Can I get a code review?

lib/react_on_rails/version_checker.rb
```diff
@@ -2,44 +2,49 @@ module ReactOnRails
...
+
+      raise_differing_versions_warning
     end
 
     private
 
-    def log_differing_versions_warning
-      msg = "**WARNING** ReactOnRails: ReactOnRails gem and node package MAJOR versions do not match\n" \
+    def raise_differing_versions_warning
+      msg = "**ERROR** ReactOnRails: ReactOnRails gem and node package versions do not match\n" \
```
pjfitzgibbons, line 33: I'm confused?  The PR/Issue and this text says "MAJOR must match..."... but the code above at ln:23 and tests below are requiring all 3 parts to be exact-match.   If the logic above is correct, maybe some different wording here to indicate the versions must be exact-match?

justin808: Should I say "major, my minor, and patch versions"?

pjfitzgibbons: Ok, So I think I mis-read the diff at ln29:33, but there is no-change at ln: 32:36 that I think should change to read `"Ensure the installed version of the gem is the same as the version of \n"\`
At Changelog.md:2, maybe should say `NOTE: exact versions of the npm module and the gem must be kept in sync.`

lib/react_on_rails/version_checker.rb
```diff
@@ -2,44 +2,49 @@ module ReactOnRails
   # Responsible for checking versions of rubygem versus yarn node package
   # against each otherat runtime.
   class VersionChecker
-    attr_reader :node_package_version, :logger
-    MAJOR_VERSION_REGEX = /(\d+)\.?/
+    attr_reader :node_package_version
+    MAJOR_MINOR_PATCH_VERSION_REGEX = /(\d+)\.(\d+)\.(\d+)/
```
kaizencodes, line 6: A bit confusing name. VERSION_REGEX might be more readable.

lib/react_on_rails/version_checker.rb
```diff
@@ -2,44 +2,49 @@ module ReactOnRails
...
+      raise msg
     end
 
     def gem_version
       ReactOnRails::VERSION
     end
 
-    def gem_major_version
-      gem_version.match(MAJOR_VERSION_REGEX)[1]
+    def gem_major_minor_patch_version
```
kaizencodes, line 45: Same here, could be just: gem_version

CHANGELOG.md
```diff
@@ -1,9 +1,16 @@
 # Change Log
-All notable changes to this project's source code will be documented in this file. Items under `Unreleased` is upcoming features that will be out in next version.
+All notable changes to this project's source code will be documented in this file. Items under `Unreleased` is upcoming features that will be out in next version. NOTE: major versions of the npm module and the gem must be kept in sync.
```
kaizencodes, line 2: Drop the major. NOTE: versions of the npm module and the gem must be kept in sync.

pjfitzgibbons: @kaizencodes has a great point.  Since major-minor-patch-build/alpha/beta must match on both sides (?!), why not just direct-match the strings and be done with it?  What effect will this have on "local-development" activity against ReactOnRails directly ? @justin808 

If the desire is to match "only" M.M.P, then yes I agree rename the const VERSION_REGEX and also maybe pin the rx to begin-of-string `VERSION_REGEX = /^(\d+)\.(\d+)\.(\d+)/`

justin808: @pjfitzgibbons b/c of the way that beta versions in npm/rubygems use dash versus dots.

justin808: @pjfitzgibbons on the beginning of the string, you might have a '^'

mapreal19:
Reviewed 9 of 10 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/version_checker.rb, line 6 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/821#-KiZbwuxX8jKlw3vlcpN-r1-6:-KifIfJ-6MeOkauLr4ZV:b-7tdf7d) ([raw file](https://github.com/shakacode/react_on_rails/blob/70ffd7661590a0d9549d488dc53129fac137e745/lib/react_on_rails/version_checker.rb#L6)):*
<details><summary><i>Previously, kaizencodes (Daniel Szatmari) wrote…</i></summary><blockquote>

A bit confusing name. VERSION_REGEX might be more readable.
</blockquote></details>

agree

---

*[node_package/src/clientStartup.js, line 104 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/821#-KifJnLxg0YULVfLIxAb:-KifJnLyerpTneL2Z7Ts:b-psaw3t) ([raw file](https://github.com/shakacode/react_on_rails/blob/ef40698fc11cb8a8ae2ef5ddef79696db3f7d0c6/node_package/src/clientStartup.js#L104)):*
> ```JavaScript
> function render(el, railsContext) {
>   const context = findContext();
>   // This must match app/helpers/react_on_rails_helper.rb:113
> ```

I wouldn't rely on a comment for this. Maybe use a constant for both ruby and js code

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/821)*

PR #823, opened 2017-04-27
Fix to handle nil values in json_safe_and_pretty

dzirtusss:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/823)


app/helpers/react_on_rails_helper.rb
```diff
@@ -231,6 +231,7 @@ def server_render_js(js_expression, options = {})
   end
 
   def json_safe_and_pretty(hash_or_string)
+    return if hash_or_string.nil?
```
justin808, line 234: can we add a quick test?

justin808:
Great job @dzirtusss!

PR #825, opened 2017-04-28
Suggested edits to README

tassock:
Hello,

I just installed react_on_rails in my project today and it's working great! Thanks for the time you put into this. Here's a couple minor changes in a README that looked off to me, perhaps because because you just released version 7 this week. 

* Reference version 7 in gemfile
* Reference "HelloWorld" component that generator created rather than "HelloWorldApp" component that I didn't find. 

Thanks again! 

Peter

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/825)


justin808:
Awesome @tassock!

Issue #697, opened 2017-01-27
Add `Rails.env` to RailsContext


robwise:
[Heroku Preboot](https://devcenter.heroku.com/articles/preboot) is a feature on Heroku that allows for faster deploy times. When you promote your staging app to production, Preboot simply switches the production server to point at the staging app's container. This means it can deploy much faster since it doesn't have to rebuild anything.

However, this means that if you use the [Define Plugin](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin) to provide the rails environment to your client code as a variable, that variable will erroneously still have a value of `Staging` instead of `Production`. Providing `Rails.env` (which is accurate even in Preboot) in the default Rails context would be a nice enhancement.

justin808:
@robwise Easy to add, either via a custom Rails Context enhancement (no PR required) or as a default for the RailsContext.

See this example: https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/config/initializers/react_on_rails.rb#L4

```ruby
module RenderingExtension
  # Return a Hash that contains custom values from the view context that will get passed to
  # all calls to react_component and redux_store for rendering
  def self.custom_context(view_context)
    if view_context.controller.is_a?(ActionMailer::Base)
      {}
    else
      {
        somethingUseful: view_context.session[:something_useful]
      }
    end
  end
end
```

robwise:
@justin808 yeah I'm already doing it the custom way, I'm saying I think it should be a default

justin808:
@robwise Maybe for features, we'll do if any of our [Coaching Clients](http://www.shakacode.com/work/shakacode-coaching-plan.pdf) or regular clients want it? Otherwise, we can wait for the PR to come in. Or maybe if the issue gets at least 20 up votes.

robwise:
okay, and maybe just having this issue as a historical record will be good enough in case other people are using preboot and are not realizing that because the webpack isn't being recompiled, the variables may be stale and to watch out for that

Conturbo:
i'm gonna take it

justin808:
Fixed in #1065.

PR #842, opened 2017-05-16
Rails.env added to railsContext

Conturbo:
Rails.env added to railsContext as a default, as suggested in Issue #697  by @robwise 
(Changes are in **react_on_rails_helper.rb**, line 443-446)

I added comments to **spec/dummy/Procfile.static** explaining how to use binding.pry and webrick (because puma acts-up sometimes). If these instructions are useful, maybe they should be included in the Procfile generated by the react_on_rails generator?

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/842)


justin808:
- [ ] We need tests
- [ ] CHANGELOG entry needed
- [ ] README and other docs

---

Reviewed 3 of 3 files at r1.

---

*[spec/dummy/Procfile.static, line 13 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/842#-KkHvJhFZmmXWmbBpOCZ:-KkHvJhGjuD-OJvwGezR:b-vekxjp) ([raw file](https://github.com/shakacode/react_on_rails/blob/2a4983b64ef28ed985036238776696f9a2d1905e/spec/dummy/Procfile.static#L13)):*
> ```
> # Comment out line 2, then run this procfile in your terminal like normal,
> # then, open up a new tab run "rails server webrick"
> # you'll be able to connect on localhost:3000
> ```

set editor so you automatically add CR to end of file

this should not go in this PR, but a separate one for RoR.

also, a comment for the Rails part about PORT might make sense

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/842#-:-KkHvlCKhJipKmTeMVBO:by058ss)*

PR #857, opened 2017-05-29
Let it be MainPage

oklas:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/857)


justin808:
We're not changing the name from HelloWorld. If you want to submit a PR to make the name used an option in the generator, that would be great.

PR #873, opened 2017-06-13
RFC: Add options block so that webpack 2.6 won't complain

ened:
Also, I had to include `autoprefixer` in the list of plugins to get it working here.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/873)

### DIFF
--- a/spec/dummy/client/webpack.client.rails.build.config.js
+++ b/spec/dummy/client/webpack.client.rails.build.config.js
@@ -52,7 +52,12 @@ module.exports = merge(config, {
                 localIdentName: '[name]__[local]__[hash:base64:5]',
               },
             },
-            'postcss-loader',
+            {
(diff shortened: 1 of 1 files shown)


justin808:
@ened can you paste the error you see without this? If you want to upgrade the webpack (and other library versions), then please put all into this PR.

ened:
@justin808 - in the end, this might not even be web pack related, but is an artifact (during previous merges). I noticed it worked with the `spec/dummy/Procfile.hot` right away, which had this fixed.

Typical error messages are: 
> Module build failed: ModuleBuildError: Module build failed: Error: No PostCSS Config found in ...

justin808:
@ened What shall we do with this PR?

justin808:
@ened Please get back to me on this one and let me know if we should re-open this PR.

Issue #879, opened 2017-06-22
Documentation missing/unclear on RailsContext?


ceres629:
I can't seem to find any solid information on how railsContext works and how to access the railsContext from within a component.

I have some server side size variables such as ENV variables that I want react components to be able to access. Normally I'd just pass them as props however reading the documentation it says:

> While you could manually pass the railsContext information in as "props", the rails_context is a convenience because it's passed consistently to all invocations of generator functions.

I thought this mean I could set up a rails_context to contain the information i need, and then whenever I use react_component it will have a railsContext parameter automatically added that I can use to access the railsContext from within the component.

the definition of these components are

export default (_props, _railsContext) => {
 }

which seems to suggest some how there is a railsContext being passed in. But although my _props contain values as expected, _railsContext is always null and I can't find any documentation that explains how you pass this value in. Searched online for a few hours to no avail. In the end I just decided to pass rails_context in as a prop, which works, but I'm still confused as to what this railsContext is for or if it is broken or just lacking documentation.

MiguelSavignano:
Hi, @ceres629 
I use the rails context a lot in my app. To sure you load the context appropriately
 see https://facebook.github.io/react/docs/context.html.

The child component need define contextTypes
```javscript
Button.contextTypes = {
  color: PropTypes.string
};
```
I miss a good documentation for react on rails context, I implemented a custom context provider and pass the context in as a prop.

justin808:
I created an example:

https://github.com/shakacode/v8-demo/pull/2

I'm working on making the documentation more clear.

justin808:
See #886, @MiguelSavignano and @ceres629. Please review.

MiguelSavignano:
The example with generator function it's great but Why if my component it's a React Class the context don't appear.
```javascript
import React from 'react';
class RailsContextExample extends React.Component {
  render(){
    return (
      <div>
        <p>props are: {JSON.stringify(this.props)}</p>
        <p>railsContext is:
          {JSON.stringify({ ...this.context, serverSide: false})}}
        </p>
      </div>
    )
  }
}
export default RailsContextExample;
```

And other thing i try to add new context key like params_action or current_user_id I try to override rails_context helper method but this method it's private and I can't do this.

justin808:
[...]

PR #886, opened 2017-07-02
Update railsContext documentation and fix generator

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/886)


justin808:
This addresses #879.

justin808:
@MiguelSavignano and @ceres629, I still welcome your feedback here.

ceres629:
Yes this is great, definitely much more useful, and while I did figure it out in the end, it would have saved me many hours of confusion and trial and error! Hopefully it does the same for new users.

Thanks!

PR #800, opened 2017-04-10
Return markup hash form react component helper

udovenko:
* Implement handling of Hash result of server_rendered_react_component_html in react_on_rails helper
* Implement additional example page for server rendered HTML that shows usage of htmlResult object
* Add conditional title to dummy app layout
* Implement rendering title string in generator function on server
* Add test for conditional title to integration_spec

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/800)

PR #950, opened 2017-10-03
Fix react_component bug

Judahmeek:
Expected behavior: If the user signals ReactOnRails that the user expects `react_component` to return a Hash (which is currently done through https://github.com/shakacode/react_on_rails/pull/800/files#diff-04c6e90faac2675aa89e2176d2eec7d8R343), then `react_component` should return a Hash or nil in all cases.

Actual behavior: The current signal for `react_component` to return a Hash requires successful rendering. Any error in the react code will result in the signal being lost and `react_component` returning a string.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/950)

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -94,7 +94,9 @@ module ReactOnRailsHelper
   #                   so long as you have the default configuration of logging_on_server set to
   #                   true, you'll still see the errors on the server.
   #   raise_on_prerender_error: <true/false> Default to false. True will raise exception on server
-  #      if the JS code throws
+  #                             if the JS code throws
+  #   hash_result: <true/false> Default to false. True will guarantee that react_component returns a hash_result
+  #                even if there is a prerendering error
   # Any other options are passed to the content tag, including the id.
   def react_component(component_name, raw_options = {})
     # Create the JavaScript and HTML to allow either client or server rendering of the
@@ -124,7 +126,12 @@ module ReactOnRailsHelper
                                                   trace: options.trace,
                                                   raise_on_prerender_error: options.raise_on_prerender_error)
 
-    server_rendered_html = result["html"]
+
+    if result["hasErrors"] && options.hash_result
+      server_rendered_html = Hash[COMPONENT_HTML_KEY, result["html"]]
+    else
+      server_rendered_html = result["html"]
+    end
     console_script = result["consoleReplayScript"]
 
     if server_rendered_html.is_a?(String)
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -93,6 +93,7 @@ module ReactOnRails
       replay_console: true,
       logging_on_server: true,
       raise_on_prerender_error: false,
+      hash_result: false,
       trace: Rails.env.development?,
       development_mode: Rails.env.development?,
       server_renderer_pool_size: 1,
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -114,26 +114,24 @@ module ReactOnRails
 
   class Configuration
     attr_accessor :node_modules_location, :server_bundle_js_file, :prerender, :replay_console,
-                  :trace, :development_mode,
-                  :logging_on_server, :server_renderer_pool_size,
+                  :trace, :development_mode, :logging_on_server, :server_renderer_pool_size,
                   :server_renderer_timeout, :skip_display_none, :raise_on_prerender_error,
-                  :generated_assets_dirs, :generated_assets_dir,
+                  :generated_assets_dirs, :generated_assets_dir, :hash_result,
                   :webpack_generated_files, :rendering_extension, :build_test_command,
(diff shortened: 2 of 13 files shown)


Judahmeek:
Tests only fail because of the the same issue failing all recent react_on_rails tests: something in the HelloWorldApp used by the Basic Example. I think it's likely https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/templates/dev_tests/spec/features/hello_world_spec.rb#L10, but I have yet to confirm this.

udovenko:
Generally LGTM, just few comments...

---

Review status: 0 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[spec/dummy/app/views/pages/broken_app.html.erb, line 1 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/950#-Kv_NlDDYtmUckAY_5AK:-Kv_NlDDYtmUckAY_5AL:bu372yv) ([raw file](https://github.com/shakacode/react_on_rails/blob/02f817c7188800442f9c7f0ab1b79e8cf7fa4fcd/spec/dummy/app/views/pages/broken_app.html.erb#L1)):*
> ```erb
> <%= render "header" %>
> ```

broken_app name seems too abstract here (while BrokenApp component name is ok). I would express that we're demonstrating hash result behavour in template name.

---

*[spec/dummy/spec/features/integration_spec.rb, line 200 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/950#-Kv_OWisJtHrlNpXrOzQ:-Kv_OWisJtHrlNpXrOzR:br86he1) ([raw file](https://github.com/shakacode/react_on_rails/blob/02f817c7188800442f9c7f0ab1b79e8cf7fa4fcd/spec/dummy/spec/features/integration_spec.rb#L200)):*
> ```Ruby
>   scenario "react_component should return hash" do
>     expect(subject.html).to include("Exception in rendering!")
>   end
> ```

Should we also add a "black" test for case when we retruning a hash but `hash_result == false`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/950#-:-Kv_OkdTU7rdWbvIvlMB:bi2q8eu)*

udovenko:
Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/950)*

justin808:
1. I'm not clear on why we need a new option on the error handling. @robwise has commented that we give too many options on this.
2. If we were to add an option, the PR needs docs...

I'd like to see "it just works" behavior. Is there possibly a super clear choice as to which is behavior is the only one we need?

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/950#-:-KvaMKxvy4sZU2dY-q2o:b-6wlvyi)*

app/helpers/react_on_rails_helper.rb
```diff
@@ -124,7 +126,12 @@ def react_component(component_name, raw_options = {})
                                                   trace: options.trace,
                                                   raise_on_prerender_error: options.raise_on_prerender_error)
 
-    server_rendered_html = result["html"]
+
+    if result["hasErrors"] && options.hash_result
+      server_rendered_html = Hash[COMPONENT_HTML_KEY, result["html"]]
```
robwise, line 131: prefer normal hash literal syntax:

```ruby
server_rendered_html = { COMPONENT_HTML_KEY => result["html"] }
```

lib/react_on_rails/react_component/options.rb
```diff
@@ -26,6 +26,10 @@ def dom_id
         @dom_id ||= options.fetch(:id) { generate_unique_dom_id }
       end
 
+      def hash_result
+        retrieve_key(:hash_result)
```
robwise, line 30: Please update the `retrieve_key` implementation as follows:

```ruby
  def retrieve_key(key)
    options.fetch(key) do
      if ReactOnRails.configuration.methods.include?(key)
        ReactOnRails.configuration.public_send(key)
      else
        nil
      end
    end
  end
```

This will allow you to remove the `hash_key` option from the initializer configuration and keep it as only an option you can pass to `react_component` directly. Previously, the assumption had been made that every option passed to `react_component` had a fallback default in the initializer, but in this scenario, that would make no sense and it's unnecessarily expanding the public API of the initializer.

lib/react_on_rails/configuration.rb
```diff
@@ -93,6 +93,7 @@ def self.configuration
       replay_console: true,
       logging_on_server: true,
       raise_on_prerender_error: false,
+      hash_result: false,
```
robwise, line 96: See comment for `react_component/options.rb`, we should remove all of this

Judahmeek:
It was decided that a documentation PR would address the concerns of this PR without the need to add yet another option parameter to `react_component`.

PR #953, opened 2017-10-10
Heroku deployment: db:setup before db:migrate

jki127:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/953)


justin808:
Thanks @jki127 !

PR #961, opened 2017-10-20
Stop FileList from trying to access system root

Judahmeek:
Denis noticed that RoR v10 tried to access his system root (rather than the project root as expected) if `config.node_modules_location` was set to the new default of `""` instead of being set to `"client"`. I was able to reproduce this behavior and this is my fix.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/961)


justin808:
I don't understand why this fix is necessary.

The new default is the rails root directory since v9.

---

Reviewed 5 of 5 files at r1.

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KwsDirm7pb7wXjfNu76:b-w1zvby) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*
> ```Ruby
> 
>     def self.prepend_cd_node_modules_directory(cmd)
>       prefix = ReactOnRails.configuration.node_modules_location
> ```

prefix is not the right variable name

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-:-KwsDmVK-IkADcny0dKX:b82twfo)*

robwise:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/configuration.rb, line 136 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwuNbrolWjlF6Wx0q73:-KwuNbrolWjlF6Wx0q74:b-v5f7no) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/configuration.rb#L136)):*
> ```Ruby
>                    i18n_dir: nil, i18n_yml_dir: nil,
>                    server_render_method: "ExecJS", symlink_non_digested_assets_regex: nil)
>       self.node_modules_location = node_modules_location
> ```

IMO this is where we should be assigning a default in case of a blank string, not inline in the method. Then the keyword arg default could just be `nil` like the others.


```ruby
self.node_modules_location = node_modules_location.present? ? node_modules_location : Rails.root
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

robwise:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 126 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwuOGVXIjqERzhvsYTO:-KwuOGVXIjqERzhvsYTP:b-nyin3g) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L126)):*
> ```Ruby
>     def self.prepend_cd_node_modules_directory(cmd)
>       prefix = ReactOnRails.configuration.node_modules_location
>       return cmd if prefix.blank? || prefix == "."
> ```

given that we've already defaulted the `ReactOnRails.configuration.node_modules_location` to `.` in the case that it's blank inside of `lib/react_on_rails/configuration.rb`, I don't think we really need to make any changes to this file

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

Judahmeek:
@justin808 Because the rails root is NOT currently the default. The system root (i.e. `""`) is. Denis & I both can attest to that.

---

Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/configuration.rb, line 136 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwuNbrolWjlF6Wx0q73:-KwwSiHb4UFn9AjJonf9:b-42xijc) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/configuration.rb#L136)):*

IMO this is where we should be assigning a default in case of a blank string, not inline in the method. Then the keyword arg default could just be `nil` like the others.


```ruby
self.node_modules_location = node_modules_location.present? ? node_modules_location : Rails.root
```
</blockquote></details>

Thanks for this. I was sure there was a better way to do it.

---

*[lib/react_on_rails/utils.rb, line 126 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwuOGVXIjqERzhvsYTO:-KwwSoR0Tq-4exkib6uL:bxc2hpd) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L126)):*

given that we've already defaulted the `ReactOnRails.configuration.node_modules_location` to `.` in the case that it's blank inside of `lib/react_on_rails/configuration.rb`, I don't think we really need to make any changes to this file
</blockquote></details>

Well, given that I made your suggested change in `configuration.rb`, `ReactOnRails.configuration.node_modules_location.blank?` should always be false, correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-:-KwwSVVqF-k8-WoYeZWP:bx3b1si)*

robwise:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 126 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwuOGVXIjqERzhvsYTO:-Kww_Gxbfysca_f5fRmM:b5ajpf0) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L126)):*
> Well, given that I made your suggested change in configuration.rb, ReactOnRails.configuration.node_modules_location.blank? should always be false, correct?

Yep, exactly. So I don't think we need to change this file?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

justin808:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 126 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-Kwy2V4kSXXHe4OCFiH3:-Kwy2V4kSXXHe4OCFiH4:b2c9zel) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L126)):*
> ```Ruby
>     def self.prepend_cd_node_modules_directory(cmd)
>       return cmd if ReactOnRails.configuration.node_modules_location.blank?
>       "cd #{ReactOnRails.configuration.node_modules_location} && #{cmd}"
> ```

@robwise @judahmeek is there a chance that we should always CD to Rails.root unless the node_modules_location is set?

My concern is that the test helper will be invoked from running rspec after CD'ing into the spec directory of the root.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

robwise:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 126 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-Kwy2V4kSXXHe4OCFiH3:-Kx-H07Km5NmvsmU5hnV:b-20c0ml) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L126)):*
> is there a chance that we should always CD to Rails.root unless the node_modules_location is set?

Well, with Judah's changes as of r2, the `node_modules_location` _is_ always set, because if the user doesn't set it, we set it for them.

> My concern is that the test helper will be invoked from running rspec after CD'ing into the spec directory of the root.

I'm not sure I follow this, the test helper will only be invoked if you run a file that requires the `rails_helper`, it shouldn't be based on `pwd`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

justin808:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 126 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-Kwy2V4kSXXHe4OCFiH3:-Kx0HBz3O6GpX_7quh45:bwiumbv) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L126)):*

> is there a chance that we should always CD to Rails.root unless the node_modules_location is set?

Well, with Judah's changes as of r2, the `node_modules_location` _is_ always set, because if the user doesn't set it, we set it for them.

> My concern is that the test helper will be invoked from running rspec after CD'ing into the spec directory of the root.

I'm not sure I follow this, the test helper will only be invoked if you run a file that requires the `rails_helper`, it shouldn't be based on `pwd`?
</blockquote></details>

Guessing PWD might be `Rails.root/spec`, @robwise.

Seems the key thing is the ability to KNOW where the node_modules is set.

Older React on Rails always went to `Rails.root/client` 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

Judahmeek:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KxC3m_7RRJ57nMB6p1Q:b-if6xh6) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*

prefix is not the right variable name
</blockquote></details>

said variable no longer exists

---

*[lib/react_on_rails/utils.rb, line 126 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-Kwy2V4kSXXHe4OCFiH3:-KxC3PWF_Nxh9hG9FK9c:bfvoj5s) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L126)):*

Guessing PWD might be `Rails.root/spec`, @robwise.

Seems the key thing is the ability to KNOW where the node_modules is set.

Older React on Rails always went to `Rails.root/client` 
</blockquote></details>

The default value of `ReactOnRails.configuration.node_modules_location` will now be `Rails.root` unless explicitly set to something else by the user.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

justin808:
see comments @judahmeek 

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/configuration.rb, line 136 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KxC5ehF0oR5vSe0xDvI:-KxC5ehF0oR5vSe0xDvJ:b-gu4m6e) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/configuration.rb#L136)):*
> ```Ruby
>                    i18n_dir: nil, i18n_yml_dir: nil,
>                    server_render_method: "ExecJS", symlink_non_digested_assets_regex: nil)
>       self.node_modules_location = node_modules_location.present? ? node_modules_location : Rails.root
> ```

seems this prevents the value from ever being blank

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KxC5acLizJmHRMAEjWV:b-zb7piu) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*

said variable no longer exists
</blockquote></details>

How can this be blank now? see comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-:-KxC5jYuP4LYI0woXAlF:b-fv188)*

Judahmeek:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/configuration.rb, line 136 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KxC5ehF0oR5vSe0xDvI:-KxC6wQOWunEZJ1FiGuk:bbd8znk) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/configuration.rb#L136)):*

seems this prevents the value from ever being blank
</blockquote></details>

That's exactly what it does.

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KxC6PdCi9zo3sIETUk3:buan28t) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*

How can this be blank now? see comment above.
</blockquote></details>

It can't be. See my discussion (now resolved) with Robwise about that same line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

Judahmeek:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KxC95R9Pt8QHR0Il27y:b-wolooa) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*

It can't be. See my discussion (now resolved) with Robwise about that same line.
</blockquote></details>

To clarify, I pointed out to Rob that I had made this line into dead code and he implied that we should leave it alone (or at least that's the way I interpreted what he said).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

robwise:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 125 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KwsDirlLAELHA4N6igp:-KxEt9wH7I5J66ZkVyPr:btnag7x) ([raw file](https://github.com/shakacode/react_on_rails/blob/433e2cc8a64f956df18545ed5a6feb2df088e82f/lib/react_on_rails/utils.rb#L125)):*

To clarify, I pointed out to Rob that I had made this line into dead code and he implied that we should leave it alone (or at least that's the way I interpreted what he said).
</blockquote></details>

if I did I apologize, this line is no longer needed after making the default in the configuration, which was why I suggested putting i there, so that we can delete this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

justin808:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 125 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KxLI1lGkWHiSinpPFei:-KxLI1lGkWHiSinpPFej:bz72fym) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L125)):*
> ```Ruby
> 
>     def self.prepend_cd_node_modules_directory(cmd)
>       return cmd if ReactOnRails.configuration.node_modules_location.blank?
> ```

I'm confused. Should this line be removed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

robwise:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/react_on_rails/utils.rb, line 125 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-KxLI1lGkWHiSinpPFei:-KxLJu5JUK4JtMJF2UZH:b2nign6) ([raw file](https://github.com/shakacode/react_on_rails/blob/49c06429ee73e59d857c7bfb3a7e77438baca0b1/lib/react_on_rails/utils.rb#L125)):*

I'm confused. Should this line be removed?
</blockquote></details>

yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961)*

justin808:
:lgtm:

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/961#-:-Kx_hu6qp9feMg9iDox6:bnfp4nl)*

PR #968, opened 2017-10-28
Remove unnecessary dependencies from released NPM package

tricknotes:
These dependencies are injected by generator spec.
However they are unnecessary for released react-on-rails package.

Even if the `dependencies` section is removed, it will back after `$ rspec spec/react_on_rails/generators/install_generator_spec.rb` is run.

I think it is unexpected behavior.
So I removed these dependencies and fix specs.

@Judahmeek @justin808
Can you review this PR? It this a correct way?

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/968)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,3 +18,7 @@ rules:
 
   # https://github.com/benmosher/eslint-plugin-import/issues/340
   import/no-extraneous-dependencies: 0
+
+settings:
+  import/core-modules:
+    - react-redux
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Contributors: please follow the recommendations outlined at [keepachangelog.com]
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
+#### Fixed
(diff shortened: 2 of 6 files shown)


justin808:
@tricknotes Can you please rebase on top of master. Then let's see the CI results.

tricknotes:
rebased

tricknotes:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

https://travis-ci.org/shakacode/react_on_rails/jobs/294304640

hmm..., some tests are timeouted :<

tricknotes:
@justin808 
Now all tests are passed on Travis CI.
Could you review this PR?

justin808:
This looks solid to me. @Judahmeek?

Judahmeek:
Nice work. However, I think that `react-redux` should at least be a peer dependency since it is referenced in ReactOnRails source code.

tricknotes:
Thanks for your review, @Judahmeek !
Yes, `react-redux` is referenced by ReactOnRails.
However it seems from specs, generators and docs, not from production code.
Additionally, the ReactOnRails users who use react without redux don't need `react-redux`.

I think it may be a optional dependencies. Or I've misunderstand?

Judahmeek:
You're correct that it is an optional dependency.

PR #978, opened 2017-11-13
Add instructions to convert rails-5 API only app to rails app before …

divya2661:
Setting react-on-rails in rails-5 API only apps

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/978)


justin808:
Thanks @divya2661!

PR #980, opened 2017-11-14
Update v9 references to v10

vfonic:
I also added some minor changes. Let me know if you think those should be changed in any way or are not needed.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/980)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,9 +1,6 @@
 # React on Rails Basic Tutorial
 
-* NOTE: Please be sure to use the BETA or RC versions of React on Rails until 9.0 is released.*
-
-
-This tutorial setups up a new Rails app with **React on Rails**, demonstrating Rails + React + Redux + Server Rendering. It is updated to 9.0.0.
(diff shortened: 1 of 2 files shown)


justin808:
Thanks @vfonic!

docs/tutorial.md
```diff
@@ -52,23 +49,22 @@ rails new test-react-on-rails --webpack=react
...
 bundle exec rails webpacker:install:react
 ```
 
-
-
 Add the **React On Rails** gem to your Gemfile:
 
 ```
-gem 'react_on_rails', '9.0.0'         # use latest gem version, prefer exact version 
+gem 'react_on_rails', '10.0.2'         # prefer exact version
```
6temes, line 62: `# prefer exact gem version to match npm version`

It seems like a good idea to explain the reason.

vfonic: Indeed it does! Thanks! I'll create another PR. I also spotted one more sentence I could improve.

vfonic: https://github.com/shakacode/react_on_rails/pull/981 Clarify tutorial scope; Explain exact gem version

PR #1001, opened 2017-12-23
no camelize react component name

MiguelSavignano:
I do not understand why I can not register a component with the name you want

Example:
I want to register HelloWorld component with the name "hello_world/index"
```js
import ReactOnRails from 'react-on-rails'
import HelloWorld from '../../views/hello_world/index'
ReactOnRails.register({
  "hello_world/index": HelloWorld,
})
```

```html
hello_world/index.html.erb
<%= react_component("hello_world/index", props: @hello_world_props, prerender: false) %>
```

When render the page react on rails say
```
ComponentRegistry.js:65 Uncaught Error: ReactOnRails encountered an error while rendering component: HelloWorld::Index.
Original message: Could not find component registered with name HelloWorld::Index. Registered
```

this error it's because react on rails camelize my component name.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1001)

### DIFF
--- a/lib/react_on_rails/react_component/options.rb
+++ b/lib/react_on_rails/react_component/options.rb
@@ -7,6 +7,8 @@ module ReactOnRails
     class Options
       include Utils::Required
 
+      attr_reader :name
+
       NO_PROPS = {}.freeze
 
       def initialize(name: required("name"), options: required("options"))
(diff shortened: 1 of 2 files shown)


justin808:
Thank you @MiguelSavignano for trying to help.

We're keeping the code as-is because:

1. The JavaScript standard is to use camelCase and not snake_case.
2. This code has been there for over 2 years.
3. Just change your registration. If others want this, I'd happily consider this change.

PR #1009, opened 2018-01-09
Fix rubocop errors:

reconstructions:
* Minimum Ruby version set to 2.1.0


reconstructions:
@justin Looks the build failed because there were some lint issues with the dummy apps that I didn't catch.  Working on that now.

reconstructions:
@justin Looks to me like rubocop is failing in some [build jobs](https://travis-ci.org/shakacode/react_on_rails/builds/326898079?utm_source=github_status&utm_medium=notification), for example the dummy app, because that app selects Ruby 2.3.1 or 2.4.1 instead of 2.2.1 (which the dummy_no_webpack app selected and passed with).  Those three Ruby versions are whitelisted for rvm in [.travis.yml](https://github.com/reconstructions/react_on_rails/blob/master/.travis.yml), but specific versions aren't specified anywhere else, thus different Rubies are getting selected by the different apps, different rubocop rule get applied as a result, which makes the same files fail in some build jobs and pass in others.  Specifying a Ruby version in the Gemfiles might fix this, I will try with 2.2.6 since that is whitelisted in .travis.yml.

justin:
Can you please stop @-ing me. Thanks.

PR #1010, opened 2018-01-09
Fix rubocop errors:

reconstructions:
* Minimum Ruby version advanced to 2.1.0

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1010)


reconstructions:
@justin808 It looks like CI is [failing](https://travis-ci.org/shakacode/react_on_rails/builds/326950070?utm_source=github_status&utm_medium=notification)  because different Ruby versions are being set/excluded for the apps in [.travis.yml matrix:exclude:](https://github.com/shakacode/react_on_rails/blob/5eea919197cef735a9ba649b97d5a7d3a6b7c9a4/.travis.yml#L54) and this is making rubocop throw errors for syntax which passes with other Ruby versions in the other builds.  I assume there is a good reason to use different Ruby versions for the different apps?  I am not sure if you would want me to ensure the same Rubies are being used in the different apps in CI so that Rubocop always passes, or to try to add rubocop configuration so that it is giving special treatment to each dummy app?  Let me know...

justin808:
@reconstructions I'm open to your recommendations. The only criteria is that we try to support older Rails and Ruby versions if convenient.

reconstructions:
It looks like the Travis build matrix is set up really well as it is, and older Rails and Ruby versions would be well supported with a little updating.  So I'll work on adding some rubocop configuration to ensure that the dummy app Ruby is checked and passes with a version 2.3.x target.  

There is probably also some updating that should happen with bugfix versions for both Ruby and rubocop, but first I will try to get everything working.

reconstructions:
@justin808 It looks like the Travis build matrix is set up really well as it is, and older Rails and Ruby versions would be well supported with a little updating. So I'll work on adding some rubocop configuration to ensure that the dummy app Ruby is checked and passes with a version 2.3.x target.

There is probably also some updating that should happen with bugfix versions for both Ruby and rubocop, but first I will try to get everything working.

PR #1012, opened 2018-01-10
Fix Utils.bundle_js_file_path for manifest.json in webpacker projects

### LINKED ISSUE #1011: ReactOnRails::Utils.bundle_js_file_path returns incorrect path for manifest.json
I noticed that `ReactOnRails::TestHelper.ensure_assets_compiled` always reports that `manifest.json` is stale when running tests, even when the file exists and is up to date:

```
Detected the following stale generated files:
  manifest.json
```

After investigating, I believe this is because `ReactOnRails::Utils.bundle_js_file_path()` does not return the correct file path for `'manifest.json'`, so `ReactOnRails::TestHelper.ensure_assets_compiled` can’t accurately check if the file exists and is up to date.

Thanks to [line 89 of `lib/react_on_rails/utils.rb`](https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/utils.rb#L89), it will always return `'manifest.json'` instead of the correct path (e.g. `'/path/to/rails/public/packs-test/manifest.json'`):
```rb
def self.bundle_js_file_path(bundle_name)
  if using_webpacker?
    return bundle_name if bundle_name == "manifest.json"
```

So in summary:

### Problem:

`ReactOnRails::Utils.bundle_js_file_path('manifest.json')` 
returns `'manifest.json'`

### Expected:

`ReactOnRails::Utils.bundle_js_file_path('manifest.json')` 
returns `'/path/to/rails/public/packs-test/manifest.json'`

### Proposed Solution:

Change `ReactOnRails::Utils.bundle_js_file_path` to

```rb
def self.bundle_js_file_path(bundle_name)
  if using_webpacker? && bundle_name != "manifest.json"
    bundle_js_file_path_from_webpacker(bundle_name)
  else
    # Default to the non-hashed name in the specified output directory, which, for legacy
    # React on Rails, this is the output directory picked up by the asset pipeline.
    File.join(ReactOnRails.configuration.generated_assets_dir, bundle_name)
  end
end
```

elstgav:
Utils.bundle_js_file_path was incorrectly returning
`'manifest.json'`
instead of, say,
`'/path/to/rails/public/packs/manifest.json'`

Fixes #1011

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1012)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Changes since last non-beta release.
 *Please add entries here for your pull requests that are not yet released.*
 
 #### Fixed
-- Fixed `Utils.bundle_js_file_path` generating the incorrect path for `manifest.json` in webpacker projects: [#1011] by [elstgav](https://github.com/elstgav)
+- Fixed `Utils.bundle_js_file_path` generating the incorrect path for `manifest.json` in webpacker projects: [Issue #1011](https://github.com/shakacode/react_on_rails/issues/1011) by [elstgav](https://github.com/elstgav)
 
 ### [10.0.2] - 2017-11-10
 #### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+#### Fixed
+- Fixed `Utils.bundle_js_file_path` generating the incorrect path for `manifest.json` in webpacker projects: [#1011] by [elstgav](https://github.com/elstgav)
+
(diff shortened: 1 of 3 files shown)


justin808:
:lgtm:

Very nice work. @elstgav Please email me at justin@shakacode.com. I'd like to invite you to our slack discussion room.

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1012#-:-L2bJcvi1P3sMRM4qmoH:bmcrrb)*

Issue #965, opened 2017-10-27
Deprecation Warnings


robwise:
Some new stuff has come out and we've bumped our internal project's deps to the point where Yarn is complaining about version ranges defined by RoR:

```
warning "react-on-rails@10.0.0" has unmet peer dependency "babel-runtime@>= 6".
warning "react-on-rails@10.0.0" has unmet peer dependency "js-yaml@>= 3.0.0".
```

This all seems to work fine so we can just loosen the dependency requirements to allow these version ranges.

Also, this happens on every page where RoR is used when using React v16:

```
Warning: render(): Calling ReactDOM.render() to hydrate server-rendered markup will stop working in React v17. Replace the ReactDOM.render() call with ReactDOM.hydrate() if you want React to attach to the server HTML.
```

justin808:
@Judahmeek Maybe you can fix this and let's ship 10.0.1? or 10.1.0?

jm96441n:
So I see this issue is closed, but after updating to the most recent release I'm still getting similar deprecation warnings when running yarn commands:
```
warning "@rails/webpacker > postcss-cssnext@3.1.0" has unmet peer dependency "caniuse-lite@^1.0.30000697".
warning " > react-on-rails@10.1.1" has unmet peer dependency "babel-runtime@>= 6".
warning " > react-on-rails@10.1.1" has unmet peer dependency "js-yaml@>= 3.0.0".
warning " > webpack-dev-server@2.11.2" has unmet peer dependency "webpack@^2.2.0 || ^3.0.0".
warning "webpack-dev-server > webpack-dev-middleware@1.12.2" has unmet peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0"
```

VincentLanglet:
The warning is still here

PR #1013, opened 2018-01-11
Use ReactDOM.hydrate() for hydrating a server-side component with Rea…

elstgav:
Fixes the `ReactDOM.render()` deprecation warning for React v16 mentioned in #965 

Calling `ReactDOM.render()` on a server-rendered component with React v16 will produce the following deprecation warning:

```
Warning: render(): Calling ReactDOM.render() to hydrate server-rendered markup will stop working in React v17. Replace the ReactDOM.render() call with ReactDOM.hydrate() if you want React to attach to the server HTML.
```

### TODO
- [ ] I’m unsure of where a test for this should go. Do we want a test? And if so, where does that go?
- [ ] Should `ReactOnRails.render()` also be updated? I’m less clear on if it’s used for server-rendered components

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1013)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+#### Fixed
+- Fixed `ReactDOM.render()` deprecation warning with React v16: [Issue #965](https://github.com/shakacode/react_on_rails/issues/965) by [elstgav](https://github.com/elstgav)
+
 ### [10.0.2] - 2017-11-10
 #### Fixed
 - Remove unnecessary dependencies from released NPM package: [PR 968](https://github.com/shakacode/react_on_rails/pull/968) by [tricknotes](https://github.com/tricknotes).
--- a/node_package/src/clientStartup.js
+++ b/node_package/src/clientStartup.js
@@ -95,8 +95,8 @@ function domNodeIdForEl(el) {
 }
 
 /**
- * Used for client rendering by ReactOnRails. Either calls ReactDOM.render or delegates
- * to a renderer registered by the user.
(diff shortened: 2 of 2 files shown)


justin808:
@elstgav Thanks! Do you have anybody using v16 that can help review this change?

In terms of "Should ReactOnRails.render() also be updated? I’m less clear on if it’s used for server-rendered components", it's a simple helper designed to be used from your JS code if you don't want to use the Rails Helper. Thus, it's not used for server rendering. In that case, would it require any updates?

In terms of tests, I'm open to your suggestions. This might be difficult as it's dependent on the React version used. I'm open to updating the code to the latest version of React so that we can test this properly.

```js
 /**
   * ReactOnRails.render("HelloWorldApp", {name: "Stranger"}, 'app');
   *
   * Does this:
   *   ReactDOM.render(React.createElement(HelloWorldApp, {name: "Stranger"}),
   *     document.getElementById('app'))
   *
   * @param name Name of your registered component
   * @param props Props to pass to your component
   * @param domNodeId
   * @returns {virtualDomElement} Reference to your component's backing instance
   */
  render(name, props, domNodeId) {
    const componentObj = ComponentRegistry.get(name);
    const reactElement = createReactElement({ componentObj, props, domNodeId });

    // eslint-disable-next-line react/no-render-return-value
    return ReactDOM.render(reactElement, document.getElementById(domNodeId));
  },
```

justin808:
@elstgav can you please rebase on master? I merged your other PR.

elstgav:
Shoot. I rebased and force-pushed, but that closed this PR 😕

elstgav:
And to answer your question, we’re using React 16 at my job. I haven’t tested the functionality yet, I got a bit bogged down with how to set up the repo

PR #1016, opened 2018-01-18
Fix rubocop errors

reconstructions:
* Minimum Ruby version advanced to 2.1.0
  * docker-compose.yml checked into git

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1016)


reconstructions:
@justin808 There were rubocop errors which were thrown when the Travis build matrix was using certain Ruby versions, but not others.  Since all of these errors were places where rubocop complained that its own disabling was unnecessary (which was only true in certain Ruby versions), [I took the approach of disabling those errors as well](https://github.com/reconstructions/react_on_rails/blob/64829bf12f1e6eef976f25d2044b75e238a22ec6/lib/generators/react_on_rails/base_generator.rb#L58).

justin808:
@reconstructions is this ready to go?

justin808:
@reconstructions Looks good. Can you please rebase on top of master. We should be good to go.

justin808:
@reconstructions I updated the branch on the github UX. If tests pass, I'll merge.

reconstructions:
@justin808 thx, I just pushed this rebased on master, tests still green.

justin808:
Thanks @reconstructions

PR #1017, opened 2018-01-22
Updated HelloWorld.jsx

AnkurVyas-BTC:
After the new facebook docs update the link to set initial state in ES6 Class has changed to https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1017)


justin808:
Thanks @AnkurVyas-BTC!

PR #1020, opened 2018-01-24
Fix server rendering rescue

jblasco3:
Latest release does not display the actual error when server-rendering fails.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1020)


justin808:
Thanks @jblasco3!

PR #1028, opened 2018-02-09
Use ReactDOM.hydrate() for hydrating a SSR component if available

theJoeBiz:
Looks like this was originally addressed in #1013, but was closed by accident and never reopened. My project is using React 16 and I can confirm that this patch prevents the deprecation warning.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1028)

### DIFF
--- a/node_package/src/clientStartup.js
+++ b/node_package/src/clientStartup.js
@@ -95,8 +95,8 @@ function domNodeIdForEl(el) {
 }
 
 /**
- * Used for client rendering by ReactOnRails. Either calls ReactDOM.render or delegates
- * to a renderer registered by the user.
+ * Used for client rendering by ReactOnRails. Either calls ReactDOM.hydrate, ReactDOM.render, or
+ * delegates to a renderer registered by the user.
  * @param el
  */
 function render(el, railsContext) {
@@ -115,18 +115,24 @@ function render(el, railsContext) {
         return;
       }
 
+      // Hydrate if available and was server rendered
+      const shouldHydrate = !!ReactDOM.hydrate && !!domNode.innerHTML;
+
       const reactElementOrRouterResult = createReactElement({
         componentObj,
         props,
         domNodeId,
         trace,
         railsContext,
+        shouldHydrate,
       });
 
       if (isRouterResult(reactElementOrRouterResult)) {
         throw new Error(`\
 You returned a server side type of react-router error: ${JSON.stringify(reactElementOrRouterResult)}
 You should return a React.Component always for the client side entry point.`);
+      } else if (shouldHydrate) {
+        ReactDOM.hydrate(reactElementOrRouterResult, domNode);
       } else {
         ReactDOM.render(reactElementOrRouterResult, domNode);
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Contributors: please follow the recommendations outlined at [keepachangelog.com]
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
+#### Fixed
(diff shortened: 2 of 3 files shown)


justin808:
@theJoeBiz Any way to test or verify this for both older React and new react?

theJoeBiz:
@justin808 I don't think so. There aren't any existing tests for `clientStartup.js` and I've never had to deal with tests for two separate versions of react. That being said, the change made here is fairly transparent and visibly falls through to the old behavior for any version (older than 16) where `ReactDOM.hydrate` is undefined.

theJoeBiz:
Also, only one test needs to be updated for a react 16 upgrade
https://github.com/theJoeBiz/react_on_rails/commit/468338c1ab0964e872efe45d5cd4e994b1515ee8#diff-85a5145721dbc772cfb6de2e6102717c

justin808:
@alexfedoseev can you please confirm this React syntax looks good.

node_package/src/clientStartup.js
```diff
@@ -128,7 +128,17 @@ function render(el, railsContext) {
 You returned a server side type of react-router error: ${JSON.stringify(reactElementOrRouterResult)}
 You should return a React.Component always for the client side entry point.`);
       } else {
-        ReactDOM.render(reactElementOrRouterResult, domNode);
+        const shouldHydrate = !!(
+          ReactDOM.hydrate &&
+          domNode &&
+          domNode.hasAttribute('data-reactroot')
+        );
```
alex35mil: Looks like you mix things here you shouldn't mix. E.g. `!!ReactDOM.hydrate` is enough for `shouldHydrate` decision, but if something is wrong w/ dom node then it's out of scope `ReactDOM.hydrate` vs `ReactDOM.render` problem.

Also, https://github.com/facebook/react/issues/10971

theJoeBiz:
@alexfedoseev Good call, updated to just check for truthiness of ReactDOM.hydrate.

justin808:
@alexfedoseev let me know if I should merge.

theJoeBiz:
We just moved out app into production with this warning, which is fine, but I was just revisiting this on a page that has a combination of server/client rendered components and found a flaw. Turns out, the `data-reactroot` check I removed in a previous iteration of this was there to check to see if the component was server rendered. In react 16, `ReactDOM.hydrate` is only supposed to be used on components that were previously rendered on the server. So, now that `data-reactroot` is unreliable, I replaced that check with `!!domNode.innerHTML` so the hydrate check is now:

```js
const shouldHydrate = !!ReactDOM.hydrate && !!domNode.innerHTML;
```

I tested this most recent commit locally on a page using React 16 and a combination of server/client rendered components. Here is the output from the console:

```
(index):68 [SERVER] RENDERED SearchForm to dom node with id: SearchForm-react-component with railsContext: {...,"serverSide":true}
(index):98 [SERVER] RENDERED Explore to dom node with id: Explore-react-component with railsContext: {...,"serverSide":true}
createReactElement.js:41 HYDRATED SearchForm in dom node with id: SearchForm-react-component using props, railsContext: {} {…,"serverSide":false}
createReactElement.js:41 HYDRATED Explore in dom node with id: Explore-react-component using props, railsContext: {} {…,"serverSide":false}
createReactElement.js:43 RENDERED FreeReport to dom node with id: FreeReport-react-component with props, railsContext: {} {…,"serverSide":false}
createReactElement.js:43 RENDERED ClientOnlyTest to dom node with id: ClientOnlyTest-react-component with props, railsContext: {} {…,"serverSide":false}
```

justin808:
@alexfedoseev Please look at the latest change. I know we want to use this as well.

justin808:
@theJoeBiz can you please rebase on top of master and add an entry to the CHANGELOG.md

theJoeBiz:
@justin808 added an entry to CHANGELOG.md and squashed the commits

alexfedoseev:
LGTM

justin808:
Shipped! 10.1.2

theJoeBiz:
Any estimate on getting this onto npm as 10.1.2?

PR #1043, opened 2018-03-08
Support multi-bundle component registry

ruswerner:
This change moves the `registeredComponents` onto the global context so when your project is using multiple webpack bundles, and thus calling `ReactOnRails.registerComponents()` multiple times prior to rendering, the components are accumulated in the same `Map` instance. Without this change, the last webpack bundle to register components overwrites all the previous registrations. I'm open to suggestions to solve this problem a different way. I've searched through the docs _and_ the code and this is the only way I got it to work. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1043)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+- ComponentRegistry now uses a Map on the global context so multiple webpack bundles can call `ReactOnRails.registerComponents()` multiple times prior to rendering.
+
 ### [10.1.3] - 2018-02-28
 #### Fixed
 - Improved error reporting on version mismatches between Javascript and Ruby packages. [PR 1025](https://github.com/shakacode/react_on_rails/pull/1025) by [theJoeBiz](https://github.com/squadette).
--- a/node_package/src/ComponentRegistry.js
+++ b/node_package/src/ComponentRegistry.js
@@ -5,11 +5,10 @@ import context from './context';
 
 const ctx = context();
 
-let registeredComponents;
-if(!ctx.reactOnRailsRegisteredComponents) {
+if (!ctx.reactOnRailsRegisteredComponents) {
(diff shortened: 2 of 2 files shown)


ruswerner:
Kia ora @justin808

I just discovered that the same thing is happening for Store registration as well. A simplified example is this...

We have 2 webpack bundles: `common.js` and `landingpage.js`. Some site-wide components and the global `appStore` are registered in `common.js`. Then some more components are registered in `landingpage.js` for just that page. When the page is loaded and everything is processed, the components registered in the `landingpage.js` bundle cannot find the hydrated store because the instance is stored in the _module-scope_ `const hydratedStores = new Map();` within the `common.js` bundle.

This situation effectively prevents using multiple webpack bundles on a site. What is your approach when doing this sort of thing? Or did I miss something in the docs? Do you register things multiple times (i.e. in every bundle)?

Many thanks!

justin808:
@ruswerner Please put some debugging here: https://github.com/shakacode/react_on_rails/blob/master/node_package/src/StoreRegistry.js

and let us know what you find.

I suspect there is something wrong with the way you are setting up your webpack bundles. If you have totally different webpack builds, I'm pretty sure that Webpack will be doing some closure stuff that will affect the results.

ruswerner:
@justin808 Yes, that is exactly what is going on. There are multiple instances of `hydratedStores`; one for each webpack bundle. They are defined at the top of `StoreRegistry.js` so they are scoped to that module, and webpack is including this module in each bundle, so there are multiple. I'll start digging around in the webpack docs to see if it can somehow "share" this module across bundles. Maybe the `CommonsChunkPlugin` can help.

Do you always build a singular client bundle for your production sites?

ruswerner:
@justin808 I was able to get it working by ensuring the `react-on-rails` package existing in a shared chunk via `CommonsChunkPlugin`. If this isn't done, and you are using multiple bundles with webpack, then you'll run into the problems I had. Thanks for the help and hopefully others can benefit from this PR discussion.

PR #1047, opened 2018-03-14
Place the rendered output before the spec tag

jeffnappi:
Many folks using the SSR features of react-on-rails are doing so in order to improve SEO while simultaneously improving user experience.

It is commonly recommended to have key content rendered into the HTML at the earliest possible point. From what I understand, there is no need for the specification tag/script to come before the rendered content and simply swapping the order can have a significant impact on the content that fits within the 100kb "SEO optimized" page size.

e.g. if I have 50kb of state data followed by 100kb of the actual rendered HTML, the bottom half of the page may not get indexed by e.g. Google.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1047)

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -305,9 +305,9 @@ module ReactOnRailsHelper
     # IMPORTANT: Ensure that we mark string as html_safe to avoid escaping.
     # rubocop:disable Layout/IndentHeredoc
     <<-HTML.html_safe
+#{rendered_output}
 #{component_specification_tag}
-    #{rendered_output}
-    #{console_script}
+#{console_script}
     HTML
     # rubocop:enable Layout/IndentHeredoc
   end


jeffnappi:
Opening this up for discussion on the topic. This change is particularly valuable for those commonly using a single root component.

justin808:
@alexfedoseev @robwise Any opinion on this one?

@jeffnappi You seem to have a good point. Any further research or examples to support this change?

robwise:
@justin808 Seems like a free win if this is accurate 👍

jeffnappi:
@justin808 This came to me via a veteran SEO industry leader. Some of the beliefs they hold are simply based on their prior experience and it can be difficult to craft scientific experiments to test the behavior of search engine black-box algorithms. To me there's some logic that the order of appearance of important text within a document may have impact on its ranking across various platforms.

From my research, Google now supports something like 15MB documents. This would still apply in that case, but of course that's not common.

I would be happy if this got merged, it seems like a free change with no negative side-effect. And SEO's viewing source on pages generated with it will be happy :smile:

justin808:
Thanks @jeffnappi

PR #1048, opened 2018-03-20
N/A

mapreal19:
N/A

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1048)

### DIFF
--- a/app/helpers/react_on_rails_helper.rb
+++ b/app/helpers/react_on_rails_helper.rb
@@ -101,26 +101,43 @@ module ReactOnRailsHelper
     server_rendered_html = internal_result["result"]["html"]
     console_script = internal_result["result"]["consoleReplayScript"]
 
-    if server_rendered_html.is_a?(String)
-      build_react_component_result_for_server_rendered_string(
-        server_rendered_html: server_rendered_html,
-        component_specification_tag: internal_result["tag"],
-        console_script: console_script,
-        options: internal_result["options"]
-      )
-    elsif server_rendered_html.is_a?(Hash)
-      puts "[DEPRECATION] ReactOnRails: Use react_component_hash to return a Hash to your ruby view code"
-      build_react_component_result_for_server_rendered_hash(
-        server_rendered_html: server_rendered_html,
-        component_specification_tag: internal_result["tag"],
-        console_script: console_script,
-        options: internal_result["options"]
-      )
-    else
-      raise "server_rendered_html is expected to be a String. If you're trying to use a generator function to
+    with_caching(component_name, internal_result["options"]) do
+      if server_rendered_html.is_a?(String)
+        build_react_component_result_for_server_rendered_string(
+          server_rendered_html: server_rendered_html,
+          component_specification_tag: internal_result["tag"],
+          console_script: console_script,
+          options: internal_result["options"]
+        )
+      elsif server_rendered_html.is_a?(Hash)
+        puts "[DEPRECATION] ReactOnRails: Use react_component_hash to return a Hash to your ruby view code"
+        build_react_component_result_for_server_rendered_hash(
+          server_rendered_html: server_rendered_html,
+          component_specification_tag: internal_result["tag"],
+          console_script: console_script,
+          options: internal_result["options"]
+        )
+      else
+        raise "server_rendered_html is expected to be a String. If you're trying to use a generator function to
       return a Hash to your ruby view code, then use react_component_hash instead of react_component and
       see https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/client/app/startup/ReactHelmetServerApp.jsx
       for an example of the necessary javascript configuration."
+      end
+    end
+  end
+
+  def with_caching(component_name, options, &block)
+    return yield unless options.cached
+
+    props_digest = Digest::MD5.hexdigest(props.to_s)
+    cache_key = "react_on_rails/#{component_name}/props-#{props_digest}"
+
+    if options.prerender
+      cache_key += "/server_bundle-#{ReactOnRails::Utils.server_bundle_file_hash}"
+    end
+
+    Rails.cache.fetch(cache_key) do
+      yield block
     end
   end
 
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -87,6 +87,7 @@ module ReactOnRails
       # generated_assets_dirs is deprecated
       generated_assets_dir: "",
       server_bundle_js_file: "",
+      cached: false,
       prerender: false,
       replay_console: true,
       logging_on_server: true,
@@ -108,7 +109,7 @@ module ReactOnRails
 
   class Configuration
     attr_accessor :node_modules_location, :server_bundle_js_file, :prerender, :replay_console,
-                  :trace, :development_mode,
+                  :trace, :development_mode, :cached,
                   :logging_on_server, :server_renderer_pool_size,
                   :server_renderer_timeout, :skip_display_none, :raise_on_prerender_error,
                   :generated_assets_dirs, :generated_assets_dir,
@@ -119,6 +120,7 @@ module ReactOnRails
 
     def initialize(node_modules_location: nil, server_bundle_js_file: nil, prerender: nil,
                    replay_console: nil,
+                   cached: nil,
                    trace: nil, development_mode: nil,
                    logging_on_server: nil, server_renderer_pool_size: nil,
                    server_renderer_timeout: nil, raise_on_prerender_error: true,
@@ -145,6 +147,7 @@ module ReactOnRails
                               else
                                 development_mode
                               end
+      self.cached = cached
       self.trace = trace.nil? ? Rails.env.development? : trace
       self.raise_on_prerender_error = raise_on_prerender_error
       self.skip_display_none = skip_display_none
--- a/lib/react_on_rails/react_component/options.rb
+++ b/lib/react_on_rails/react_component/options.rb
@@ -34,6 +34,10 @@ module ReactOnRails
         retrieve_key(:prerender)
       end
 
+      def cached
+        retrieve_key(:cached)
+      end
+
       def trace
         retrieve_key(:trace)
       end
--- a/lib/react_on_rails/utils.rb
+++ b/lib/react_on_rails/utils.rb
@@ -20,20 +20,28 @@ module ReactOnRails
       pathname.basename.to_s
     end
 
-    # Returns the hashed file name of the server bundle when using webpacker.
-    # Nececessary fragment-caching keys.
     def self.server_bundle_file_name
-      return @server_bundle_hash if @server_bundle_hash && !Rails.env.development?
+      File.join(
+        ReactOnRails.configuration.generated_assets_dir,
+        ReactOnRails.configuration.server_bundle_js_file
+      )
+    end
 
-      @server_bundle_hash = begin
-        server_bundle_name = ReactOnRails.configuration.server_bundle_js_file
-        bundle_file_name(server_bundle_name)
+    def self.server_bundle_file_hash
+      if Rails.env.development?
+        calculate_server_bundle_file_hash
+      else
+        @server_bundle_file_hash ||= calculate_server_bundle_file_hash
       end
     end
 
     ###########################################################
     # PRIVATE API -- Subject to change
     ###########################################################
+
+    def self.calculate_server_bundle_file_hash
+      Digest::MD5.file(server_bundle_file_name)
+    end
 
     # https://forum.shakacode.com/t/yak-of-the-week-ruby-2-4-pathname-empty-changed-to-look-at-file-size/901
     # return object if truthy, else return nil
--- a/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
+++ b/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
@@ -163,6 +163,44 @@ describe ReactOnRailsHelper, type: :helper do
         expect(result).not_to match(/data-trace=/)
       end
     end
+
+    describe "caching", :caching do
+      context "with 'cached' == true" do
+        it "caches the content" do
+          react_component("App", cached: true)
+
+          expect(cache_data.keys).to include(%r{react_on_rails/App/})
+        end
+
+        context "with props" do
+          # INFO: Rails uses Digest::MD5 too
+          it "includes props digest in the cache key" do
+            props = { name: "My Test Name" }
+            digest = Digest::MD5.hexdigest(props.to_s)
+            react_component("App", cached: true, props: props)
+
+            expect(cache_data.keys).to include("react_on_rails/App/props-#{digest}")
+          end
+        end
+
+        context "with 'prerender' == true" do
+          it "includes server bundle hash in the cache key" do
+            react_component("App", cached: true, prerender: true)
+
+            expected_key = /server_bundle-#{ReactOnRails::Utils.server_bundle_file_hash}/
+            expect(cache_data.keys).to include(expected_key)
+          end
+        end
+      end
+
+      context "with 'cached' == false" do
+        it "doesn't caches the content" do
+          react_component("App", cached: false)
+
+          expect(cache_data.keys).to be_empty
+        end
+      end
+    end
   end
 
   describe "#redux_store" do
--- a/spec/dummy/spec/spec_helper.rb
+++ b/spec/dummy/spec/spec_helper.rb
@@ -56,8 +56,8 @@ RSpec.configure do |config|
   #   # to individual examples or groups you care about by tagging them with
   #   # `:focus` metadata. When nothing is tagged with `:focus`, all examples
   #   # get run.
-  #   config.filter_run :focus
-  #   config.run_all_when_everything_filtered = true
+  config.filter_run :focus
+  config.run_all_when_everything_filtered = true
   #
   #   # Allows RSpec to persist some state between runs in order to support
   #   # the `--only-failures` and `--next-failure` CLI options. We recommend
--- a/spec/dummy/spec/support/caching.rb
+++ b/spec/dummy/spec/support/caching.rb
@@ -1,0 +1,26 @@
+module CachingHelpers
+  # INFO: helper for accessing key/value pairs included in the cache
+  #
+  #    cache_data #=> {"jbuilder/views/users/1-2018...": <user_partial content>}
+  def cache_data
+    Rails.cache.instance_variable_get(:@data)
+  end
+end
+
+RSpec.configure do |config|
+  config.include CachingHelpers
+
+  config.before(:each, :caching) do
+    cache_store = ActiveSupport::Cache::MemoryStore.new
+    allow(controller).to receive(:cache_store).and_return(cache_store)
+    allow(::Rails).to receive(:cache).and_return(cache_store)
+    Rails.cache.clear
+  end
+
+  config.around(:each, :caching) do |example|
+    caching = ActionController::Base.perform_caching
+    ActionController::Base.perform_caching = true
+    example.run
+    ActionController::Base.perform_caching = caching
+  end
+end


dzirtusss:
@mapreal19 nice job so far

dzirtusss:
@mapreal19 couple of comments / things to check. 

---

Reviewed 5 of 7 files at r1, 8 of 8 files at r2.
Review status: 12 of 13 files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[CHANGELOG.md, line 12 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L87-hhwHxs4TZteRKSQ:-L87-hhxdY86UmexPMxS:b-wr0q50) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/CHANGELOG.md#L12)):*
> ```Markdown
> 
> #### Added
>  - Added `cached` option for `react_component` helper. [PR 1048](https://github.com/shakacode/react_on_rails/pull/1048)
> ```

guess you need to add `by ...` here

---

*[README.md, line 560 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L87231tTrkxQgtQT8k6:-L87231tTrkxQgtQT8k7:b1pgqj) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/README.md#L560)):*
> ```Markdown
>   + **props:** Ruby Hash which contains the properties to pass to the react object, or a JSON string. If you pass a string, we'll escape it for you.
>   + **prerender:** enable server-side rendering of a component. Set to false when debugging!
>   + **cached:** set to true if you want to cache the component.
> ```

I guess if we will not add some component digest, we need to say at least that name should be unique app-wise.

As well I guess we need to say that bundle filename may be without digest since we digest the content?

---

*[docs/additional-reading/caching-and-performance.md, line 6 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L872tymRpgM0k4FfIie:-L872tymRpgM0k4FfIif:b-brxr8o) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/docs/additional-reading/caching-and-performance.md#L6)):*
> ```Markdown
> ## Caching
> 
> If you want to cache your component, pass the `cached` option:
> ```

Maybe we should describe cache key composition like it was done in the deleted sentence. Meaning that the user understands what affects/invalidates the cache.

---

*[lib/react_on_rails/react_component/cache.rb, line 17 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L871QCkk5znM4u08D3U:-L871QCkk5znM4u08D3V:b-73o2oy) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L17)):*
> ```Ruby
> 
>         def cache_key(component_name, options)
>           result = "react_on_rails/#{component_name}/#{props_digest(options)}"
> ```

Don't we need to add some component/bundle digest here? Meaning not for server rendering mode also.

---

*[lib/react_on_rails/react_component/cache.rb, line 28 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L875NAcENVo0BoFQbKl:-L875NAcENVo0BoFQbKm:bdqhz9j) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L28)):*
> ```Ruby
>         def server_bundle_digest
>           "server_bundle-#{ReactOnRails::Utils.server_bundle_file_hash}"
>         end
> ```

Should we memoize it somehow? As I understand this one will be called with every cache check.

I mean when we had `server-bundle.123412341234.js` hash check is fast because it is only the filename, and don't need read whole big file.

---

*[spec/dummy/spec/spec_helper.rb, line 60 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L873zJjWsthNqDPo_ki:-L873zJjWsthNqDPo_kj:b-ybnd3p) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/spec/dummy/spec/spec_helper.rb#L60)):*
> ```Ruby
>   #   # get run.
>   config.filter_run :focus
>   config.run_all_when_everything_filtered = true
> ```

I guess better leave it like it was that CI will always run.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-:-L8769A4DfTKB1isxrLr:b-j7u6jv)*

mapreal19:
Review status: 9 of 16 files reviewed at latest revision, 7 unresolved discussions.

---

*[lib/react_on_rails/react_component/cache.rb, line 17 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L871QCkk5znM4u08D3U:-L87CQN2z8L7wD7vCNsG:b-g1qgw8) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L17)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

Don't we need to add some component/bundle digest here? Meaning not for server rendering mode also.
</blockquote></details>

For the client you mean? I'm not sure. @robwise may know?

---

*[lib/react_on_rails/react_component/cache.rb, line 28 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L875NAcENVo0BoFQbKl:-L87C9szb1mU1JQxhJvD:be7p4mp) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L28)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

Should we memoize it somehow? As I understand this one will be called with every cache check.

I mean when we had `server-bundle.123412341234.js` hash check is fast because it is only the filename, and don't need read whole big file.
</blockquote></details>

If you check in the implementation we memoize that in `server_bundle_file_hash`

---

*[spec/dummy/spec/spec_helper.rb, line 60 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L873zJjWsthNqDPo_ki:-L87ChwkhXLt0QcSwYad:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/spec/dummy/spec/spec_helper.rb#L60)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

I guess better leave it like it was that CI will always run.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

justin808:
LOTS OF COMMENTS

@dzirtusss @mapreal19 Please confirm that caching only makes sense if server rendering. The HTML generated for non server rendering is tiny. The props is the same script tag. However, if we have to convert the props to a string from the hash is already done, and then if we convert the string to a digest, just to get back the string, then that doesn't make much sense.

Overall, I have a concern that we're just caching the HTML from the server rendering, and possibly the big amount of time is in creating the props, so this may be minimal at this level, and we should really have the computation of the props short-circuited before even getting to this level of the code.

---

Reviewed 5 of 7 files at r1, 4 of 8 files at r2, 7 of 7 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[CHANGELOG.md, line 12 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L87-hhwHxs4TZteRKSQ:-L89IU6avjEeSwepsWRq:b9ajgge) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/CHANGELOG.md#L12)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

guess you need to add `by ...` here
</blockquote></details>

see other entries

---

*[README.md, line 560 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L87231tTrkxQgtQT8k6:-L89I7-l6EFUqFP1qR-5:b-jj88js) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/README.md#L560)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

I guess if we will not add some component digest, we need to say at least that name should be unique app-wise.

As well I guess we need to say that bundle filename may be without digest since we digest the content?
</blockquote></details>

This is not clear: 

- Caching only applies if pre-rendering (check the code).
- Link to the cache page that shows the equivalent if doing this manually.

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89IXIoBJpDBS5wy_6Y:-L89IXIoBJpDBS5wy_6Z:b-65f3ug) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L88)):*
> ```Ruby
>   #      not pass any props if you are separately initializing the store by the `redux_store` helper.
>   #   prerender: <true/false> set to false when debugging!
>   #   cached: <true/false> set to true if you want to cache
> ```

when prerender, etc

---

*[app/helpers/react_on_rails_helper.rb, line 101 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89L64Fruvd-t2SBjYD:-L89L64Fruvd-t2SBjYE:b-9uhaw4) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L101)):*
> ```Ruby
>   # Any other options are passed to the content tag, including the id.
>   def react_component(component_name, raw_options = {})
>     internal_result = internal_react_component(component_name, raw_options)
> ```

@mapreal19 this is where all the work is done to create the server rendering, so the cache part seems to be in the wrong place.

https://github.com/shakacode/react_on_rails/blob/master/app/helpers/react_on_rails_helper.rb#L331

and this is the main part of the code to avoid:
https://github.com/shakacode/react_on_rails/blob/master/app/helpers/react_on_rails_helper.rb#L377


Overall, we need to not get into ANY of the react on rails code **IF** cached and simply return the value from the cache.

We need to think at a high level on what the point of this fix really is?

Is it just to avoid JS execution of some JS code with the props?

And what are the costs already incurred to get to this point, including extra costs EVERY time:

* Calculation of the props with DB queries
* Conversion of a hash of props to a string
* Conversion of the string to a digest
* lookup in cache

---

*[app/helpers/react_on_rails_helper.rb, line 107 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89ImqpYVyU7t0YyTRn:-L89ImqpYVyU7t0YyTRo:b-odyvma) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L107)):*
> ```Ruby
> 
>     ReactOnRails::ReactComponent::Cache.cache_if_flagged(component_name, request, options) do
>       if server_rendered_html.is_a?(String)
> ```

caching only applies if server rendered?

---

*[docs/additional-reading/caching-and-performance.md, line 6 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L872tymRpgM0k4FfIie:-L89MMThthyt8lIoLmxy:bozi4sm) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/docs/additional-reading/caching-and-performance.md#L6)):*
<details><summary><i>Previously, dzirtusss (Sergey Tarasov) wrote…</i></summary><blockquote>

Maybe we should describe cache key composition like it was done in the deleted sentence. Meaning that the user understands what affects/invalidates the cache.
</blockquote></details>

See my other comments.

In most cases, it's **better** to put the cache block outside of the calculation of the calculation of the props. 

So 100% explain how to do that...

I'm starting to think this is like the difference of calling:


```ruby
Rails.logger.info("BLAH #{some_super_time_consuming_crap}")
```

vs. **Using a BLOCK**
```ruby
Rails.logger.info { "BLAH #{some_super_time_consuming_crap}" }
```

So the block is never executed if the log level is not at info or debug.

---

*[lib/react_on_rails/utils.rb, line 23 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L84-RkGTw1ezLtd-a7R:-L89N6cHwmI3ipu27jcg:bry4icb) ([raw file](https://github.com/shakacode/react_on_rails/blob/b6bf5969d9f4832a7b9c1ae4af1a21c4bd982f7a/lib/react_on_rails/utils.rb#L23)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

this is a workaround for getting the server bundle. We got the manifiest being overwritten by the client bundle cc @robwise @justin808 
</blockquote></details>

I'm totally not clear how this relates to the rest of the fix.

just refactoring? or b/c server bundle hash _might_ not be using a sha in the file name?

and should we account for this, given that we're going to run this every time in development so if the sha is in the file name, computing an unnecessary digest from the HUGE JS file (React source etc.) is unnecessary.

---

*[lib/react_on_rails/react_component/cache.rb, line 17 at r2](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L871QCkk5znM4u08D3U:-L89OYaF89vz3IT3nsDT:bbk8ydj) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L17)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

For the client you mean? I'm not sure. @robwise may know?
</blockquote></details>

@dzirtusss server_bundle_digest handles that part pl.us the component name. There's only one file for server rendering.

---

*[lib/react_on_rails/react_component/cache.rb, line 12 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OOM324BTAejtBgfa:-L89OOM324BTAejtBgfb:b-ua8202) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L12)):*
> ```Ruby
>       class << self
>         def cache_if_flagged(component_name, request, options, &block)
>           return yield block unless options.cached
> ```

I think this needs to be at a different level so **NONE** of the RoR code runs if we're caching. This is called after ALL the work is already done.

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L89OAk_hagGS33UEWBl:b245c4i) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*
> ```Ruby
>                      "mobile"
>                    elsif browser.device.tablet?
>                      "tablet"
> ```

How do we know that device.mobile? does not return true for a tablet? I recently saw a bug where a tablet was displaying mobile layout.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L89NwLGGA2Ym6Ft3tdW:bdv6snt) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*
> ```Ruby
> 
>         def props_digest(options)
>           "props-#{Digest::MD5.hexdigest(options.props.to_s)}"
> ```

if you profile this stuff, you'd see that if we're **EVER** converting a giant hash to a string more than once, this code is NOT helping with performance.

---

*[spec/react_on_rails/spec_helper.rb, line 97 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OnbIKwp_KXcfIVvA:-L89OnbIKwp_KXcfIVvB:b-o8shoh) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/spec/react_on_rails/spec_helper.rb#L97)):*
> ```Ruby
>   #   # get run.
>   # config.filter_run :focus
>   # config.run_all_when_everything_filtered = true
> ```

what's going on here?

---

*[spec/react_on_rails/utils_spec.rb, line 38 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OwWPHlqUHF0NUFXd:-L89OwWPHlqUHF0NUFXe:b-y1gju8) ([raw file](https://github.com/shakacode/react_on_rails/blob/a6fc8d8ff1e37ab434777ff9c5b6b4a0015ca59d/spec/react_on_rails/utils_spec.rb#L38)):*
> ```Ruby
>             allow(Webpacker).to receive_message_chain("manifest.lookup")
>               .with("server-bundle.js")
>               .and_return("/webpack/production/server-bundle-0123456789abcdef.js")
> ```

why this changed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-:-L89JewWUCrEs2qsH20j:b38mgu8)*

robwise:
Reviewed 2 of 8 files at r2, 2 of 7 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[README.md, line 560 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L87231tTrkxQgtQT8k6:-L89jpla11M4rJh7kdlx:b-94e1p8) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/README.md#L560)):*

This is not clear: 

- Caching only applies if pre-rendering (check the code).
- Link to the cache page that shows the equivalent if doing this manually.

</blockquote></details>

See my comment below, we can make this into something where we avoid computing the props if a cache key is hit. This would then be useful regardless of server rendering.

---

*[lib/react_on_rails/react_component/cache.rb, line 17 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L871QCkk5znM4u08D3U:-L89j2hK7UtOhNSm-EIG:b-5id0vz) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/lib/react_on_rails/react_component/cache.rb#L17)):*

@dzirtusss server_bundle_digest handles that part pl.us the component name. There's only one file for server rendering.
</blockquote></details>

What Justin said. 

Remember, if server rendering is off, `react_component` is just putting the component name and some props into a string in the HTML. 

With server rendering, it's generating the HTML that the server bundle returns from getting run with this component, but the client bundle is irrelevant to that process as it only gets executed on the client side.

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L89ewx22nliXgphaSKP:b-eo2wlv) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*

How do we know that device.mobile? does not return true for a tablet? I recently saw a bug where a tablet was displaying mobile layout.
</blockquote></details>

Different devices and browsers return different hints in the request headers that the Browser library uses to guess at what device the request came from. Unfortunately, it's not guaranteed to be accurate. This is probably why you saw that bug. 

Note that this is not necessary unless you are specifically server-rendering React components with JavaScript media queries (e.g., [react-responsive](https://github.com/contra/react-responsive)). Note that normal CSS media queries don't care.

I'm not really sure whether or not we should be building this into React on Rails. I would lean toward not and then putting a warning in the documentation that you should avoid JavaScript media queries or manually employ this technique with Browser which has the known limitations I mentioned above.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L89fdE_C_m9tKXJjIGk:b2vnb4j) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*

if you profile this stuff, you'd see that if we're **EVER** converting a giant hash to a string more than once, this code is NOT helping with performance.
</blockquote></details>

We absolutely need to profile all scenarios to ensure this PR brings a performance benefit at all, and if so, to identify which scenarios get that benefit and which do not.

After discussing with Justin, I think the first layer of caching should be based on a cache key instead of the props. This is because a lot of the computation, DB queries, and JSON rendering already happens by the time you get to the props. Therefore, we could either add a separate helper that works very similarly to `stale?` except automatically including the server bundle digest, or we could make `react_component` take a block where you pass the props in there. This way it only executes the block.

For example:

```ruby
react_component("MyComponent", cache: @user, prerender: true) do
  render("user.json", locals: { user: @user })
end
```

^ This way we avoid doing the rendering of the user.json which is costly in of itself and also may have additional costly computations or DB queries.

The `cache` option just passes through the value to a normal rails fragment cache except it puts the `server_bundle_digest` in there as well. We would need to add documentation emphasizing that `cache` should only be used if passing the props as a block instead of as an option.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

justin808:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[README.md, line 560 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L87231tTrkxQgtQT8k6:-L89yv0U5gojUPJVKIip:be9tntq) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/README.md#L560)):*

See my comment below, we can make this into something where we avoid computing the props if a cache key is hit. This would then be useful regardless of server rendering.
</blockquote></details>

Great idea @robwise. I commented below. 

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L89uQmCDJi8oW-SsLIf:bbjip4r) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*

We absolutely need to profile all scenarios to ensure this PR brings a performance benefit at all, and if so, to identify which scenarios get that benefit and which do not.

After discussing with Justin, I think the first layer of caching should be based on a cache key instead of the props. This is because a lot of the computation, DB queries, and JSON rendering already happens by the time you get to the props. Therefore, we could either add a separate helper that works very similarly to `stale?` except automatically including the server bundle digest, or we could make `react_component` take a block where you pass the props in there. This way it only executes the block.

For example:

```ruby
react_component("MyComponent", cache: @user, prerender: true) do
  render("user.json", locals: { user: @user })
end
```

^ This way we avoid doing the rendering of the user.json which is costly in of itself and also may have additional costly computations or DB queries.

The `cache` option just passes through the value to a normal rails fragment cache except it puts the `server_bundle_digest` in there as well. We would need to add documentation emphasizing that `cache` should only be used if passing the props as a block instead of as an option.
</blockquote></details>

@robwise that's a great idea. `props` is either a named param or is a block.

Maybe instead of naming the param `cache`, we use `cache_key`?

and yes, with this change, cache does make sense for the props string even without server rendering

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

justin808:
@robwise @mapreal19 Given the nature of the changes suggested, we need to timebox how much effort we put into this relative to expected gain over just having a helper to help compute the cache key and using the base rails helper call:

http://api.rubyonrails.org/classes/ActionView/Helpers/CacheHelper.html

---

Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-:-L89z5l09H7_yxlUggqm:bpsblpq)*

justin808:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L89zQeJ8tHL6Iv_B7oA:bdaskxv) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*
@robwise I'd probably check tablet before mobile, unless we're sure that both do not return true. I consider my tablet a "mobile" device.

> I'm not really sure whether or not we should be building this into React on Rails. I would lean toward not and then putting a warning in the documentation that you should avoid JavaScript media queries or manually employ this technique with Browser which has the known limitations I mentioned above.

I don't have an opinion on this one other than to avoid having this delay the fix one way or another.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L8BeN7x7Zy4ckuxE87R:bj5fviu) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*

@robwise I'd probably check tablet before mobile, unless we're sure that both do not return true. I consider my tablet a "mobile" device.

> I'm not really sure whether or not we should be building this into React on Rails. I would lean toward not and then putting a warning in the documentation that you should avoid JavaScript media queries or manually employ this technique with Browser which has the known limitations I mentioned above.

I don't have an opinion on this one other than to avoid having this delay the fix one way or another.

</blockquote></details>

I was copying what we had on HC. Maybe we don't want to include this for now. It's adding a new dependency and will not work for everyone

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L8BgIhVCGcHIB-iwTkV:b-x87mt5) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*

@robwise that's a great idea. `props` is either a named param or is a block.

Maybe instead of naming the param `cache`, we use `cache_key`?

and yes, with this change, cache does make sense for the props string even without server rendering



</blockquote></details>

I think this will add too much complexity and not sure if worth it. We already have options for caching json like `json.cache` in jBuilder.
What about using `props.hash` as @dzirtusss suggested? We should profile though to see if that's costly.

http://ruby-doc.org/core-2.4.1/Object.html#method-i-hash
Hash implementation in Rubinius: https://github.com/rubinius/rubinius/blob/5eab61ed73703362d612939fa8fb9f0ce9cf9f88/core/hash.rb#L387-L397

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

*[spec/react_on_rails/spec_helper.rb, line 97 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OnbIKwp_KXcfIVvA:-L8BhJNX9WCuK959Zuer:b-jcy4cq) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/spec/react_on_rails/spec_helper.rb#L97)):*

what's going on here?
</blockquote></details>

Reverted the change, I wanted this for filtering few specs but could be dangerous if someone commits a focused spec and goes into CI. We could a check `if ENV['CI'] == false` here though

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

justin808:

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L8BoMaXESPQIyMEKZpg:brr7u52) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

I was copying what we had on HC. Maybe we don't want to include this for now. It's adding a new dependency and will not work for everyone
</blockquote></details>

What we had there was clearly wrong so we have to fix both places ASAP.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L8BoVdmAvvq8c7Hrcj-:bexd55k) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

I think this will add too much complexity and not sure if worth it. We already have options for caching json like `json.cache` in jBuilder.
What about using `props.hash` as @dzirtusss suggested? We should profile though to see if that's costly.

http://ruby-doc.org/core-2.4.1/Object.html#method-i-hash
Hash implementation in Rubinius: https://github.com/rubinius/rubinius/blob/5eab61ed73703362d612939fa8fb9f0ce9cf9f88/core/hash.rb#L387-L397
</blockquote></details>

 Robynne I totally disagree with you. Let’s get this done. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

robwise:
It's a simple helper convenience change to the API, you could just use fragment caching but it's more verbose and the nonsense with the server bundle could be hidden by RoR. But it's not something that should take that long. Our discussions about how long it will take have taken longer than writing the code.

---

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L8Ea-jb90cEuFoayToD:b79vlgc) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*

What we had there was clearly wrong so we have to fix both places ASAP.
</blockquote></details>

Incorrect, it's not wrong on HC, it's the best we can do. There's nothing we can really do to fix it on HC, and it's been a known limitation for half a year. 

However, I don't think it should be included in RoR so I would remove. You're drastically reducing potential cache hits for all scenarios regardless of whether it's really necessary which 99% of the time it won't be. 

Users who have this problem can just include the device type in their `cache:` value and it should work fine for them, but we shouldn't force it on everyone.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L8E_WNSDiS19i4dbc0j:b5d3shv) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*

 Robynne I totally disagree with you. Let’s get this done. 
</blockquote></details>

Even with some partial caching, you'd still be avoiding DB queries and instantiating objects into memory in many situations if you do it this way. This whole concept is just as sugar for fragment caching so that we don't have to manually pass the server bundle name into the cache key every time ourselves. It's just a convenience thing.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-:-L8EacZO2uuQN88cu9nL:b-vy6ni2)*

justin808:
@mapreal19 what's the plan?

mapreal19:
@justin808 take a look now

---

Review status: 8 of 11 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[app/helpers/react_on_rails_helper.rb, line 101 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89L64Fruvd-t2SBjYD:-L8HS2JMMuyHIl35SpOo:b-q59rpa) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L101)):*

@mapreal19 this is where all the work is done to create the server rendering, so the cache part seems to be in the wrong place.

https://github.com/shakacode/react_on_rails/blob/master/app/helpers/react_on_rails_helper.rb#L331

and this is the main part of the code to avoid:
https://github.com/shakacode/react_on_rails/blob/master/app/helpers/react_on_rails_helper.rb#L377


Overall, we need to not get into ANY of the react on rails code **IF** cached and simply return the value from the cache.

We need to think at a high level on what the point of this fix really is?

Is it just to avoid JS execution of some JS code with the props?

And what are the costs already incurred to get to this point, including extra costs EVERY time:

* Calculation of the props with DB queries
* Conversion of a hash of props to a string
* Conversion of the string to a digest
* lookup in cache

</blockquote></details>

Done now?

---

*[app/helpers/react_on_rails_helper.rb, line 107 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89ImqpYVyU7t0YyTRn:-L8HS5fXKWrH5FCGEACi:bhpkkq0) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L107)):*

caching only applies if server rendered?
</blockquote></details>

what I got is that we include the server bundle if prerendering?

---

*[docs/additional-reading/caching-and-performance.md, line 6 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8HSHkS36wirxBAnE19:-L8HSHkS36wirxBAnE1A:bibzkdx) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/docs/additional-reading/caching-and-performance.md#L6)):*
> ```Markdown
> ## Caching
> 
> If you want to cache your component, pass the `cached` option:
> ```

Reminder: Update docs

---

*[lib/react_on_rails/utils.rb, line 23 at r1](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L84-RkGTw1ezLtd-a7R:-L8HSTx3-10Pc6ukEKFN:bk72oeh) ([raw file](https://github.com/shakacode/react_on_rails/blob/b6bf5969d9f4832a7b9c1ae4af1a21c4bd982f7a/lib/react_on_rails/utils.rb#L23)):*

I'm totally not clear how this relates to the rest of the fix.

just refactoring? or b/c server bundle hash _might_ not be using a sha in the file name?

and should we account for this, given that we're going to run this every time in development so if the sha is in the file name, computing an unnecessary digest from the HUGE JS file (React source etc.) is unnecessary.
</blockquote></details>

The problem was the client manifiest was overwriting  the server manifiest, so this is a temporary workaround. Would be nice if we could do get the server bundle hash as you had here before. @robwise knows about this issue

---

*[lib/react_on_rails/react_component/cache.rb, line 12 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OOM324BTAejtBgfa:-L8HSoOEiGU3N0iQr70p:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L12)):*

I think this needs to be at a different level so **NONE** of the RoR code runs if we're caching. This is called after ALL the work is already done.
</blockquote></details>

Done.

---

*[lib/react_on_rails/react_component/cache.rb, line 36 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OAk_hagGS33UEWBk:-L8HSsm57H3ap65q04j1:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L36)):*

Incorrect, it's not wrong on HC, it's the best we can do. There's nothing we can really do to fix it on HC, and it's been a known limitation for half a year. 

However, I don't think it should be included in RoR so I would remove. You're drastically reducing potential cache hits for all scenarios regardless of whether it's really necessary which 99% of the time it won't be. 

Users who have this problem can just include the device type in their `cache:` value and it should work fine for them, but we shouldn't force it on everyone.
</blockquote></details>

Done.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L8HSukrUk6Id1Gro9Er:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*

Even with some partial caching, you'd still be avoiding DB queries and instantiating objects into memory in many situations if you do it this way. This whole concept is just as sugar for fragment caching so that we don't have to manually pass the server bundle name into the cache key every time ourselves. It's just a convenience thing.
</blockquote></details>

Done.

---

*[spec/react_on_rails/utils_spec.rb, line 38 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89OwWPHlqUHF0NUFXd:-L8HSw18XnLJihTauFay:bhl2zih) ([raw file](https://github.com/shakacode/react_on_rails/blob/a6fc8d8ff1e37ab434777ff9c5b6b4a0015ca59d/spec/react_on_rails/utils_spec.rb#L38)):*

why this changed?
</blockquote></details>

See other comment on the manifiest being overwritten  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-:-L8HRmB3tOf_nonWZSCK:b4m7aap)*

mapreal19:
Review status: 8 of 11 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[README.md, line 548 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8HTC1UCbXM_MRtFjYT:-L8HTC1UCbXM_MRtFjYU:b2gy5mu) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/README.md#L548)):*
> ```Markdown
>                 props: {},
>                 prerender: nil,
>                 cached: false,
> ```

Reminder: update

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

dzirtusss:
Review status: 8 of 11 files reviewed at latest revision, 16 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/react_component/cache.rb, line 44 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89NwLGGA2Ym6Ft3tdV:-L8HyHnLDUmYhN_-kgdU:bdsaypr) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/lib/react_on_rails/react_component/cache.rb#L44)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

@mapreal19 @robwise @justin808

I vote 200% for rob's way (we should never execute props block if not needed), `props.hash` is a quick solution with the only idea in mind not to have huge cache keys. E.g. HC welcome page user props are 135k currently.

Furthermore, I guess we should do smth like:

Normal code:
```
def index
  @index_objects = Rails.cache (...) { very_complicated_logic }
end
```

Lazy code:
```
def index
end

private

def index_objects
  very_complicated_logic
end
helper_method :index_objects
```

Meaning if props are not called, json is not rendered, index_objects is not called. Everything w/o caching

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

justin808:
@mapreal19 Looks solid. Just change the cache => cache_key and make sure the docs are solid and I'll release this. Then let's get this on the https://github.com/shakacode/react-webpack-rails-tutorial

---

Review status: 8 of 11 files reviewed at latest revision, 19 unresolved discussions, some commit checks broke.

---

*[app/helpers/react_on_rails_helper.rb, line 107 at r3](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L89ImqpYVyU7t0YyTRn:-L8RBrAiiD9jlsDhW2gv:b-dgn1d6) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L107)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

what I got is that we include the server bundle if prerendering?
</blockquote></details>

My comment above is not relevant if you lazy calculating props.

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8RBmRaXCBTOTjMjbx5:-L8RBmRb63BEvknfIkNA:b-k6lagy) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/app/helpers/react_on_rails_helper.rb#L88)):*
> ```Ruby
>   #      not pass any props if you are separately initializing the store by the `redux_store` helper.
>   #   prerender: <true/false> set to false when debugging!
>   #   cache: TODO: explain
> ```

Maybe `cache_key`?

---

*[app/helpers/react_on_rails_helper.rb, line 262 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8RDc-ek3zVhj8nrZF5:-L8RDc-ek3zVhj8nrZF6:b-j5a917) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/app/helpers/react_on_rails_helper.rb#L262)):*
> ```Ruby
>       for an example of the necessary javascript configuration."
>     end
>   end
> ```

any diff or only moved?

---

*[spec/dummy/spec/helpers/react_on_rails_helper_spec.rb, line 183 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8REB7D5-zFk5HIAnYx:-L8REB7D5-zFk5HIAnYy:b-kapp9y) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb#L183)):*
> ```Ruby
>             props = { a: 1, b: 2 }
>             cache_keys = %w[a b]
>             react_component("App", cache: cache_keys) do
> ```

let's rename the prop to `cache_keys`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-:-L8RELtrEZOrKeOXSW-9:bmpt33r)*

justin808:
Reviewed 3 of 8 files at r4.
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks broke.

---

*[app/helpers/react_on_rails_helper.rb, line 262 at r4](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8RDc-ek3zVhj8nrZF5:-L8WJkNLCkyha0w8Ald-:bpt03r8) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/app/helpers/react_on_rails_helper.rb#L262)):*

any diff or only moved?
</blockquote></details>

only moved

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: 5 of 11 files reviewed at latest revision, 19 unresolved discussions.

---

*[CHANGELOG.md, line 12 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L87-hhwHxs4TZteRKSQ:-L8WXWeRCC1nnvoXX3aK:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/CHANGELOG.md#L12)):*

see other entries
</blockquote></details>

Done.

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r3](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L89IXIoBJpDBS5wy_6Y:-L8WVI3VEMdoiC8H2J3Y:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/b4964d4ed26f2621e9cabef2066227416e97abdd/app/helpers/react_on_rails_helper.rb#L88)):*

when prerender, etc
</blockquote></details>

Done.

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r4](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8RBmRaXCBTOTjMjbx5:-L8WUPmI5uwU5CrODEVp:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/app/helpers/react_on_rails_helper.rb#L88)):*

Maybe `cache_key`?
</blockquote></details>

Done.

---

*[spec/dummy/spec/helpers/react_on_rails_helper_spec.rb, line 183 at r4](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8REB7D5-zFk5HIAnYx:-L8WUO8e7k89LQwUc9Es:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb#L183)):*

let's rename the prop to `cache_keys`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: 5 of 11 files reviewed at latest revision, 18 unresolved discussions.

---

*[README.md, line 560 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L87231tTrkxQgtQT8k6:-L8WXflO4ULaQGfu5Vx6:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/README.md#L560)):*

Great idea @robwise. I commented below. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: 5 of 11 files reviewed at latest revision, 17 unresolved discussions.

---

*[docs/additional-reading/caching-and-performance.md, line 6 at r2](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L872tymRpgM0k4FfIie:-L8WXnlUE3Vyciq4tPOv:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/41fc72277029ef42b869bb00e7a81fb526259d50/docs/additional-reading/caching-and-performance.md#L6)):*

See my other comments.

In most cases, it's **better** to put the cache block outside of the calculation of the calculation of the props. 

So 100% explain how to do that...

I'm starting to think this is like the difference of calling:


```ruby
Rails.logger.info("BLAH #{some_super_time_consuming_crap}")
```

vs. **Using a BLOCK**
```ruby
Rails.logger.info { "BLAH #{some_super_time_consuming_crap}" }
```

So the block is never executed if the log level is not at info or debug.




</blockquote></details>

Done.

---

*[lib/react_on_rails/utils.rb, line 23 at r1](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L84-RkGTw1ezLtd-a7R:-L8WXs4Q4rhxvYsvkb64:b-wwsgbe) ([raw file](https://github.com/shakacode/react_on_rails/blob/b6bf5969d9f4832a7b9c1ae4af1a21c4bd982f7a/lib/react_on_rails/utils.rb#L23)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

The problem was the client manifiest was overwriting  the server manifiest, so this is a temporary workaround. Would be nice if we could do get the server bundle hash as you had here before. @robwise knows about this issue
</blockquote></details>

anything to do here?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

justin808:
A few small comments. Close! Get CI to pass as well. Then SHIP.

---

Reviewed 6 of 6 files at r5.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks broke.

---

*[CHANGELOG.md, line 15 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8Wa9tx7d0EfR4zsJmb:-L8Wa9tx7d0EfR4zsJmc:b-ogrkrs) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/CHANGELOG.md#L15)):*
> ```Markdown
> 
> #### Added
>  - Added `cache_key` option for `react_component` helper. [PR 1048](https://github.com/shakacode/react_on_rails/pull/1048) by [mapreal19](https://github.com/mapreal19)
> ```

and block for props

---

*[README.md, line 548 at r4](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8HTC1UCbXM_MRtFjYT:-L8WaDBH7CxAUIeDsY2I:b2nh8oq) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/README.md#L548)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

Reminder: update
</blockquote></details>

???

---

*[README.md, line 560 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WaIUJ1-E4S9ef0tZI:-L8WaIUJ1-E4S9ef0tZJ:b-lcfqfl) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/README.md#L560)):*
> ```Markdown
>   + **props:** Ruby Hash which contains the properties to pass to the react object, or a JSON string. If you pass a string, we'll escape it for you.
>   + **prerender:** enable server-side rendering of a component. Set to false when debugging!
>   + **cache_key:** set your cache keys here if using caching and passing props as a block.
> ```

Can we say that you specify this the same way as the Rails cache view helper?

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WaTrz0prKcpieKXYv:-L8WaTrz0prKcpieKXYw:b-haj98j) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/app/helpers/react_on_rails_helper.rb#L88)):*
> ```Ruby
>   #      not pass any props if you are separately initializing the store by the `redux_store` helper.
>   #   prerender: <true/false> set to false when debugging!
>   #   cache_key: String or Array containing your cache keys. If using caching you could pass props
> ```

"could" ==> "should" so that the props are not calculated if we have a cache value.

---

*[app/helpers/react_on_rails_helper.rb, line 107 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WaqmK6wEsodLbHkAe:-L8WaqmK6wEsodLbHkAf:b-810a72) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/app/helpers/react_on_rails_helper.rb#L107)):*
> ```Ruby
>     if use_caching
>       sanitized_options = raw_options
>       sanitized_options[:props] = yield
> ```

docs don't say you have to use a block for this....

I'd say to always use the block if caching, and throw if not provided and the key "props" is used.

Other alternative is to use the key props.

---

*[docs/additional-reading/caching-and-performance.md, line 7 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8Wb9Cv0vR31ptWK4mo:-L8Wb9Cv0vR31ptWK4mp:b9upog5) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/docs/additional-reading/caching-and-performance.md#L7)):*
> ```Markdown
> 
> If you want to cache your component, pass the `cache_key` option containing your cache keys.
> Also you should pass `props` as a block so you avoid generating them on each call:
> ```

could/should  ==> let's  make this consistent. See prior comments on this.

---

*[docs/additional-reading/caching-and-performance.md, line 11 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WbJrCAgpmik0pet7u:-L8WbJrCAgpmik0pet7v:bjb75kh) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/docs/additional-reading/caching-and-performance.md#L11)):*
> ```Markdown
> ```ruby
> react_component("App", cache_key: "cache_key") do
>   { prop1: "a", prop2: "b" }
> ```

This is not really instructive of why we're doing a block...

We should call some method like "slow_method_to_get_props"

---

*[lib/react_on_rails/react_component/cache.rb, line 15 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WbfbA9FwAKK4jviOq:-L8WbfbA9FwAKK4jviOr:btl383q) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/lib/react_on_rails/react_component/cache.rb#L15)):*
> ```Ruby
> 
>         def cache_key(component_name, options)
>           cache_keys = Array(options[:cache_key]).join("/")
> ```

document that Array can be passed!

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-:-L8WbuPM3COd0WU0e9U6:b-6l5kxd)*

mapreal19:
Review status: 7 of 11 files reviewed at latest revision, 17 unresolved discussions.

---

*[CHANGELOG.md, line 15 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8Wa9tx7d0EfR4zsJmb:-L8XOBL6ak3EYzSmotgx:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/CHANGELOG.md#L15)):*

and block for props
</blockquote></details>

Done.

---

*[README.md, line 548 at r4](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8HTC1UCbXM_MRtFjYT:-L8XLvqxkVIaQlsaJek1:b48nksc) ([raw file](https://github.com/shakacode/react_on_rails/blob/0eae49d1f5f9e7410115f49047dcd2e894edc469/README.md#L548)):*

???
</blockquote></details>

This was just a note for me for updating docs

---

*[app/helpers/react_on_rails_helper.rb, line 88 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WaTrz0prKcpieKXYv:-L8XOLjKCWGY2XzzoLEh:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/app/helpers/react_on_rails_helper.rb#L88)):*

"could" ==> "should" so that the props are not calculated if we have a cache value.
</blockquote></details>

Done.

---

*[app/helpers/react_on_rails_helper.rb, line 107 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WaqmK6wEsodLbHkAe:-L8X_186KI9oEmAMCrcI:b7p4dxc) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/app/helpers/react_on_rails_helper.rb#L107)):*

docs don't say you have to use a block for this....

I'd say to always use the block if caching, and throw if not provided and the key "props" is used.

Other alternative is to use the key props.
</blockquote></details>

Done. Added a throw for that case

---

*[docs/additional-reading/caching-and-performance.md, line 7 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8Wb9Cv0vR31ptWK4mo:-L8XNjbcYK6jtqDeOATm:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/docs/additional-reading/caching-and-performance.md#L7)):*

could/should  ==> let's  make this consistent. See prior comments on this.
</blockquote></details>

Done.

---

*[docs/additional-reading/caching-and-performance.md, line 11 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WbJrCAgpmik0pet7u:-L8XNrlWiquthg-Vcc0F:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/docs/additional-reading/caching-and-performance.md#L11)):*

This is not really instructive of why we're doing a block...

We should call some method like "slow_method_to_get_props"

</blockquote></details>

Done.

---

*[lib/react_on_rails/react_component/cache.rb, line 15 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WbfbA9FwAKK4jviOq:-L8XNwdAfu5uucB7Zg7B:b-q6rwpu) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/lib/react_on_rails/react_component/cache.rb#L15)):*

document that Array can be passed!
</blockquote></details>

Where? We have one spec on the helper spec showing that we can pass an array.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

mapreal19:
Review status: 6 of 11 files reviewed at latest revision, 17 unresolved discussions.

---

*[README.md, line 560 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WaIUJ1-E4S9ef0tZI:-L8X_PXpwpJ1KY0MSRzY:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/README.md#L560)):*

Can we say that you specify this the same way as the Rails cache view helper?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

dzirtusss:
Reviewed 1 of 7 files at r3, 6 of 6 files at r6.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

justin808:
Left a few comments. Can we please get this passing CI and then let's merge ASAP.

---

Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/react_component/cache.rb, line 15 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WbfbA9FwAKK4jviOq:-L8aDVG95kkzFkSoVUua:b-alic04) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/lib/react_on_rails/react_component/cache.rb#L15)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

Where? We have one spec on the helper spec showing that we can pass an array.
</blockquote></details>

In the allowable values for the cache. either a string or array of strings? does `hash` get called on each item in the array?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-:-L8aDkLG4jDAi7YazTii:b-vrobba)*

mapreal19:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[lib/react_on_rails/react_component/cache.rb, line 15 at r5](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048#-L8WbfbA9FwAKK4jviOq:-L8aNrD0KEVfB_57q9tA:bq1fg6c) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/lib/react_on_rails/react_component/cache.rb#L15)):*
> does hash get called on each item in the array?

No, it's just a string or array of strings that will go into the cache key.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react_on_rails/1048)*

dzirtusss:
Reviewed 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048)*

justin808:
What's remaining?

---

Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[README.md, line 562 at r7](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8co0GR8gCIEk-Y6GbQ:-L8co0GR8gCIEk-Y6GbR:bq8ialx) ([raw file](https://github.com/shakacode/react_on_rails/blob/6d0587a4da435d3329bcdf90739711cd34d0d0c1/README.md#L562)):*
> ```Markdown
>   + **props:** Ruby Hash which contains the properties to pass to the react object, or a JSON string. If you pass a string, we'll escape it for you.
>   + **prerender:** enable server-side rendering of a component. Set to false when debugging!
>   + **cache_key:** set your cache keys here if using caching and passing props as a block. Same as if using Rails `cache` helper.
> ```

@mapreal19 did you verify this?

---

*[lib/react_on_rails/react_component/cache.rb, line 15 at r5](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-L8WbfbA9FwAKK4jviOq:-L8co49F516V6I9Wi00z:b-o7vq9a) ([raw file](https://github.com/shakacode/react_on_rails/blob/f98d2037080eace0b5bafc201a7d7a423255718d/lib/react_on_rails/react_component/cache.rb#L15)):*
<details><summary><i>Previously, mapreal19 (Mario Pérez) wrote…</i></summary><blockquote>

> does hash get called on each item in the array?

No, it's just a string or array of strings that will go into the cache key.
</blockquote></details>

see comment above regarding same as rails cache helper

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1048#-:-L8co9-7Bz7n8oQYIMpc:b-4tprv0)*

mapreal19:
fyi `utils_spec` failing on CI but not in local. Don't know how to reproduce...

PR #1057, opened 2018-04-12
Update README.md

hrishimittal:
In the basic installation instructions, add the step for committing the code to git before running the generator. Without it, anyone not familiar with the gem will get stuck before they can start using it.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1057)


justin808:
Thanks @hrishimittal!

PR #1061, opened 2018-04-19
Add Information about deploying with Capistrano

mdibraheem:
I could not find any details regarding deploying with Capistrano in the repository. Following the tutorial results in the following error at assets:precompile stage. 
> No such file or directory - /node_modules/.bin/webpack

Because node_modules are usually not checked in the repo, It could not find webpack. Adding `capistrano-yarn` basically runs `yarn install` before running webpack.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1061)


mdibraheem:
My Bad. I didn't realize there were other pages linking to this page and renaming it will break that.

mdibraheem:
If you are ok with the content, let me know if you want me to submit another PR with a separate file for Capistrano Deployment or something else.

justin808:
@mdibraheem Does this solve a problem other than:

> Because node_modules are usually not checked in the repo, It could not find webpack. Adding capistrano-yarn basically runs yarn install before running webpack.

If one uses the buildpack for node:


```
heroku buildpacks:set heroku/ruby
heroku buildpacks:add --index 1 heroku/nodejs
```

I think that yarn runs.

mdibraheem:
This is for the scenario that you are deploying to a VPS like digital ocean using Capistrano. I am not sure how to use buildpacks for that. But using Capistrano it does not run by default.

mdibraheem:
My mistake on adding this to heroku-deployment file. I thought it would be better to add it here and rename the file to deployment.md

justin808:
@mdibraheem I would not change the heroku deployment file name. Maybe create a new file for capistrano?

mdibraheem:
I have added capistrano-deployment.md and left heroku-deployment.md untouched.

Issue #1062, opened 2018-04-20
Version checking mismatch bug


dzirtusss:
There is a possibility of versions mismatch when version in `package.json` is accidentally specified in non-exact style e.g. `^10.1.3`. If later version is bumped to say `10.1.4` in node packages this is not noticed by the gem.

I guess this could be fixed easily by forcing only exact version in `package.json` by `version_checker.rb`, or at least having a warning that version is not exact.

justin808:
Here is the file to change: https://github.com/shakacode/react_on_rails/blob/49401b2b48aaf397de625f988ab0617ed4499fc4/lib/react_on_rails/version_checker.rb

and the test to update: https://github.com/shakacode/react_on_rails/blob/94868d0810a4f0e51b8e8d7ed79d73d7bcaae342/spec/react_on_rails/version_checker_spec.rb

justin808:
https://github.com/bundler/bundler/issues/6500 to see if I can find out if the ruby gem has a semver wildcard.

justin808:
Fixed in #1063.

PR #1063, opened 2018-04-21
More exact version checking

justin808:
We keep the react_on_rails gem and the react-on-rails node package at
the same exact versions so that we can be sure that the interaction
between them is precise.

This is so that if a bug is detected after some update, it's critical that
both the gem and the node package get the updates.

This change ensures that the package.json specification does not use a
~ or ^ as reported in https://github.com/shakacode/react_on_rails/issues/1062

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1063)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ Changes since last non-beta release.
 - Removed ReactOnRails::Utils.server_bundle_file_name and ReactOnRails::Utils.bundle_file_name.
 - No longer logging the `railsContext` when server logging.
 
+#### Fixes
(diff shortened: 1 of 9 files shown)


justin808:
@dzirtusss @mapreal19 can you guys still review?

mapreal19:
:lgtm_strong: code looks good

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/shakacode/react_on_rails/1063#-:-LAnh1n1ERx10bj5RizX:b-9m6o47)*

PR #1082, opened 2018-05-10
spec/dummy/README missed package name change

ened:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1082)


justin808:
https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/client/package.json#L70

Let's mention this is done in the postinstall script.

justin808:
Thanks @ened. We can apply my suggestion in another PR if you like.

PR #1083, opened 2018-05-11
corrects bundle exec rake instead of rails in README.md

RolandStuder:
Just small error in the readme. You put `rails webpacker...` instead of `rake...`

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1083)

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,8 +206,8 @@ See [Issue 982: Tutorial Generating Correct Project Structure?](https://github.c
 
 2. Run the following 2 commands to install Webpacker with React:
    ```
-   bundle exec rails webpacker:install
-   bundle exec rails webpacker:install:react
+   bundle exec rake webpacker:install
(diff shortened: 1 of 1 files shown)


rodrigobdz:
Since Rails 5.0+ it is not wrong to call a task with `rails`, such as `webpacker:install`.
@RolandStuder Even the [rails docs](http://guides.rubyonrails.org/command_line.html#bin-rails) **recommend** running tasks with `rails`.

> Since Rails 5.0+ has rake commands built into the rails executable, bin/rails is the new default for running commands.

justin808:
Agree with @rodrigobdz. I also think that most 4.x Rails users will know that.

RolandStuder:
Ah, I learned something, thanks for the feedback. And have a wonderful day!

PR #1085, opened 2018-05-12
Update react_on_rails gem version

rodrigobdz:
When running `bin/setup` as suggested in the [contributing guide](https://github.com/shakacode/react_on_rails/blob/fa52ac65d0248e002dd7fb4b22e1fec7a84aebd4/CONTRIBUTING.md#prereqs), the Gemfile.lock gets automatically updated to the newest react_on_rails version. This should point to the newest version already.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1085)

### DIFF
--- a/spec/dummy/Gemfile.lock
+++ b/spec/dummy/Gemfile.lock
@@ -7,7 +7,7 @@ GIT
 PATH
   remote: ../..
   specs:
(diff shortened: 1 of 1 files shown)


justin808:
Thanks @rodrigobdz

PR #1107, opened 2018-06-24
Handle <script async> transparently #1099

### LINKED ISSUE #290: React on Rails and async loading of application.js
Hi guys,
i'm wondering if there is any possibility to use react on rails with async loading like:
```
<%= javascript_include_tag 'application', async: Rails.env.production? %>
```

In this case, the ```componentDidMount``` does not get fired. The issue is well known, described and fixed here: https://github.com/reactjs/react-rails/issues/218

What i'm searching for is a hook, like ```ReactRailsUJS.mountComponents()```. Right now i'm trying to write something like ```mountComponents()``` but i did not have success until now. That's why i'm wondering if i'm missing something?

Thanks a lot in advance.

justin808:
Fixes #290, transparently handling <script async> tag. If the document was already loaded,
the DOMContentLoaded event handler no longer fires, so we just execute it directly, the
same way as jQuery does in $.ready().

Co-authored-by: Alexey Makhotkin

Discussion in #1099.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1107)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,23 +9,24 @@ See [Upgrading React on Rails](./docs/basics/upgrading-react-on-rails.md) for mo
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
-### [11.0.9] - 2018-05-22
+
+### [11.0.9] - 2018-06-24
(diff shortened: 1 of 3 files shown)


squadette:
Thank you!

So far it works in my testing, but more combinations of browsers / connection speeds is needed.  I'll update accordingly.

PR #1136, opened 2018-08-21
Don't apply babel-plugin-transform-runtime inside react-on-rails to work with babel 7

tricknotes:
Don't apply babel-plugin-transform-runtime inside react-on-rails to work with babel@7

It should be applied by user land.

Why?
The generated code depends on built babel's version.
It means that user couldn' build their code with babel@7 that depends on react-on-rails built by babel@6.

For example, babel-plugin-transform-runtime (babel 6) generates the 
following code.
``` js
var _assign = require('babel-runtime/core-js/object/assign');
```

However @babel/plugin-transform-runtime (babel 7) expects the 
path that beginning of `@babel/runtime/`.
                     
The change allows to use react-on-rails on babel@7.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1136)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+#### Fixed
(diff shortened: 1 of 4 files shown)


justin808:
@tricknotes what about babel@6?

@alexfedoseev any idea on this one?

PR #1140, opened 2018-09-08
Work on babel 7 with transform-runtime

tricknotes:
Previously, I sent a PR to stop using babel-plugin-transform-runtime. https://github.com/shakacode/react_on_rails/pull/1136
But this breaks build for legacy browser.

So I fixed to apply babel-plugin-transform-runtime.
And it is allow to use for Babel 7 users by providing babel-runtime as "dependencies".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changes since last non-beta release.
 
 #### Fixed
 - Ignore Arrays in Rails i18n yml files. [PR 1129](https://github.com/shakacode/react_on_rails/pull/1129) by [vcarel](https://github.com/vcarel).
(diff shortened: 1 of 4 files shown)


tricknotes:
Thanks for your review, @justin808 !

> Should we upgrade the whole project to babel 7 with this?

It doesn't depend on this change.
However upgrading to babel 7 is not difficult.
Which do you like pushing the change to this PR or creating another PR?

And I'm sorry not to see your review comment from reviewable.io. (It seems leaving empty :<)


Could you tell me what you want to comment?

tricknotes:
@justin808 
Awesome thanks ✨ 
I sent a PR to upgrade Babel version to 7. https://github.com/shakacode/react_on_rails/pull/1141

PR #1142, opened 2018-09-12
Update installation docs

ashgaliyev:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1142)


justin808:
Thanks @ashgaliyev. Longer term, we should maybe use XX.XX.XX and ask the person following the instructions to get the version number from the main readme. Actually, even better, please copy the badge over to these 2 pages. Let's do that for this change. Otherwise, we'll never fix this.

justin808:
i updated the to the current version already 11.1.4 with the new release.

Issue #1150, opened 2018-09-27
reactOnRailsPageLoaded invoked before deferred scripts are completed


rakelley:
When Turbolinks is disabled, `clientStartup.js` can invoke `reactOnRailsPageLoaded` in a manner that appears premature and unintentional (and is causing my team no end to grief).

The offending code is the timeout declared [here](https://github.com/shakacode/react_on_rails/blob/master/node_package/src/clientStartup.js#L199) with the condition `if (document.readyState === 'complete' || document.readyState !== 'loading' && !document.documentElement.doScroll) {`.  A readyState of `interactive` while deferred scripts are still being processed will lead to `reactOnRailsPageLoaded` being called before all components have been registered.  From our testing, Chrome seems to be the only browser fast enough to actually trigger this behavior, but it's a serious issue for us, made worse if utilizing preloading/server-push.

By the surrounding debug messages and the choice to use `DOMContentLoaded` event handlers, the intention seems to be that `reactOnRailsPageLoaded` should not be invoked until the page is actually loaded, which is not merely `readyState = interactive` but post-`DOMContentLoaded`.

Is it intentional that `reactOnRailsPageLoaded` is called before `DOMContentLoaded`, if any only if Turbolinks is disabled?

justin808:
@rakelley I'm open to a PR for this. I can push a beta version for you.

> Is it intentional that reactOnRailsPageLoaded is called before DOMContentLoaded, if any only if Turbolinks is disabled?

Certainly not. I really cannot recollect this logic. It's been this way for years.

justin808:
@rakelley please email me justin@shakacode.com and we can probably work this out pretty quickly.

justin808:
Shipped in 11.0.6.

@rakelley can you confirm that it works as expected?

rakelley:
@justin808 Yes, 11.0.6 is working for us in all scenarios, thank you for the quick turn-around.

PR #1151, opened 2018-10-01
Fix render during interactive readyState

rakelley:
This fixes a regression introduced by [PR 1099](https://github.com/shakacode/react_on_rails/pull/1099), explained in [issue 1150](https://github.com/shakacode/react_on_rails/issues/1150).  The potential for this issue [was raised](https://github.com/shakacode/react_on_rails/pull/1099#issuecomment-397938518) during review of PR 1099 but not reproduced at that time, and is currently only reproducible for us in Chrome when utilizing `defer`ed scripts which are preloaded.

The code was cloned from jQuery `$.ready()`, with the offending part of the condition existing to support IE versions < 9.  This PR introduces a backwards-compatibility-break for those versions of IE, but should have no impact on any non-EOL browsers other than resolving this bug.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1151)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
+- Fix client startup invoking render prematurely.  Closes [issue #1150](https://github.com/shakacode/react_on_rails/issues/1150).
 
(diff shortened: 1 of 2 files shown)


justin808:
@alexeyMohnatkin Can you please review this? This change would revert part of your change for #1099. 

@rakelley Thanks for submitting this. Once this passes CI and so long as @alexeyMohnatkin does not have any objection, I'll merge.

PR #1152, opened 2018-10-03
Fix failure to render during interactive readyState

rakelley:
This is a follow-up to [PR 1151](https://github.com/shakacode/react_on_rails/pull/1151).  I apologize for doubling up on this, but the last change was insufficient and we've found additional edge cases of undesired behavior with aggressive preloading of deferred scripts.

In addition to solving behavior for non-turbolinks users, this change also restores the original intent of `clientStartup()`, which explicitly calls out the turbolink check as needing to be done AFTER page load.  Since [PR 1099](https://github.com/shakacode/react_on_rails/pull/1099) this check has potentially been done prior to load.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1152)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
+- Fix additional cases of client startup failing during interactive readyState
 
 ### [11.1.5] - 2018-09-12
 #### Fixed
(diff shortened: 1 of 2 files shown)


justin808:
I just released 11.0.6. 11.0.5 is broken. 

@rakelley can you check that this comment is addressed. https://github.com/shakacode/react_on_rails/commit/46350cec56297aef1f94b98f1606ad02780ae107#commitcomment-30794250

If there are more changes to this, I'll push another release.

PR #1129, opened 2018-08-10
Ignore Arrays in Rails i18n yml files

vcarel:
Fixes #1126

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1129)

PR #1153, opened 2018-10-10
Fixed intl parsing bug when trying to parse non-string entries

sepehr500:
Fixed intl parsing bug when trying to parse non-string entries.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1153)

### DIFF
--- a/lib/react_on_rails/locales_to_js.rb
+++ b/lib/react_on_rails/locales_to_js.rb
@@ -107,7 +107,7 @@ module ReactOnRails
       translations.each_with_object({}) do |(k, v), h|
         if v.is_a? Hash
           flatten(v).map { |hk, hv| h["#{k}.#{hk}".to_sym] = hv }
-        elsif !v.is_a? Array # Arrays are not supported by react-intl
+        elsif v.is_a?(String) # Arrays are not supported by react-intl
           h[k] = v.gsub("%{", "{")
         end
       end
--- a/lib/react_on_rails/locales_to_js.rb
+++ b/lib/react_on_rails/locales_to_js.rb
@@ -107,7 +107,7 @@ module ReactOnRails
       translations.each_with_object({}) do |(k, v), h|
         if v.is_a? Hash
           flatten(v).map { |hk, hv| h["#{k}.#{hk}".to_sym] = hv }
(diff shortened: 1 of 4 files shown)


justin808:
@sepehr500 please consult with @vcarel regarding #1129.

After that is done, I can assess this.

lib/react_on_rails/locales_to_js.rb
```diff
@@ -107,7 +107,7 @@ def flatten(translations)
       translations.each_with_object({}) do |(k, v), h|
         if v.is_a? Hash
           flatten(v).map { |hk, hv| h["#{k}.#{hk}".to_sym] = hv }
-        elsif !v.is_a? Array # Arrays are not supported by react-intl
+        elsif v.is_a? String # Arrays are not supported by react-intl
```
vcarel: The comment does not seem appropriate anymore

sepehr500: @vcarel Do you think we should support int and bools as well as strings?

vcarel:
@sepehr500 What issue does it address? I confess I'm concerned about not exporting empty keys, because react-intl would then output the name of the missing keys instead of just blank strings.

@justin808 The root cause of those issues is that we attempt to pass translations to react-intl which will never be used on React's end. Instead, shouldn't we encouraging users to cleanup their translations files and pass only what React will really use?

lib/react_on_rails/locales_to_js.rb
vcarel: Looks like int and bools are not supported -> https://formatjs.io/guides/message-syntax/

vcarel: ...I mean by react-intl.
I don't know, see my comment to @justin808 below. I don't feel this is going in the right direction, although my last PR is part of the problem.
The decision is up to both of you.

sepehr500:
@vcarel There are people who use part of their yml for react and the other part for rails. So forcing people to clean up their yml will not work because the libraries support different data types. This PR is to stop the build from failing when you have a type that can not have `.gsub` called on it.

justin808:
Nice job @sepehr500. @vcarel please give this a try asap. I'm pushing 11.0.7 ASAP.

PR #1159, opened 2018-10-19
`package.json` updates

coopersamuel:
This PR updates _most_ of the packages in `package.json`. I'll leave webpack for a separate PR because I anticipate breaking changes moving to v4.

More comments are in the code


.eslintrc
```diff
@@ -15,6 +15,9 @@ env:
 
 rules:
   no-console: 0
+  function-paren-newline: 0
+  object-curly-newline: 0
+
```
coopersamuel: After updating `eslint` these two rules produced over 120 errors. I disabled them in the `.eslintrc` file

node_package/src/buildConsoleReplay.js
```diff
@@ -5,6 +5,7 @@ import scriptSanitizedVal from './scriptSanitizedVal';
 
 export function consoleReplay() {
   // console.history is a global polyfill used in server rendering.
+  // $FlowFixMe
```
coopersamuel, line 8: As per [this flow article](https://medium.com/flow-type/new-flow-errors-on-unknown-property-access-in-conditionals-461da66ea10), flow is now throwing errors for unknown property access in conditionals. I flagged this with `$FlowFixMe` for the time being to suppress the error. @justin808 what are your thoughts on fixing this?

node_package/src/clientStartup.js
```diff
@@ -204,7 +204,7 @@ function renderInit() {
 }
 
 export function clientStartup(context) {
-  const document = context.document;
+  const { document } = context;
```
coopersamuel, line 207: Reformatted to satiate `eslint`

node_package/src/serverRenderReactComponent.js
```diff
@@ -42,7 +42,7 @@ See https://github.com/shakacode/react_on_rails#renderer-functions`);
 
       if (reactElementOrRouterResult.redirectLocation) {
         if (trace) {
-          const redirectLocation = reactElementOrRouterResult.redirectLocation;
+          const { redirectLocation } = reactElementOrRouterResult;
```
coopersamuel, line 45: More destructuring to make `eslint` happy

node_package/tests/ReactOnRails.test.js
```diff
@@ -21,7 +21,7 @@ test('ReactOnRails render returns a virtual DOM element for component', (assert)
   ReactOnRails.register({ R1 });
 
   // eslint-disable-next-line no-underscore-dangle
-  const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalInstance._currentElement.type;
+  const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalFiber.type;
```
coopersamuel, line 24: When updating to React v16 there was a breaking change that caused the below test to fail
```
test('ReactOnRails render returns a virtual DOM element for component', (assert) => {
  assert.plan(1);
  const R1 = createReactClass({
    render() {
      return (
        <div> WORLD </div>
      );
    },
  });
  ReactOnRails.register({ R1 });
  // eslint-disable-next-line no-underscore-dangle
  const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalInstance._currentElement.type;
  assert.deepEqual(actual, R1,
    'ReactOnRails render should return a virtual DOM element for component');
});
```
The error occurred because _reactInternalInstance was deprecated in React 16. The object returned from ReactDOM.render() for v15.6 looked like this (simplified):
```
{
    _reactInternalInstance: {
        _currentElement: {
            type: // Stuff here
        }
    }
}
```
After upgrading to React v16, the object looked like this (again, simplified):
```
{
    _reactInternalFiber: {
        type: // Stuff here
    },
    _reactInternalInstance: {
        // EMPTY!
    }
}
```
A one line update to the test above resulted in all tests passing
`const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalFiber.type;`

node_package/tests/buildConsoleReplay.test.js
```diff
@@ -61,8 +61,11 @@ console.warn.apply(console, ["other message","{\\"c\\":3,\\"d\\":4}"]);`;
 test('consoleReplay replays converts script tag inside of object string to be safe ', (assert) => {
   assert.plan(1);
   console.history = [
-    { arguments: ['some message </script><script>alert(\'WTF\')</script>',
-      { a: 'Wow</script><script>alert(\'WTF\')</script>', b: 2 }],
+    {
+      arguments: [
+        'some message </script><script>alert(\'WTF\')</script>',
+        { a: 'Wow</script><script>alert(\'WTF\')</script>', b: 2 },
+      ],
```
coopersamuel, line 68: Reformatted for `eslint`

coopersamuel:
@justin808 regarding your comments, what steps do you think need to be taken to pass this PR?

1. Regarding the `eslint` rules, if we remove those rules we'd have a bunch of errors (> 120) - my position is that we leave the rules for now and add `prettier` in the project in a separate PR

2. Regarding the test, I agree that test doesn't really do much, personally I think the test is now obsolete because of React's behavior removing the `data-react-checksum` attribute. Should we scrap the test entirely?

Let me know your thoughts on these and we'll get this pushed through

PR #1185, opened 2018-12-26
[Issue-1182] Change the default generator output directory

wanglixiang1992:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1185)


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -18,13 +18,21 @@ class BaseGenerator < Rails::Generators::Base
                    desc: "Install Redux gems and Redux version of Hello World Example",
                    aliases: "-R"
 
+      def change_webpacker_source_path
+        webpacker_yml = File.join(destination_root, "config/webpacker.yml")
+        return unless File.exist?(webpacker_yml)
+
+        contents = File.read(webpacker_yml).gsub(/source_path:.*/, "source_path: client/app")
```
justin808: Let's think about this.

* Probably, we should match exactly.
* The problem case is when this was configured differently.
* Webpacker may have been previously used for Vue.js, or even React
* We will be changing the directory to what we think is the best approach
* rails/webpacker and the app will break b/c this is changed
* We can do one of 2 things:
  * Not change the default in this case
  * Add such details to the post-generation message so that it's easy for the developer to revert this change.

lib/generators/react_on_rails/react_no_redux_generator.rb
```diff
@@ -11,9 +11,9 @@ class ReactNoReduxGenerator < Rails::Generators::Base
       source_root(File.expand_path("templates", __dir__))
 
       def copy_base_files
-        base_js_path = "base/base"
-        base_files = %w[app/javascript/bundles/HelloWorld/components/HelloWorld.jsx]
-        base_files.each { |file| copy_file("#{base_js_path}/#{file}", file) }
+        base_js_path = "base/base/app/javascript"
```
justin808: This should change to `base/base/client/app`

create a client directory here: https://github.com/shakacode/react_on_rails/tree/master/lib/generators/react_on_rails/templates/base/base

lib/generators/react_on_rails/base_generator.rb
justin808: Could we DRY up `client/app` so that this case and other places use this?

justin808: Let's update this file: https://github.com/shakacode/react_on_rails/blob/master/lib/generators/USAGE

to indicate where the files are installed and what you should do to revert back to the rails app/javascript default.

Indicate:

```
The team at ShakaCode prefers to have all client side development files in a top-level directory called /client. This more cleanly separates the front-end code from the back-end code.

To change back to the default...
```

try to finish up this message to what you think is best.

justin808: Make sure this message runs after running the generator.

justin808: @wanglixiang1992 Let's use a constant for "client/app" to DRY this up. DRY = Don't Repeat Yourself

justin808:
@wanglixiang1992 I updated from master. CI should now pass.

justin808:
@wanglixiang1992 a few more things that we didn't consider:

* if the user is already using /app/javascript and then we change the webpacker.yml setup to use /client/app, that might be confusing. I think the postinstall message should make it clear.

* I'll test this out. If you have any thoughts, let me know.

Also, we need a CHANGELOG.md line.

justin808:
@Judahmeek and @behraaangm how useful would this one be?

justin808:
I probably don't want the default generator to deviate from the standard Rails configuration. I'll keep this open and look at this in a bit more detail.

justin808:
As of 2020, I'm going to be recommending to stick with the default rails/webpacker setup so that React on Rails installs are most familiar to all those in the Rails community.

PR #1200, opened 2019-03-18
Converting spec/dummy test app to use rails/webpacker for config

justin808:
Moving from a custom config to rails/webpacker config
See docs https://github.com/rails/webpacker/blob/master/docs/webpack.md.

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1200)


justin808:
See #1247

PR #1206, opened 2019-04-02
Fix dependencies for dummy applications

joeytepp:
While moving a project to the [recommended project structure](https://github.com/shakacode/react_on_rails/blob/master/docs/basics/recommended-project-structure.md) I noticed an issue with the configuration in [/spec/dummy/client](https://github.com/shakacode/react_on_rails/tree/master/spec/dummy/client), which is recommended as an example for configuration. The issue was that the command `babel-node` was present in one of the scripts but not listed in the dependencies, thus needing it to be manually installed. A very small fix but should help other developers use this in their own projects with ease!

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1206)

### DIFF
--- a/spec/dummy/client/package.json
+++ b/spec/dummy/client/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
+    "@babel/node": "^7.2.2",
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/polyfill": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
--- a/spec/dummy/client/yarn.lock
+++ b/spec/dummy/client/yarn.lock
@@ -212,6 +212,17 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
+"@babel/node@^7.2.2":
+  version "7.2.2"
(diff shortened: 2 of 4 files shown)


justin808:
@joeytepp I took a look at https://babeljs.io/docs/en/babel-node

It seems useful, but I don't believe that it's necessary given the current setup. 

Yes, babel-node is referenced but I'm not sure the hot-assets node scripts works:
https://github.com/shakacode/react_on_rails/search?q=babel-node&unscoped_q=babel-node

We've always been planning to add hot reloading examples back to the project, but we never got around to it. 

Maybe there's a bit more in terms of setup that you can add?

Or does the hot-assets task actually just work as-is with just this change from the PR?

joeytepp:
@justin808 The `hot-assets` command works with my changes, fails without. Happy to take a look and see what else can be improved for the setup but I feel this change is necessary 👍

justin808:
> @justin808 The `hot-assets` command works with my changes, fails without. Happy to take a look and see what else can be improved for the setup but I feel this change is necessary 👍

Hi @joeytepp please ask me for a slack invite at [justin@shakacode.com](mailto:justin@shakacode.com). I'd like to do a screen share.

joeytepp:
Will do 👌

PR #1210, opened 2019-04-10
Add webpacker for Rails 5.1+ in tests

kvokka:
Since Rails 5.1+ you can find the option the new projects `rails new foo --webpack`. 
After the Rails 6.0 webpack should be installed by default.

Environment:
Ruby 2.6.1p33 (2019-01-30 revision 66950) [x86_64-darwin18]
Rails 5.2.3

Steps to reproduce:

`$ rake`

error message
```
(...cut)

cd /Users/mike/proj/react_on_rails/gen-examples/examples/basic && touch .gitignore
cd /Users/mike/proj/react_on_rails/gen-examples/examples/basic && rake webpacker:install
rake aborted!
Don't know how to build task 'webpacker:install' (See the list of available tasks with `rake --tasks`)

(See full trace by running task with --trace)
rake aborted!
Command failed with status (1): [cd /Users/mike/proj/react_on_rails/gen-exa...]
/Users/mike/proj/react_on_rails/rakelib/task_helpers.rb:22:in `block in sh_in_dir'
/Users/mike/proj/react_on_rails/rakelib/task_helpers.rb:22:in `each'
/Users/mike/proj/react_on_rails/rakelib/task_helpers.rb:22:in `sh_in_dir'
/Users/mike/proj/react_on_rails/rakelib/examples.rake:36:in `block (3 levels) in <top (required)>'
Tasks: TOP => default => run_rspec => run_rspec:run_rspec => run_rspec:examples => examples:gen_all => examples:gen_basic
(See full trace by running task with --trace)
```

So it looks like somewhere between Rails 5.1.0 and Rails 5.2.3 breaking change appears.

This PR fix this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1210)

### DIFF
--- a/rakelib/example_type.rb
+++ b/rakelib/example_type.rb
@@ -49,8 +49,9 @@ module ReactOnRails
       end
 
       # Options we pass when running `rails new` from the command-line.
+      attr_writer :rails_options
       def rails_options
-        "--skip-bundle --skip-spring --skip-git --skip-test-unit --skip-active-record"
+        @rails_options ||= "--skip-bundle --skip-spring --skip-git --skip-test-unit --skip-active-record"
       end
 
       %w[gen clobber npm_install build_webpack_bundles].each do |task_type|
--- a/rakelib/examples.rake
+++ b/rakelib/examples.rake
@@ -6,6 +6,8 @@
 # Also see example_type.rb
 
 require "yaml"
+require 'rails/version'
+
 require_relative "example_type"
 require_relative "task_helpers"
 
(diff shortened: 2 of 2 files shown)


justin808:
@kvokka did this show up in CI?

kvokka:
@justin808  No, but in CI you can find this line https://travis-ci.org/shakacode/react_on_rails/jobs/511766961#L467 under the cut. So, technically, with this we are creating extra hidden dependency. It may be a problem for new contributions - it takes time to figure out, that this ENV variable MUST be set even for development.

At least will be good idea to add some extra info about this hack to the documentation (or just merge this fix)

justin808:
I think I'd rather just have some documentation. The `--webpack` option shouldn't ever be necessary because another bit of code handles this. I don't recall exactly where. Nevertheless, CI runs with Rails 5.2 just fine.

justin808:
@kvokka How about just having this value be the default?

`export BUNDLE_GEMFILE=$PWD/spec/dummy/Gemfile`

kvokka:
@justin808  It was exactly what i tried to avoid. It may affect other components which use bundler. + you'll require to calculate the actual path, which will add a complexity.

+ if you will use `export`, u overlap the previous value if the global variable, which may cause some artifacts.

The another option will be to manually provide this value in the rake tasks, but actually it is much more heavy than just `--webpack` option.

As an option developer may enter the ENV variable directly (and document this approach) but it looks dirty, don't you think?

justin808:
@kvokka the only reason the alternate gemfile exists is to support Rails 3.2, so that probably won't continue forever. I think a default value that can be overridden is reasonable. Agree?

justin808:
@kvokka thank you for the contribution. Seems useful.

PR #1214, opened 2019-05-04
Add configuration option to camelCase props

joeytepp:
When passing props in server-side rendering, they are often in snake_case, since this is ruby standard, however many prefer to use camelCase when working with javascript. This PR introduces the configuration option to automatically convert any props that have keys in snake_case to camelCase. Tests included.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1214)

### DIFF
--- a/docs/basics/configuration.md
+++ b/docs/basics/configuration.md
@@ -183,6 +183,13 @@ ReactOnRails.configure do |config|
   
   ################################################################################
   ################################################################################
+  # CAMELCASE PROPS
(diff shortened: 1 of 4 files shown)


justin808:
@joeytepp I'm not convinced that that convenience option is needed in the gem because it's easy enough to create your own wrapper that does this conversion of props.

joeytepp:
@justin808 that is fair, also seems to be problematic as well since `deep_transform_keys!` is not supported by older versions of rails, so gonna close this for now.

PR #1244, opened 2020-01-21
Changing to standard webpacker config for the spec/dummy

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1244)


justin808:
Merged in #1247

PR #1258, opened 2020-03-27
rubocop fixes

behraaangm:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1258)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,7 +3,7 @@
 
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.5
 
   Include:
     - '**/Rakefile'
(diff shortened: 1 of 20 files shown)


behraaangm:
___
*[spec/dummy/spec/requests/console_logging_spec.rb, line 9 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1258#-M3TEFCw-EDFb_yTzJYO:-M3TXcN02d9hjHKX_I2n:bx30nbs) ([raw file](https://github.com/shakacode/react_on_rails/blob/71e09ebd4868ac8c96c6fa7bab8d9df1cd4874f5/spec/dummy/spec/requests/console_logging_spec.rb#L9)):*

Super...

I remember something about this from a long time ago.

Can you explain why this one was the only one that needed this to keep the indent like this?
</blockquote></details>

Yes, those lines were getting compared in test against expected_lines which doesn't have the spaces starting the line, I guess disabling the rubocop indentation was a better tradeoff to trim those lines of whitespaces.

Issue #1198, opened 2019-03-05
React Hooks (Hooks can only be called inside the body of a function component)


lukaskamp:
I'm trying Testing React Hooks with react_on_rails gem and noticed there is a Error: 

"Hooks can only be called inside the body of a function component error."

Without using ReactOnRails config (e.g in packs/applications.js) everything works correctly. Also when I used react-rails gem everything works. 

Where could be problem ?

amauryfischer:
Got same problem there, i tried a couple of react suggestion(veryfing i'm using hooks correctly, and veryfing that i got a single version of React) but it's not one of them...
Seems we are not alone, don't know if it comes from rails webpacker, React on rails, or anything else.
Related issue on rails webpacker : https://github.com/rails/webpacker/issues/1840

justin808:
@lukaskamp @amauryfischer can one of you give me a simple reproduction case? a github repo?

wuz:
As a note, a quick fix for this is to just wrap your React On Rails rendered component in a function, like this:

```
export default props => <ComponentHere {...props} />;
```

amauryfischer:
```
import React,{useState} from 'react'
const Test = function() {
    const [state, setstate] = useState(0);
    return (
        <div onClick={setstate(state + 1,)}>{state}</div>
    )
}

export default props => <Test {...props} />;
```

@wuz didn't work, same error :
```
Invariant Violation: Hooks can only be called inside the body of a function component.
```

wuz:
@amauryfischer Is that the exact code you are using? That `setstate` call in the button won't work, since that is just getting called on render, causing an infinite loop.

Using `export default props => <Test {...props} />;` worked for me.

amauryfischer:
Yes my bad, but even corrected
```
import React,{useState} from 'react'
const Test = function() {
    const [state, setstate] = useState(0);
    return (
        <div onClick={() => setCount(count + 1)}>{state}</div>
    )
}

export default props => <Test {...props} />;
```

got the same error.
Even with only <div></div> in the return.
This is so strange ><

juliusdelta:
We're having the same issue.

We have multiple components rendering on the same page. We believe this is related to the issue described [in the react docs](https://reactjs.org/warnings/invalid-hook-call-warning.html) about loading 2 instances of React, possibly.

The easiest way to reproduce is to make 2 components with hooks and register them then pass those two into a Rails view. If I have enough time later today I'll make a reproduction repo and update this comment.

[...]

PR #1268, opened 2020-04-16
Major change to fix issues with React Hooks

justin808:
See details in https://github.com/shakacode/react_on_rails/issues/1198

# BREAKING CHANGE

I’m updating React on Rails. The new API for **ReactOnRails.register** will take an object which registers either

1. React functional or class components
2. render-functions, as defined below.

The update concerns render-functions (formerly called "generator functions"). Previously, generator functions would return JSX and there were sort of like a React function component. But not quite, as you could not use React Hooks within them.

With this update, render-functions need to return either a

1. React Function Component
2. React Class Component
3. An object representing server-side rendering output

***Note, you can no longer return JSX!***

If you use React component or create a function that returns a React component, then the React on Rails library will call React.createElement and then call ReactDOM.render (or hydrate).

This ensures that React Hooks can always be used with both the React functional or class components and render-functions.

In the case of the render-function, if you could just return the React element (or JSX) and you used the hook API, then you’d get an obscure “hooks don’t work” error.

Use regular React functional or class component or a render-function for your client-side bundle:

```javascript
// React functional component
export default (props) => (
  <App {...props} />
);
```

Or a render-function. Note you can't return just the JSX (React element), but you need to return either a React functional or class component.
```javascript
// React functional component
export default (props, railsContext) => (
  () => <App {{railsContext, ...props}} />
);
```

Note, this doesn't work, because this function just returns a React Element
```javascript
// React functional component
export default (props, railsContext) => (
  <App {{railsContext, ...props}} />
);
```

While this does result in changing all the cases that looked like a React functional component, but had 2 params, I do believe this is worth it.

#1268  Performance? 

In my update to React on Rails, I’m changing some APIs. I’m wondering if calling a function to create a function that has a closure with say very a large props JS object, and then returning that function and then
the function will again be called by React.createElement(MyComponent, props)…So the props passed to the React function are not really necessary b/c they can be in a closure.

I’m concerned that having the props passed to the real React component will be bad for performance if they already should be used from the closure. However, **JavaScript passes objects by reference, so I don't see any issue.**

This could be optimized by not passing any props (which would go into unused param _props below). However, that’s going to be confusing in some cases. I can see people defining the inner component using the same name props and that would shadow the outer one.

```js
import React from 'react';
const MyAppComponent = (props, railsContext) => (
  // NOTE: need to wrap in a function so this is proper React function component that can use
  // hooks
  // the props get passed again, but we ignore since we use a closure
  // or should we
  (_props) =>
      <div>
        <p>props are: {JSON.stringify(props)}</p>
        <p>railsContext is: {JSON.stringify(railsContext)}
        </p>
      </div>
);
export default MyAppComponent;
```

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1268)


spec/dummy/client/app/startup/DeferredRenderAppServer.jsx
```diff
@@ -34,7 +34,7 @@ const DeferredRenderAppServer = (_props, railsContext) => {
     return { error, redirectLocation };
   }
 
-  return <RouterContext {...routerProps} />;
+  return () => <RouterContext {...routerProps} />;
```
justin808, line 37: This makes sense, @Judahmeek. It needs a React function component, not JSX.

spec/dummy/spec/system/integration_spec.rb
```diff
@@ -66,10 +66,10 @@ def finished_all_ajax_requests?
 
     context "React Hooks" do
       context "Simple stateless component" do
-        include_examples "React Component", "div#HelloWorldApp-react-component-6"
+        include_examples "React Component", "div#HelloWorld-react-component-6"
       end
       context "render function that takes props" do
-        include_examples "React Component", "div#HelloWorldApp-react-component-7"
+        include_examples "React Component", "div#HelloWorld-react-component-7"
```
justin808, line 72: An idea on why this changed?

justin808:
Only pending thing for merge is to double check the CHANGELOG, README, and upgrade docs.

docs/basics/upgrading-react-on-rails.md
```diff
@@ -5,6 +5,54 @@ If you would like help in migrating between React on Rails versions or help with
 
 We specialize in helping companies to quickly and efficiently move from versions before 9 to current. The older versions use the Rails asset pipeline to package client assets. The current and recommended way is to use Webpack 4 for asset preparation. You may also need help migrating from the `rails/webpacker`'s Webpack configuration to a better setup ready for Server Side Rendering.
 
+## Upgrading to v13
+* Make sure that are are on a relatively more recent version of rails and webpacker.
```
tahsin352: There is extra `are`. Please remove.

docs/contributor-info/errors-with-hooks.md
```diff
@@ -0,0 +1,46 @@
...
+
+npm ERR! invalid: react-on-rails@12.0.0 /Users/justin/shakacode/react-on-rails/react_on_rails/spec/dummy/node_modules/react-on-rails
+npm ERR! extraneous: react@16.13.1 /Users/justin/shakacode/react-on-rails/react_on_rails/spec/dummy/node_modules/react-on-rails/node_modules/react
+```
+
+Make sure there is only one version of React installed!
+
+If you used yarn link, then you'll have two versions of React installed.
+
+Instead use Yalc.
```
tahsin352: cool.

CONTRIBUTING.md
```diff
@@ -97,16 +97,26 @@ In addition to testing the Ruby parts out, you can also test the node package pa
 ```sh
 cd react_on_rails/
 yarn
-yarn run build
-yarn install-react-on-rails
+yarn run build:watch
+yalc pubish react-on-rails
```
tahsin352: `publish`

spec/dummy/spec/system/integration_spec.rb
Judahmeek: Those are the Ids that you used in your initial commit: https://github.com/shakacode/react_on_rails/pull/1268/commits/be37b3410de1974f16066951bc41fa73fa6f3c61?file-filters%5B%5D=.erb#diff-13675373e6eaa2b6e91995c3ec8974f7

justin808:
Thanks for the help @Judahmeek!

jacksonrayhamilton:
Glad to see this merged!

Should we expect React on Rails v12 to be released soon?

simalexan:
@justin808 this is great, appreciate your work! 
Just wondering as @jacksonrayhamilton when should we see this published?

justin808:
OMG! Your thank you messages really fire me up! I'll try to get this one by this weekend!

BTW, if any of you can consider even the tiniest sponsorship, that would be so appreciated!

https://github.com/sponsors/shakacode

CONTRIBUTING.md
justin808: Thank you @tahsin352 for your review!!!!

@Judahmeek did we get all of these?

PR #1270, opened 2020-04-16
Add a heroimage and buttons to README.md

earkhipov:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1270)


earkhipov:
@justin808 Where should we store images?

PR #1280, opened 2020-05-05
Allowing a render function to return JSX and console.error a message.

justin808:
to return JSX and detcting that

If you were a user of React on Rails, would you mind the small required changes to ensure correct functionality? In other words, we will always call React.createElement in the library, rather than just taking JSX from some function that took props and was invoked by React on Rails.

See the discussion here: 
https://github.com/shakacode/react_on_rails/issues/1198#issuecomment-623967098

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1280)

### DIFF
--- a/node_package/src/createReactOutput.ts
+++ b/node_package/src/createReactOutput.ts
@@ -45,13 +45,20 @@ export default function createReactOutput({
       // We just return at this point, because calling function knows how to handle this case and
       // we can't call React.createElement with this type of Object.
       return (renderFunctionResult as ServerRenderResult);
-    } // else we'll be calling React.createElement
-    // TODO: Can we detect if this is a React Element or a React Function Component?
-    // If already a ReactElement, then just return it.
+    }
+
+    if (React.isValidElement(renderFunctionResult)) {
+      // If already a ReactElement, then just return it.
+      console.error(
+`Warning: ReactOnRails: Your registered render-function (ReactOnRails.register) for ${name}
+incorrectly returned a React Element (JSX). Instead, return a React Function Component by
+wrapping your JSX in a function. ReactOnRails v13 will error on this, as React Hooks do not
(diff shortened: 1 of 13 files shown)


node_package/src/createReactOutput.ts
```diff
@@ -45,13 +45,15 @@ export default function createReactOutput({
...
-    // TODO: Can we detect if this is a React Element or a React Function Component?
-    // If already a ReactElement, then just return it.
-    // If a component, then wrap in an element
+    }
 
-    const reactComponent = renderFunctionResult as ReactComponent;
+    if (React.isValidElement(renderFunctionResult)) {
+      // If already a ReactElement, then just return it.
+      return renderFunctionResult;
+    }
```
justin808, line 58: Basically, this skips the maybe essential call of React.createElement(component, props).

This 100% breaks if hooks are used.

Should we just error out and force the change to wrap the result JSX in a function with zero params? 

or should we make the upgrade easier for 99% of the cases that did not use Hooks.

justin808:
If you take 2, correct is to use a React Function Component:

```js
import React from 'react';
import AppComponent from './AppComponent';
const AppComponentWithRailsContext = (props, railsContext) => (
  // It's better to create a React Function Component so you can
  // use the React Hooks API in this React Function Component
  () => <AppComponent {...{...props, railsContext}}/>
)
export default AppComponentWithRailsContext
```

Currently, this works, to just return a React Element (JSX)
```js
import React from 'react';
import AppComponent from './AppComponent';
const AppComponentWithRailsContext = (props, railsContext) => (
  <AppComponent {...{...props, railsContext}}/>
)
export default AppComponentWithRailsContext
```

I propose that I'll allow the second case for v12, but print a **warning** console message indicating that the JSX should get wrapped in a function to make the return value a React Function Component.

PR #1287, opened 2020-06-06
Include definitions in node package

Judahmeek:
[DefinitelyTyped recommends including definitions directly in the package when possible](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package), so here are the necessary changes.

I haven't tested consuming these definitions with a new typescript project.

I changed `types/index.d.ts` to `types/index.ts` because [Typescript's compiler doesn't copy definition files.](https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1287)


package.json
```diff
@@ -67,8 +67,8 @@
     "clean": "rm -rf node_package/lib",
     "start": "nps",
     "prepare": "yarn run build",
-    "prepublish": "npm run prepare",
-    "build": "yarn run clean && yarn run tsc",
+    "prepublishOnly": "npm run prepare",
```
Judahmeek: This change isn't relevant to typescript, but stops yarn from running `yarn build` twice after `yarn install`.

PR #1293, opened 2020-06-10
Add HMR via @pmmmwh/react-refresh-webpack-plugin

justin808:
Server Rendering Bundle seems to be including the refresh plugin thus getting the error 

```
rails               | ERROR in SERVER PRERENDERING
rails               | Encountered error: "ReferenceError: window is not defined"
```

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1293)


justin808:
@ashgaliyev @MUTOgen We need to apply the same solution that we applied to our customer in terms of only enabling the plugin for use with the client bundle on the webpack-dev-server.

PR #1294, opened 2020-06-13
Compile JSON locales by default in tests

peterzhu2118:
The Rake task compiles JSON locales by default now, so we should do the same when compiling for tests.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1294)


peterzhu2118:
Hey @justin808, could I get a review on this?

peterzhu2118:
@justin808 I added a test case to address your comment

PR #1298, opened 2020-06-22
Default to rails/webpacker precompile

justin808:
If the webpacker webpack config files exist, then React on Rails
will not override the default assets:precompile setup by rails/webpacker

The fix is to remove the JS files inside of config/webpack,
like config/webpack/production.js.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1298)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ Changes since last non-beta release.
 ## UPCOMING 12.0 RELEASE
 
 #### Improved
-* Generator supports React Hooks
+### [12.0.0.pre.beta.2]
+* Changed the precompile task to use the rails/webpacker one by default
 
 ### [12.0.0.pre.beta.1]
 * Updated generators to use React hooks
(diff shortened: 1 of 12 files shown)


justin808:
@peterzhu2118 would you like to let me know what you think of this change?

@ashgaliyev, @Judahmeek thoughts?

@ashgaliyev this should stop double compilation.

peterzhu2118:
@justin808 I just tried it out, it works great! Maybe out of scope for this PR, but have you considered adding a feature that runs the `react_on_rails:locale` rake task before precompile? It's a little inconvenient to have to run that rake task every time I update the locales.

justin808:
> @justin808 I just tried it out, it works great! Maybe out of scope for this PR, but have you considered adding a feature that runs the `react_on_rails:locale` rake task before precompile? It's a little inconvenient to have to run that rake task every time I update the locales.

1. The older react on rails way does that.
2. Locale conversion is probably not recommended as there are advantages to using the native format from https://formatjs.io/. Keeping the localizations only makes sense if the messages will be used by both the Rails views and also the React code.

So I think it's OK to document that in the locale section.

justin808:
@peterzhu2118 I'm thinking now that the check for the existence of the config/webpack/production.js file is unnecessary and the whole monkey patching of the rake task might be removable if the command is moved to the bin/webpack script since that will be called by the rails/webpacker patch to assets:precompile.

See https://github.com/rails/webpacker/issues/2640.

PR #1299, opened 2020-06-25
Don't compile locales in tests by default

peterzhu2118:
Since we don't compile locales in development and `rails assets:precompile`, then it should be default that locales aren't compiled in tests either. The old behaviour can be brought back by passing in `compile_locales: true` to `ReactOnRails::TestHelper.ensure_assets_compiled`.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1299)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+* `ReactOnRails::TestHelper.ensure_assets_compiled` no longer compiles locales by default.
+  You can compile locales by passing in `compile_locales: true`.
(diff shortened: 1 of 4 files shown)


justin808:
@peterzhu2118 see #1300.

peterzhu2118:
> if using the old way, with a custom command, the i18n rake task is run, and that does nothing unless `config.i18n_dir`.

I haven't realized that the locales aren't compiled when `config.i18n_dir` is not set. I had `config.i18n_dir` set in my initializer, the locales aren't compiled in tests when I remove this configuration. I'm not sure if this PR is useful anymore.

justin808:
Closing because not necessary since there is no performance gain.

PR #1348, opened 2020-12-20
CH6906 - Update CircleCI config to trigger SC website build using CI

MUTOgen:
CH story: [CH6906](https://app.clubhouse.io/shakacode/story/6906/update-circleci-config-to-trigger-sc-website-build-using-ci)

We need to trigger SC website build using CI.
We could use `when` attribute to run an additional step in `Main CI` job.
Also, we could check the current branch to use this login only on `master`.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1348)


MUTOgen:
Env variable
Снимок экрана 2020-12-20 в 21 20 51

PR #1356, opened 2021-02-17
Don't set NODE_ENV in assets.rake

justin808:
Setting NODE_ENV here interferes with Rails'/webpacker's control of this
variable, particularly in production. In particular, this means that
compiling assets in production using
`bundle exec rails assets:precompile` uses the development webpacker
configuration rather than the production one.

1. NODE_ENV is not set when running precompile
2. React on Rails sees it's empty and sets to development
3. rails/webpacker does not apply default of production

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1356)

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -6,7 +6,6 @@
 require "active_support"
 
 ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
-ENV["NODE_ENV"]  ||= "development"
 
 unless ReactOnRails::WebpackerUtils.webpacker_webpack_production_config_exists?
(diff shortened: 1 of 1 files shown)


justin808:
@Judahmeek I think this PR is OK. Can you check and we can release a 12.0.5 if you say it's good.

Judahmeek:
@justin808 LGTM. Should we change line 8 to default to "production" since [webpack will default to production if a mode isn't set](https://webpack.js.org/configuration/mode/)?

justin808:
> @justin808 LGTM. Should we change line 8 to default to "production" since [webpack will default to production if a mode isn't set](https://webpack.js.org/configuration/mode/)?

No. That's for rails_env...and it's fine.

PR #1364, opened 2021-03-24
Move doc file for SC website docs

MUTOgen:
We could place `md` files right in the root `docs` folder, but it's better to keep it organized.
It looks much better in the docs menu on SC website.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1364)


justin808:
There are several references to this file: https://github.com/rails/webpacker/blob/master/README.md

So I'd rather not move it.

PR #1374, opened 2021-05-18
Added support for @hotwired/turbo.

pgruener:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1374)

### DIFF
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -58,6 +58,13 @@ function turbolinksInstalled(): boolean {
   return (typeof Turbolinks !== 'undefined');
 }
 
+function turboInstalled() {
+  var context = findContext();
+  if (context.ReactOnRails) {
+    return context.ReactOnRails.option('turbo') === true;
+  }
+}
+
 function reactOnRailsHtmlElements(): HTMLCollectionOf<Element>  {
   return document.getElementsByClassName('js-react-on-rails-component');
 }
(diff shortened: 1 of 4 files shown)


node_package/src/clientStartup.ts
```diff
@@ -58,6 +58,13 @@ function turbolinksInstalled(): boolean {
   return (typeof Turbolinks !== 'undefined');
 }
 
+function turboInstalled() {
+  var context = findContext();
+  if (context.ReactOnRails) {
+    return context.ReactOnRails.option('turbo') === true;
+  }
+}
```
justin808, line 67: Is a configuration better than detection?

pgruener: No, I also would prefer a detection if possible. But as I noticed Turbo will be included only as module (for webpack(er)) in contrast to Turbolinks, which will inject itself to the global window.

So I don't know a way to inspect the global environment to detect if the Turbo module is used somewhere in the application.. if you have a hint for me, we can adjust that of course.

justin808: Try require within a try block. 

I can get you a cod snippet if needed.

justin808:
@pgruener consider joining [our Slack](https://join.slack.com/t/reactrails/shared_invite/enQtNjY3NTczMjczNzYxLTlmYjdiZmY3MTVlMzU2YWE0OWM0MzNiZDI0MzdkZGFiZTFkYTFkOGVjODBmOWEyYWQ3MzA2NGE1YWJjNmVlMGE)

justin808:
@pgruener Let me know if I should take over your PR or if you have an ETA. I'd like to ship it in the next release.

pgruener:
@justin808 Thanks for your message. You can of course take over, as I didn't get time yet to go on with your suggestion (even if it should only take a few minutes). My best ETA would be around the coming weekend, but I'd really appreciate it to be shipped soon.
I also did join the slack channel, but didn't get a chance to get active there.

justin808:
@pgruener Say 👋 when you get to the Slack. Next weekend would be fine. I've got a ton of other pending tasks. I really appreciate the help!

justin808:
@pgruener Any update?

justin808:
@Judahmeek can you review also?

justin808:
@pgruener can you see if you can fix the CI issues?

justin808:
We need some short docs on this. Maybe we should update this page: https://github.com/shakacode/react_on_rails/blob/9ef6b326213480d0022db42710e5517cad0317bc/docs/rails/turbolinks.md ?

or is turbo different enough that we should create a new doc page?

node_package/src/clientStartup.ts
Judahmeek: The problem with detection logic like that is that it would only detect if turbo is being bundled through `webpacker` even though turbo supports `sprockets` as well.

PR #1380, opened 2021-08-07
WIP: Enable registered components to return promises for server rendering by React on Rails Pro Node Renderer

Judahmeek:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)

### DIFF
--- a/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb
+++ b/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb
@@ -150,7 +150,7 @@ module ReactOnRails
         end
 
         def execjs_timer_polyfills
-          if ReactOnRails::Utils.react_on_rails_pro? && ReactOnRailsPro.configuration.server_renderer == "NodeRenderer"
+          if ReactOnRails::Utils.react_on_rails_pro? && ReactOnRailsPro.configuration.execjs_polyfills == false
             return ""
           end
 
--- a/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb
+++ b/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb
@@ -150,6 +150,10 @@ module ReactOnRails
         end
 
         def execjs_timer_polyfills
+          if ReactOnRails::Utils.react_on_rails_pro? && ReactOnRailsPro.configuration.server_renderer == "NodeRenderer"
+            return ""
+          end
+
           <<~JS
             function getStackTrace () {
               var stack;
--- a/node_package/src/ReactOnRails.ts
+++ b/node_package/src/ReactOnRails.ts
@@ -205,7 +205,7 @@ ctx.ReactOnRails = {
    * Used by server rendering by Rails
    * @param options
    */
-  serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult> {
+  serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
     return serverRenderReactComponent(options);
   },
 
--- a/node_package/src/ReactOnRails.ts
+++ b/node_package/src/ReactOnRails.ts
@@ -13,6 +13,7 @@ import context from './context';
 import type {
   RegisteredComponent,
   RenderParams,
+  RenderResult,
   ErrorOptions,
   ReactComponentOrRenderFunction,
   AuthenticityHeaders,
@@ -204,7 +205,7 @@ ctx.ReactOnRails = {
    * Used by server rendering by Rails
    * @param options
    */
-  serverRenderReactComponent(options: RenderParams): string {
+  serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult> {
     return serverRenderReactComponent(options);
   },
 
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -8,7 +8,7 @@ import type {
 } from './types/index';
 
 import createReactOutput from './createReactOutput';
-import isServerRenderResult from './isServerRenderResult';
+import {isServerRenderHash} from './isServerRenderResult';
 
 declare global {
   interface Window {
@@ -161,7 +161,7 @@ function render(el: Element, railsContext: RailsContext): void {
         shouldHydrate,
       });
 
-      if (isServerRenderResult(reactElementOrRouterResult)) {
+      if (isServerRenderHash(reactElementOrRouterResult)) {
         throw new Error(`\
 You returned a server side type of react-router error: ${JSON.stringify(reactElementOrRouterResult)}
 You should return a React.Component always for the client side entry point.`);
--- a/node_package/src/createReactOutput.ts
+++ b/node_package/src/createReactOutput.ts
@@ -3,7 +3,7 @@
 import React from 'react';
 import type { ServerRenderResult,
   CreateParams, ReactComponent, RenderFunction, CreateReactOutputResult } from './types/index';
-import isServerRenderResult from "./isServerRenderResult";
+import {isServerRenderHash, isPromise} from "./isServerRenderResult";
 
 /**
  * Logic to either call the renderFunction or call React.createElement to get the
@@ -44,10 +44,16 @@ export default function createReactOutput({
       console.log(`${name} is a renderFunction`);
     }
     const renderFunctionResult = (component as RenderFunction)(props, railsContext);
-    if (isServerRenderResult(renderFunctionResult as CreateReactOutputResult)) {
+    if (isServerRenderHash(renderFunctionResult as CreateReactOutputResult)) {
       // We just return at this point, because calling function knows how to handle this case and
       // we can't call React.createElement with this type of Object.
       return (renderFunctionResult as ServerRenderResult);
+    }
+
+    if (isPromise(renderFunctionResult as CreateReactOutputResult)) {
+      // We just return at this point, because calling function knows how to handle this case and
+      // we can't call React.createElement with this type of Object.
+      return (renderFunctionResult as Promise<string>);
     }
 
     if (React.isValidElement(renderFunctionResult)) {
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -57,6 +57,9 @@ ROUTER REDIRECT: ${name} to dom node with id: ${domNodeId}, redirect to ${redire
         renderResult = (reactElementOrRouterResult as { renderedHtml: string }).renderedHtml;
       }
     } else if (isPromise(reactElementOrRouterResult)) {
+      if (!returnPromise) {
+        console.error(`Your render function returned a Promise, which is only supported by a node renderer, not ExceJS.`)
+      }
       renderResult = reactElementOrRouterResult;
     } else {
       try {
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -10,7 +10,7 @@ import handleError from './handleError';
 import type { RenderParams, RenderResult, RenderingError } from './types/index';
 
 export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
-  const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
+  const { name, domNodeId, trace, props, railsContext, renderingReturnsPromises, throwJsErrors } = options;
 
   let renderResult: null | string | Promise<string> = null;
   let hasErrors = false;
@@ -24,7 +24,7 @@ Detected a renderer while server rendering component '${name}'. \
 See https://github.com/shakacode/react_on_rails#renderer-functions`);
     }
 
-    const reactElementOrRouterResult = createReactOutput({
+    const reactRenderingResult = createReactOutput({
       componentObj,
       domNodeId,
       trace,
@@ -32,46 +32,56 @@ See https://github.com/shakacode/react_on_rails#renderer-functions`);
       railsContext,
     });
 
-    if (isServerRenderHash(reactElementOrRouterResult)) {
-      // We let the client side handle any redirect
-      // Set hasErrors in case we want to throw a Rails exception
-      hasErrors = !!(reactElementOrRouterResult as {routeError: Error}).routeError;
+    const processServerRenderHash = () => {
+        // We let the client side handle any redirect
+        // Set hasErrors in case we want to throw a Rails exception
+        hasErrors = !!(reactRenderingResult as {routeError: Error}).routeError;
 
-      if (hasErrors) {
-        console.error(
-          `React Router ERROR: ${JSON.stringify((reactElementOrRouterResult as {routeError: Error}).routeError)}`,
-        );
-      }
-
-      if ((reactElementOrRouterResult as {redirectLocation: {pathname: string; search: string}}).redirectLocation) {
-        if (trace) {
-          const { redirectLocation } = (reactElementOrRouterResult as {redirectLocation: {pathname: string; search: string}});
-          const redirectPath = redirectLocation.pathname + redirectLocation.search;
-          console.log(`\
-ROUTER REDIRECT: ${name} to dom node with id: ${domNodeId}, redirect to ${redirectPath}`,
+        if (hasErrors) {
+          console.error(
+            `React Router ERROR: ${JSON.stringify((reactRenderingResult as {routeError: Error}).routeError)}`,
           );
         }
-        // For redirects on server rendering, we can't stop Rails from returning the same result.
-        // Possibly, someday, we could have the rails server redirect.
-      } else {
-        renderResult = (reactElementOrRouterResult as { renderedHtml: string }).renderedHtml;
+
+        if ((reactRenderingResult as {redirectLocation: {pathname: string; search: string}}).redirectLocation) {
+          if (trace) {
+            const { redirectLocation } = (reactRenderingResult as {redirectLocation: {pathname: string; search: string}});
+            const redirectPath = redirectLocation.pathname + redirectLocation.search;
+            console.log(`\
+  ROUTER REDIRECT: ${name} to dom node with id: ${domNodeId}, redirect to ${redirectPath}`,
+            );
+          }
+          // For redirects on server rendering, we can't stop Rails from returning the same result.
+          // Possibly, someday, we could have the rails server redirect.
+          return '';
+        }
+        return (reactRenderingResult as { renderedHtml: string }).renderedHtml;
+    };
+
+    const processPromise = () => {
+      if (!renderingReturnsPromises) {
+        console.error('Your render function returned a Promise, which is only supported by a node renderer, not ExecJS.')
       }
-    } else if (isPromise(reactElementOrRouterResult)) {
-      if (!returnPromise) {
-        console.error(`Your render function returned a Promise, which is only supported by a node renderer, not ExceJS.`)
-      }
-      renderResult = reactElementOrRouterResult;
-    } else {
+      return reactRenderingResult;
+    }
+
+    const processReactElement = () => {
       try {
-        renderResult = ReactDOMServer.renderToString(reactElementOrRouterResult as ReactElement);
+        return ReactDOMServer.renderToString(reactRenderingResult as ReactElement);
       } catch (error) {
-        console.error(
-            `Invalid call to renderToString. Possibly you have a renderFunction, a function that already
-calls renderToString, that takes one parameter. You need to add an extra unused
-parameter to identify this function as a renderFunction and not a simple React
-Function Component.`);
+        console.error(`Invalid call to renderToString. Possibly you have a renderFunction, a function that already
+calls renderToString, that takes one parameter. You need to add an extra unused parameter to identify this function
+as a renderFunction and not a simple React Function Component.`);
         throw error;
       }
+    };
+
+    if (isServerRenderHash(reactRenderingResult)) {
+      renderResult = processServerRenderHash();
+    } else if (isPromise(reactRenderingResult)) {
+      renderResult = processPromise() as Promise<string>;
+    } else {
+      renderResult = processReactElement();
     }
   } catch (e) {
     if (throwJsErrors) {
@@ -95,13 +105,31 @@ Function Component.`);
     };
   }
 
-  if(returnPromise) {
+  if(renderingReturnsPromises) {
     const resolveRenderResult = async () => {
-      const promiseResult = {
-        html: await renderResult,
-        consoleReplayScript,
-        hasErrors,
-      };
+      let promiseResult;
+
+      try {
+        promiseResult = {
+          html: await renderResult,
+          consoleReplayScript,
+          hasErrors,
+        };
+      } catch (e) {
+        if (throwJsErrors) {
+          throw e;
+        }
+        promiseResult = {
+          html: handleError({
+            e,
+            name,
+            serverSide: true,
+          }),
+          consoleReplayScript,
+          hasErrors: true,
+        }
+        renderingError = e;
+      }
 
       if (renderingError !== null) {
         addRenderingErrors(promiseResult, renderingError);
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -1,20 +1,20 @@
 import ReactDOMServer from 'react-dom/server';
-import type { ReactElement } from 'react';
+import type { ReactElement, Component } from 'react';
 
 import ComponentRegistry from './ComponentRegistry';
 import createReactOutput from './createReactOutput';
-import isCreateReactElementResultNonReactComponent from
+import {isServerRenderHash, isPromise} from
     './isServerRenderResult';
 import buildConsoleReplay from './buildConsoleReplay';
 import handleError from './handleError';
 import type { RenderParams, RenderResult } from './types/index';
 
-export default function serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult> {
+export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
   const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
 
-  let htmlResult = '';
+  let renderResult: null | string | Promise<string> = null;
   let hasErrors = false;
-  let renderingError = null;
+  let renderingError: any = null;
 
   try {
     const componentObj = ComponentRegistry.get(name);
@@ -32,7 +32,7 @@ See https://github.com/shakacode/react_on_rails#renderer-functions`);
       railsContext,
     });
 
-    if (isCreateReactElementResultNonReactComponent(reactElementOrRouterResult)) {
+    if (isServerRenderHash(reactElementOrRouterResult)) {
       // We let the client side handle any redirect
       // Set hasErrors in case we want to throw a Rails exception
       hasErrors = !!(reactElementOrRouterResult as {routeError: Error}).routeError;
@@ -54,11 +54,13 @@ ROUTER REDIRECT: ${name} to dom node with id: ${domNodeId}, redirect to ${redire
         // For redirects on server rendering, we can't stop Rails from returning the same result.
         // Possibly, someday, we could have the rails server redirect.
       } else {
-        htmlResult = (reactElementOrRouterResult as { renderedHtml: string }).renderedHtml;
+        renderResult = (reactElementOrRouterResult as { renderedHtml: string }).renderedHtml;
       }
+    } else if (isPromise(reactElementOrRouterResult)) {
+      renderResult = reactElementOrRouterResult;
     } else {
       try {
-        htmlResult = ReactDOMServer.renderToString(reactElementOrRouterResult as ReactElement);
+        renderResult = ReactDOMServer.renderToString(reactElementOrRouterResult as ReactElement);
       } catch (error) {
         console.error(
             `Invalid call to renderToString. Possibly you have a renderFunction, a function that already
@@ -74,7 +76,7 @@ Function Component.`);
     }
 
     hasErrors = true;
-    htmlResult = handleError({
+    renderResult = handleError({
       e,
       name,
       serverSide: true,
@@ -83,23 +85,41 @@ Function Component.`);
   }
 
   const consoleReplayScript = buildConsoleReplay();
+  const addRenderingErrors = (resultObject: RenderResult) => {
+    resultObject.renderingError = {
+      message: renderingError.message,
+      stack: renderingError.stack,
+    };
+    return resultObject;
+  }
+
+  if(returnPromise) {
+    const resolveRenderResult = async () => {
+      const promiseResult = {
+        html: await renderResult,
+        consoleReplayScript,
+        hasErrors,
+      };
+
+      if (renderingError) {
+        addRenderingErrors(promiseResult);
+      };
+
+      return promiseResult;
+    };
+
+    return resolveRenderResult();
+  }
 
   const result = {
-    html: htmlResult,
+    html: renderResult,
     consoleReplayScript,
     hasErrors,
   } as RenderResult;
 
   if (renderingError) {
-    result.renderingError = {
-      message: renderingError.message,
-      stack: renderingError.stack,
-    }
-  }
-
-  if(returnPromise) {
-    return Promise.resolve(result);
-  }
+    addRenderingErrors(result);
+  };
 
   return JSON.stringify(result);
 }
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -9,8 +9,8 @@ import buildConsoleReplay from './buildConsoleReplay';
 import handleError from './handleError';
 import type { RenderParams, RenderResult } from './types/index';
 
-export default function serverRenderReactComponent(options: RenderParams): string {
-  const { name, domNodeId, trace, props, railsContext, throwJsErrors } = options;
+export default function serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult> {
+  const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
 
   let htmlResult = '';
   let hasErrors = false;
@@ -63,7 +63,7 @@ ROUTER REDIRECT: ${name} to dom node with id: ${domNodeId}, redirect to ${redire
         console.error(
             `Invalid call to renderToString. Possibly you have a renderFunction, a function that already
 calls renderToString, that takes one parameter. You need to add an extra unused
-parameter to identify this function as a renderFunction and not a simple React 
+parameter to identify this function as a renderFunction and not a simple React
 Function Component.`);
         throw error;
       }
@@ -95,6 +95,10 @@ Function Component.`);
       message: renderingError.message,
       stack: renderingError.stack,
     }
+  }
+
+  if(returnPromise) {
+    return Promise.resolve(result);
   }
 
   return JSON.stringify(result);
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -1,5 +1,5 @@
 import ReactDOMServer from 'react-dom/server';
-import type { ReactElement, Component } from 'react';
+import type { ReactElement } from 'react';
 
 import ComponentRegistry from './ComponentRegistry';
 import createReactOutput from './createReactOutput';
@@ -7,14 +7,14 @@ import {isServerRenderHash, isPromise} from
     './isServerRenderResult';
 import buildConsoleReplay from './buildConsoleReplay';
 import handleError from './handleError';
-import type { RenderParams, RenderResult } from './types/index';
+import type { RenderParams, RenderResult, RenderingError } from './types/index';
 
 export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
   const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
 
   let renderResult: null | string | Promise<string> = null;
   let hasErrors = false;
-  let renderingError: any = null;
+  let renderingError: null | RenderingError = null;
 
   try {
     const componentObj = ComponentRegistry.get(name);
@@ -85,12 +85,11 @@ Function Component.`);
   }
 
   const consoleReplayScript = buildConsoleReplay();
-  const addRenderingErrors = (resultObject: RenderResult) => {
-    resultObject.renderingError = {
-      message: renderingError.message,
-      stack: renderingError.stack,
+  const addRenderingErrors = (resultObject: RenderResult, renderError: RenderingError) => {
+    resultObject.renderingError = { // eslint-disable-line no-param-reassign
+      message: renderError.message,
+      stack: renderError.stack,
     };
-    return resultObject;
   }
 
   if(returnPromise) {
@@ -101,9 +100,9 @@ Function Component.`);
         hasErrors,
       };
 
-      if (renderingError) {
-        addRenderingErrors(promiseResult);
-      };
+      if (renderingError !== null) {
+        addRenderingErrors(promiseResult, renderingError);
+      }
 
       return promiseResult;
     };
@@ -118,8 +117,8 @@ Function Component.`);
   } as RenderResult;
 
   if (renderingError) {
-    addRenderingErrors(result);
-  };
+    addRenderingErrors(result, renderingError);
+  }
 
   return JSON.stringify(result);
 }
--- a/node_package/src/types/index.ts
+++ b/node_package/src/types/index.ts
@@ -39,9 +39,9 @@ interface ServerRenderResult {
   error?: Error;
 }
 
-type CreateReactOutputResult = ServerRenderResult | ReactElement;
+type CreateReactOutputResult = ServerRenderResult | ReactElement | Promise<string>;
 
-type RenderFunctionResult = ReactComponent | ServerRenderResult;
+type RenderFunctionResult = ReactComponent | ServerRenderResult | Promise<string>;
 
 interface RenderFunction {
   (props?: Record<string, unknown>, railsContext?: RailsContext, domNodeId?: string): RenderFunctionResult;
@@ -106,7 +106,7 @@ export interface RenderingError {
 }
 
 export interface RenderResult {
-  html: string;
+  html: string | null;
   consoleReplayScript: string;
   hasErrors: boolean;
   renderingError?: RenderingError;
@@ -128,7 +128,7 @@ export interface ReactOnRails {
     name: string, props: Record<string, string>, domNodeId: string, hydrate: boolean
   ): void | Element | Component;
   getComponent(name: string): RegisteredComponent;
-  serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult>;
+  serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult>;
   handleError(options: ErrorOptions): string | undefined;
   buildConsoleReplay(): string;
   registeredComponents(): Map<string, RegisteredComponent>;
--- a/node_package/src/types/index.ts
+++ b/node_package/src/types/index.ts
@@ -80,6 +80,7 @@ interface Params {
 export interface RenderParams extends Params {
   name: string;
   throwJsErrors: boolean;
+  returnPromise: boolean;
 }
 
 export interface CreateParams extends Params {
@@ -127,7 +128,7 @@ export interface ReactOnRails {
     name: string, props: Record<string, string>, domNodeId: string, hydrate: boolean
   ): void | Element | Component;
   getComponent(name: string): RegisteredComponent;
-  serverRenderReactComponent(options: RenderParams): string;
+  serverRenderReactComponent(options: RenderParams): string | Promise<RenderResult>;
   handleError(options: ErrorOptions): string | undefined;
   buildConsoleReplay(): string;
   registeredComponents(): Map<string, RegisteredComponent>;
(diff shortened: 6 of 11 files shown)


lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb
```diff
@@ -150,6 +150,10 @@ def create_js_context
         end
 
         def execjs_timer_polyfills
+          if ReactOnRails::Utils.react_on_rails_pro? && ReactOnRailsPro.configuration.server_renderer == "NodeRenderer"
+            return ""
+          end
```
Judahmeek, line 155: Removes all execjs_timer_polyfills if node rendering.

node_package/src/ReactOnRails.ts
```diff
@@ -204,7 +205,7 @@ ctx.ReactOnRails = {
    * Used by server rendering by Rails
    * @param options
    */
-  serverRenderReactComponent(options: RenderParams): string {
+  serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
```
Judahmeek, line 208: The types here don't actually match index.js. I'm not sure why TS compiler didn't complain. I'll fix in the morning.

node_package/src/types/index.ts
```diff
@@ -105,7 +106,7 @@ export interface RenderingError {
 }
 
 export interface RenderResult {
-  html: string;
+  html: string | null;
```
Judahmeek, line 109: Because await statements have a return type of optional `null`

node_package/src/serverRenderReactComponent.ts
```diff
@@ -83,18 +85,39 @@ Function Component.`);
...
+    resultObject.renderingError = { // eslint-disable-line no-param-reassign
+      message: renderingError.message,
+      stack: renderingError.stack,
+    };
+  }
+
+  if(returnPromise) {
+    const resolveRenderResult = async () => {
+      const promiseResult = {
+        html: await renderResult,
```
Judahmeek: The entire `promiseResult` gets returned by this async function as a promise, but an `await` statement doesn't seem to resolve internal promises, so I resolve the promise we get from the react component here.

node_package/src/serverRenderReactComponent.ts
```diff
@@ -83,18 +85,39 @@ Function Component.`);
   }
 
   const consoleReplayScript = buildConsoleReplay();
+  const addRenderingErrors = (resultObject: RenderResult, renderingError: RenderingError) => {
+    resultObject.renderingError = { // eslint-disable-line no-param-reassign
+      message: renderingError.message,
+      stack: renderingError.stack,
+    };
+  }
```
Judahmeek, line 106: I put this logic in a function so this logic can be used inside the async function as well as normally.

node_package/src/createReactOutput.ts
```diff
@@ -44,12 +44,18 @@ export default function createReactOutput({
...
       // We just return at this point, because calling function knows how to handle this case and
       // we can't call React.createElement with this type of Object.
       return (renderFunctionResult as ServerRenderResult);
     }
 
+    if (isPromise(renderFunctionResult as CreateReactOutputResult)) {
+      // We just return at this point, because calling function knows how to handle this case and
+      // we can't call React.createElement with this type of Object.
+      return (renderFunctionResult as Promise<string>);
+    }
```
Judahmeek, line 57: Same logic as before, just with different types to make TypeScript happy

node_package/src/clientStartup.ts
```diff
@@ -8,7 +8,7 @@ import type {
 } from './types/index';
 
 import createReactOutput from './createReactOutput';
-import isServerRenderResult from './isServerRenderResult';
+import {isServerRenderHash} from './isServerRenderResult';
```
justin808, line 11: linting should put in some spaces

justin808 (changes requested):
> Reviewed 7 of 7 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 8 unresolved discussions (waiting on @Judahmeek)

___
*[-- commits, line 12 at r3](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgoGgqr9OU1Gpzo0H9t:-MgoGgqr9OU1Gpzo0H9u:btcc0ao) ([raw file](https://github.com/shakacode/react_on_rails/blob/a754f557e88aea05ca8386af51c76779f4a62d2b/-- commits#L12)):*
What is this file?
___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 155 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPo-r1-155:-MgoImY3-uGSRc20QT-M:bhwr9tf) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L155)):*

Removes all execjs_timer_polyfills if node rendering.
</blockquote></details>

This needs to be clearly documented in the changelog.

And major version bump? Not sure, since it requires React on Rails Pro for htis.
___
*[node_package/src/createReactOutput.ts, line 57 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgjavUGc8TlUi9T5sHn-r1-57:-MgoNCAUASLgij_0fHk5:b32w0x) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/node_package/src/createReactOutput.ts#L57)):*

Same logic as before, just with different types to make TypeScript happy
</blockquote></details>

OK
___
*[node_package/src/serverRenderReactComponent.ts, line 93 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPq-r1-93:-MgoMK6q3qeiKGfgujoN:b32w0x) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/node_package/src/serverRenderReactComponent.ts#L93)):*

I put this logic in a function so this logic can be used inside the async function as well as normally.
</blockquote></details>

OK
___
*[node_package/src/serverRenderReactComponent.ts, line 98 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPq-r1-98:-MgoLxP_Epp8OOiuvhBU:b8b4qhy) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/node_package/src/serverRenderReactComponent.ts#L98)):*

The entire `promiseResult` gets returned by this async function as a promise, but an `await` statement doesn't seem to resolve internal promises, so I resolve the promise we get from the react component here.
</blockquote></details>

strange, but OK
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r3](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgoLqvW7rirFYIai7Ah:-MgoLqvW7rirFYIai7Ai:b-tzi3tx) ([raw file](https://github.com/shakacode/react_on_rails/blob/a754f557e88aea05ca8386af51c76779f4a62d2b/node_package/src/serverRenderReactComponent.ts#L13)):*
> ```XML
> 
> export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
>   const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
> ```

Who sets "returnPromise" option?
___
*[node_package/src/types/index.ts, line 109 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPr-r1-109:-MgoNDYz7ekfzrfZRuaG:b32w0x) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/node_package/src/types/index.ts#L109)):*

Because await statements have a return type of optional `null`
</blockquote></details>

OK

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 4 unresolved discussions (waiting on @justin808)

___
*[-- commits, line 12 at r3](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgoGgqr9OU1Gpzo0H9t:-MhHL9bF9qoVy-Nv19s4:b19xy9w) ([raw file](https://github.com/shakacode/react_on_rails/blob/a754f557e88aea05ca8386af51c76779f4a62d2b/-- commits#L12)):*

What is this file?
</blockquote></details>

It's a new Reviewable feature. They should change their code so that they aren't reusing the same component as they use for their files so people don't get confused like you did.
___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 155 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPo-r1-155:-MhHLVIT9gnsjEE3cUIO:b-5122tg) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L155)):*

This needs to be clearly documented in the changelog.

And major version bump? Not sure, since it requires React on Rails Pro for htis.
</blockquote></details>

A major version bump wouldn't hurt.
___
*[node_package/src/clientStartup.ts, line 11 at r3](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgjavUGc8TlUi9T5sHm-r3-11:-MhHN9558AWW0_k0eiPG:bqne2ax) ([raw file](https://github.com/shakacode/react_on_rails/blob/a754f557e88aea05ca8386af51c76779f4a62d2b/node_package/src/clientStartup.ts#L11)):*

linting should put in some spaces
</blockquote></details>

Do you want me to add that linting change in this PR or another PR?
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r3](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgoLqvW7rirFYIai7Ah:-MhHMr6TAYJ5Ksd27K4F:b5olp44) ([raw file](https://github.com/shakacode/react_on_rails/blob/a754f557e88aea05ca8386af51c76779f4a62d2b/node_package/src/serverRenderReactComponent.ts#L13)):*

Who sets "returnPromise" option?
</blockquote></details>

`ReactOnRailsPro::ServerRenderingJsCode.render` will once https://github.com/shakacode/react_on_rails_pro/pull/210 is merged

node_package/src/clientStartup.ts
justin808: Let’s do linting right after this merges. I’m surprised it’s not already there!

justin808 (approved):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 1 unresolved discussion (waiting on @Judahmeek)

___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 155 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPo-r1-155:-MhIsnix5aiOJSwlaenJ:b-e3hhwy) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L155)):*

A major version bump wouldn't hurt.
</blockquote></details>

Please add the changelog entry.

I do think we'd be better we could toggle this, then it's a minor feature bump.

justin808 (approved):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 1 unresolved discussion (waiting on @Judahmeek)

___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 155 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPo-r1-155:-MhItOX95gaIgCVru5Kz:blaet5r) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L155)):*

Please add the changelog entry.

I do think we'd be better we could toggle this, then it's a minor feature bump.
</blockquote></details>

Actually, not a major version bump for React on Rails, but one for React on Rails Pro, but no, let's make this configurable.

justin808 (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 8 unresolved discussions (waiting on @Judahmeek)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-Mhr6lJgBGOueQLrQ65d:b-3h78ac) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*
> ```Markdown
> #### Added
> 
> - Ability to skip stubbing of setTimeout, setInterval, & clearTimeout polyfills when using properly configured React on Rails Pro. Also, ability to have render functions return a promise to be awaited by React on Rails Pro Node Renderer. [PR 1380](https://github.com/shakacode/react_on_rails/pull/1380) by [judahmeek](https://github.com/judahmeek)
> ```

Needs summary here on the configuration change.


Needs changes to doc files.
___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 153 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr3nhYDlzYbaECQHJm:-Mhr3nhYDlzYbaECQHJn:b-vwexj9) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L153)):*
not operator?

> ```
> ReactOnRailsPro.configuration.execjs_polyfills
> ```

```ruby
 if ReactOnRails::Utils.react_on_rails_pro? && !ReactOnRailsPro.configuration.execjs_polyfills
```
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr5qn51XsJGKQt3tqE:-Mhr5qn51XsJGKQt3tqF:b-vvew4k) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*
> ```XML
> 
> export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
>   const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
> ```

can we find a more descriptive name than `returnPromise`?

maybe `rendererReturnsPromises`

I'm not sure where this comes from.

if we have to call the rendering helper with knowledge of the return type, could we use `returnsPromise`

also, could this be just inferred? so we don't need an option on the helper?
___
*[node_package/src/serverRenderReactComponent.ts, line 35 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr60ixErG43uHumL5J:-Mhr60ixErG43uHumL5K:b-cxpok3) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L35)):*
> ```XML
>     });
> 
>     if (isServerRenderHash(reactElementOrRouterResult)) {
> ```

Should we consider checking if a promise first?
___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-Mhr7pQZ960lnLxnKTYa:b-ta0jnk) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*
> ```XML
>         renderResult = (reactElementOrRouterResult as { renderedHtml: string }).renderedHtml;
>       }
>     } else if (isPromise(reactElementOrRouterResult)) {
> ```

Maybe the check for this error condition can be encapsulated in a separate function and checked a bit earlier.

this giant if is hard to follow.
___
*[node_package/src/serverRenderReactComponent.ts, line 60 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6Jmz6-tOL9dLzkgM:-Mhr6Jmz6-tOL9dLzkgN:b38lvgr) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L60)):*
> ```XML
>       }
>     } else if (isPromise(reactElementOrRouterResult)) {
>       if (!returnPromise) {
> ```

maybe `rendererReturnsPromises`
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-Mhr54MgAh5RAJcvcgCo:b86nmpt) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*
> ```XML
>     const resolveRenderResult = async () => {
>       const promiseResult = {
>         html: await renderResult,
> ```

let's have this await be its own statement with its own catch block with error handling

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 7 unresolved discussions (waiting on @justin808)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-MhrFAZBD310KihIxqOg:b-6mszu) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*

Needs summary here on the configuration change.


Needs changes to doc files.
</blockquote></details>

I document the configuration change in the React on Rails Pro PR.
___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 155 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MgWu81GJYWmEjdXlOPo-r1-155:-MhrFsMzBwYIwBhKYLwW:bb6un1j) ([raw file](https://github.com/shakacode/react_on_rails/blob/ffe12e333910a9ac7e9b422e1cba33a5dbb7ce58/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L155)):*

Actually, not a major version bump for React on Rails, but one for React on Rails Pro, but no, let's make this configurable.
</blockquote></details>

Done
___
*[lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb, line 153 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr3nhYDlzYbaECQHJm:-MhrFkom4omzYI-PzpOx:brj4q4g) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb#L153)):*

not operator?

> ```
> ReactOnRailsPro.configuration.execjs_polyfills
> ```

```ruby
 if ReactOnRails::Utils.react_on_rails_pro? && !ReactOnRailsPro.configuration.execjs_polyfills
```
</blockquote></details>

Can do
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr5qn51XsJGKQt3tqE:-MhrG030CvnlRvaS6d6N:b-5fqb93) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*

can we find a more descriptive name than `returnPromise`?

maybe `rendererReturnsPromises`

I'm not sure where this comes from.

if we have to call the rendering helper with knowledge of the return type, could we use `returnsPromise`

also, could this be just inferred? so we don't need an option on the helper?

</blockquote></details>

The parameter comes from React on Rails Pro server_render_js_code.

I called the parameter `returnPromise` because the parameter is telling the `serverRenderReactComponent` to return a promise.

We could infer this by having `serverRenderReactComponent return a promise whenever the NodeRenderer is being used instead.
___
*[node_package/src/serverRenderReactComponent.ts, line 35 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr60ixErG43uHumL5J:-MhrJUZL1MPUfPGUHMhM:b-349zs1) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L35)):*

Should we consider checking if a promise first?
</blockquote></details>

See below
___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-MhrK5WN-gf4b5frCM2I:b-usbd4x) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*

Maybe the check for this error condition can be encapsulated in a separate function and checked a bit earlier.

this giant if is hard to follow.
</blockquote></details>

I'll see what I can do to make all this conditional logic easier to read & comprehend.

I think I'll wrap the inner logic inside functions, so you'll just see:

```javascript
    if (isServerRenderHash(reactElementOrRouterResult)) {
      renderResult = processServerHash(reactElementOrRouterResult)
    } else if (isPromise(reactElementOrRouterResult)) {
      renderResult = processPromise(reactElementOrRouterResult);
    } else {
      renderResult = renderReactElement(reactElementOrRouterResult)
    }
```
___
*[node_package/src/serverRenderReactComponent.ts, line 60 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6Jmz6-tOL9dLzkgM:-MhrIJrh1JNIhY0ZKHvH:baxmpws) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L60)):*

maybe `rendererReturnsPromises`
</blockquote></details>

No, the parameter is telling the `serverRenderReactComponent` function to return a promise.

If this function isn't returning a promise, however, then it's going to use `JSON.stringify` on whatever result it got from the render function, so I'm logging an error here.
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-MhrJ2c17xvPP11J5s3C:bmcine8) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

let's have this await be its own statement with its own catch block with error handling
</blockquote></details>

How should the error be handled?

Should I use the same logic from above?

```javascript
} catch (e) {
    if (throwJsErrors) {
      throw e;
    }

    hasErrors = true;
    renderResult = handleError({
      e,
      name,
      serverSide: true,
    });
    renderingError = e;
  }
```

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 7 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-MhrNnoJ5aefXfgo8Iey:b-m3rrhz) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*

I document the configuration change in the React on Rails Pro PR.
</blockquote></details>

We can mention that here in the CHANGELOG and indicate that it's not relevant to the non-pro version.
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr5qn51XsJGKQt3tqE:-MhrO2_i5vCL89rU1O1N:b-jbeemf) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*

The parameter comes from React on Rails Pro server_render_js_code.

I called the parameter `returnPromise` because the parameter is telling the `serverRenderReactComponent` to return a promise.

We could infer this by having `serverRenderReactComponent return a promise whenever the NodeRenderer is being used instead.
</blockquote></details>

I think that's a much better solution. Many APIs work like this. AFAIK, there's no performance hit given that the majority of the work is doing the SSR.
___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-MhrOXrNCa3rgoAjTUHs:b-5m20ds) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*

I'll see what I can do to make all this conditional logic easier to read & comprehend.

I think I'll wrap the inner logic inside functions, so you'll just see:

```javascript
    if (isServerRenderHash(reactElementOrRouterResult)) {
      renderResult = processServerHash(reactElementOrRouterResult)
    } else if (isPromise(reactElementOrRouterResult)) {
      renderResult = processPromise(reactElementOrRouterResult);
    } else {
      renderResult = renderReactElement(reactElementOrRouterResult)
    }
```
</blockquote></details>

This might become simpler if we always return a promise, right?

And yes, much better.

I think the name `reactElementOrRouterResult` can be shorted to `reactRenderingResult` as the current name does not reflect it might be a promise.
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-MhrP2in1EaBF_clTrWC:bey9pt7) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

How should the error be handled?

Should I use the same logic from above?

```javascript
} catch (e) {
    if (throwJsErrors) {
      throw e;
    }

    hasErrors = true;
    renderResult = handleError({
      e,
      name,
      serverSide: true,
    });
    renderingError = e;
  }
```
</blockquote></details>

Yes, that should work.

We might need to rethrow. Check the logic of the other blocks.

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 9 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRSz9Erkebt4FPUDX:-MhrRSz9Erkebt4FPUDY:bih0xu4) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*
> ```XML
> 
> export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
>   const { name, domNodeId, trace, props, railsContext, returnPromise, throwJsErrors } = options;
> ```

How about this for the new name:

```js
renderingReturnsPromises
```

And it's set on the React on Rails Pro side.
___
*[node_package/src/serverRenderReactComponent.ts, line 61 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRmKk2Bd5TEj0FY_u:-MhrRmKk2Bd5TEj0FY_v:b-9h0utx) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L61)):*
> ```XML
>     } else if (isPromise(reactElementOrRouterResult)) {
>       if (!returnPromise) {
>         console.error(`Your render function returned a Promise, which is only supported by a node renderer, not ExceJS.`)
> ```

This logic implies that returnPromise really means `renderedWithNodeRenderer`

I think changing the name to `renderingReturnsPromises` and that's optionally set from ReactOnRailsPro will work.

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 9 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-MhrRdDf4u4sNU1URyG4:b-v5d2vx) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*

We can mention that here in the CHANGELOG and indicate that it's not relevant to the non-pro version.
</blockquote></details>

Something like this perhaps?
```
Ability to stop stubbing of setTimeout, setInterval, & clearTimeout conditional on NEW execjs_polyfills React on Rails Pro configuration option.
```
___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-MhrS6EZBd0cjqH6wLoq:b-l6o1zt) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*
> This might become simpler if we always return a promise, right?

No. If using ExecJS, the result from `serverRenderReactComponent` turns into a string & promises don't serialize.
___
*[node_package/src/serverRenderReactComponent.ts, line 61 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRmKk2Bd5TEj0FY_u:-MhrU57V0G8VXUTURD0f:b-z2a0mw) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L61)):*

This logic implies that returnPromise really means `renderedWithNodeRenderer`

I think changing the name to `renderingReturnsPromises` and that's optionally set from ReactOnRailsPro will work.
</blockquote></details>

Yes, that implication is correct. `rendererAcceptsPromises` is another way you could phrase it.

Do you want an optional parameter from ReactOnRailsPro or do you want to check if `ReactOnRailsPro.configuration.server_renderer` == "NodeRenderer"?

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 9 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[node_package/src/serverRenderReactComponent.ts, line 61 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRmKk2Bd5TEj0FY_u:-MhrYN0rEwTQmmE-tXOy:bwm2oyx) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L61)):*
> Ability to stop stubbing of setTimeout, setInterval, & clearTimeout conditional on NEW execjs_polyfills React on Rails Pro configuration option.

Let's use an additional parameter. I don't want to change the current behavior for all current installations.

```ruby
ReactOnRailsPro.renderingReturnsPromises = true
```

Default is false.

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 7 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-MhrYmUu54tP4DXwmjpb:bvhts3z) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*
> NEW execjs_polyfills React on Rails Pro configuration option
I don't understand the `execjs_polyfills` name. 

Maybe it should be `include_execjs_polyfills` 


Ability to stop stubbing of setTimeout, setInterval, & clearTimeout conditional by setting `ReactOnRailsPro.config.include_execjs_polyfills = false` in the React on Rails Pro configuration file.

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: 7 of 12 files reviewed, 7 unresolved discussions (waiting on @justin808)

___
*[CHANGELOG.md, line 22 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr6lJgBGOueQLrQ65c:-MhvdXoK-dvBAzTzMhp_:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/CHANGELOG.md#L22)):*

> NEW execjs_polyfills React on Rails Pro configuration option
I don't understand the `execjs_polyfills` name. 

Maybe it should be `include_execjs_polyfills` 


Ability to stop stubbing of setTimeout, setInterval, & clearTimeout conditional by setting `ReactOnRailsPro.config.include_execjs_polyfills = false` in the React on Rails Pro configuration file.
</blockquote></details>

Done.
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr5qn51XsJGKQt3tqE:-MhvkaKN0x80Z9zKURaD:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*

I think that's a much better solution. Many APIs work like this. AFAIK, there's no performance hit given that the majority of the work is doing the SSR.
</blockquote></details>

Done.
___
*[node_package/src/serverRenderReactComponent.ts, line 13 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRSz9Erkebt4FPUDX:-Mhvk4zR6YlhvGM6VABu:bb6un1j) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L13)):*

How about this for the new name:

```js
renderingReturnsPromises
```

And it's set on the React on Rails Pro side.
</blockquote></details>

Done
___
*[node_package/src/serverRenderReactComponent.ts, line 61 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MhrRmKk2Bd5TEj0FY_u:-MhvpM3I9bCbhzg6GlyW:b-896fix) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L61)):*

> Ability to stop stubbing of setTimeout, setInterval, & clearTimeout conditional on NEW execjs_polyfills React on Rails Pro configuration option.

Let's use an additional parameter. I don't want to change the current behavior for all current installations.

```ruby
ReactOnRailsPro.renderingReturnsPromises = true
```

Default is false.
</blockquote></details>

Done.
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-Mhvse_z0gWOIa30IVVP:beixtey) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

Yes, that should work.

We might need to rethrow. Check the logic of the other blocks.
</blockquote></details>

Done.

Btw, the current logic here just awaits & then submits the resolved value, which doesn't match [the logic we use for normal serverRendering](https://github.com/shakacode/react_on_rails/blob/e5b1b6a882a1ea1bc9279f004c649369e38f8996/node_package/src/serverRenderReactComponent.ts#L57)

This is why I made [these changes](https://github.com/shakacode/react_on_rails_pro/pull/210/files?file-filters%5B%5D=.jsx#diff-31b7c574eb973f57630217425f3718a6794b197a967fd4cff5e87e5c2f9e04f2) for my original proof of concept.

We definitely need a more robust solution.

Should we await the renderResult & then run [the conditional logic](https://github.com/shakacode/react_on_rails/blob/judahmeek/ssr-promise/node_package/src/serverRenderReactComponent.ts#L79-L89) on the resolved value just in case the client is using both react-router & a graphql library (in which case we might have a promise resolve a redirect or routeError)?

justin808 (changes requested):
@judahmeek Getting close

We need to pair on the promises block of code.

> Reviewed 3 of 5 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-:-MiBonoBDq4eRUcaPFzr:b-aiiryd)* status: all files reviewed, 5 unresolved discussions (waiting on @Judahmeek and @justin808)

___
*[node_package/babel.config.js, line 2 at r6](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MiBmyIO0HEu7R2C8Nzi:-MiBmyIO0HEu7R2C8Nzj:b-9yhw07) ([raw file](https://github.com/shakacode/react_on_rails/blob/97d749d17007a30dcce19d674baf9a4a58221d3a/node_package/babel.config.js#L2)):*
> ```JavaScript
> module.exports = {
>   presets: [['@babel/preset-env', { targets: { node: 'current' } }]],
> ```

@judahmeek, what motivated this one? 
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-MiBo1xt8oYfwWSXleDh:b-1q45b) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

Done.

Btw, the current logic here just awaits & then submits the resolved value, which doesn't match [the logic we use for normal serverRendering](https://github.com/shakacode/react_on_rails/blob/e5b1b6a882a1ea1bc9279f004c649369e38f8996/node_package/src/serverRenderReactComponent.ts#L57)

This is why I made [these changes](https://github.com/shakacode/react_on_rails_pro/pull/210/files?file-filters%5B%5D=.jsx#diff-31b7c574eb973f57630217425f3718a6794b197a967fd4cff5e87e5c2f9e04f2) for my original proof of concept.

We definitely need a more robust solution.

Should we await the renderResult & then run [the conditional logic](https://github.com/shakacode/react_on_rails/blob/judahmeek/ssr-promise/node_package/src/serverRenderReactComponent.ts#L79-L89) on the resolved value just in case the client is using both react-router & a graphql library (in which case we might have a promise resolve a redirect or routeError)?
</blockquote></details>

Can the server do anything with routeError or redirects?  We'll need to pair on these changes.

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 5 unresolved discussions (waiting on @justin808)

___
*[node_package/babel.config.js, line 2 at r6](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-MiBmyIO0HEu7R2C8Nzi:-MiCeXyX2qUCz7xSFLQ9:b-w5hop) ([raw file](https://github.com/shakacode/react_on_rails/blob/97d749d17007a30dcce19d674baf9a4a58221d3a/node_package/babel.config.js#L2)):*

@judahmeek, what motivated this one? 
</blockquote></details>

await/promise support for jest testing
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-MiCf_AR7D8vbc9CmePV:b51r1ug) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

Can the server do anything with routeError or redirects?  We'll need to pair on these changes.
</blockquote></details>

The NodeRenderer or the Rails server?

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 2 unresolved discussions (waiting on @Judahmeek)

___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-MiFM0tS-3HOnFzIlMg6:b-p7r7aw) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*

> This might become simpler if we always return a promise, right?

No. If using ExecJS, the result from `serverRenderReactComponent` turns into a string & promises don't serialize.
</blockquote></details>

Looks like you did this, even though you wrote "No".
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-MiFLzPaA_afeitg1J8l:bb7dssm) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

The NodeRenderer or the Rails server?
</blockquote></details>

both

Judahmeek (commented):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 2 unresolved discussions (waiting on @justin808)

___
*[node_package/src/serverRenderReactComponent.ts, line 59 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr7pQZ960lnLxnKTY_:-Mk0Wz0z4nLkIUri3Ad9:b107xap) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L59)):*

Looks like you did this, even though you wrote "No".
</blockquote></details>

I'm not sure why you say that.
___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-Mk0XJHR4SAjohPobq9Y:bd9aaxo) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

both
</blockquote></details>

Theoretically, the rails server could conditionally make another request to the NodeRenderer if the NodeRenderer returns a Redirect notice, in which case the logic would be something like:

1. Rails makes request to NodeRenderer
2. NodeRenderer returns the redirect object
3. Rails checks the NodeRenderResult to see if it is a redirect
4. If a redirect, Rails sends a new render request with the corrected location & props

My naive assumption is that there could only possibly be a single redirect per HTML request, but we could always throw a RenderError if there happens to be a second redirect for whatever reason.

justin808 (commented):
All files are reviewed. LGTM.

> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-:-Mk1RAdI2xZmSlMLfaT_:b-zc1dvs)* status: all files reviewed, 2 unresolved discussions (waiting on @justin808)

justin808 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1380)* status: all files reviewed, 1 unresolved discussion (waiting on @Judahmeek)

___
*[node_package/src/serverRenderReactComponent.ts, line 101 at r4](https://reviewable.io/reviews/shakacode/react_on_rails/1380#-Mhr54MgAh5RAJcvcgCn:-Mk61OKY6LuLwTNdtX3f:b-cphzk9) ([raw file](https://github.com/shakacode/react_on_rails/blob/027d135e46374ddc7d834aed1b23a33ec79e1087/node_package/src/serverRenderReactComponent.ts#L101)):*

Theoretically, the rails server could conditionally make another request to the NodeRenderer if the NodeRenderer returns a Redirect notice, in which case the logic would be something like:

1. Rails makes request to NodeRenderer
2. NodeRenderer returns the redirect object
3. Rails checks the NodeRenderResult to see if it is a redirect
4. If a redirect, Rails sends a new render request with the corrected location & props

My naive assumption is that there could only possibly be a single redirect per HTML request, but we could always throw a RenderError if there happens to be a second redirect for whatever reason.
</blockquote></details>

I agree with your assessment above, although the renderer does not have to be the node renderer.

I really have no idea about this use case. I suspect that the client can handle the redirect and return the correct path. While maybe not optimal, I don't see a reason to worry about this right now.

PR #1421, opened 2022-01-14
Resolve double mounting error with Turbo(links)

### LINKED ISSUE #1139: Component is rendered twice when using Turbolinks
I've noticed while using `turbolinks` with `react_on_rails` that each component is rendered twice on subsequent visits. It seems that by using the `turbolinks:render` event, the listeners are called twice; once for the cache and once for the new state (see [turbolinks#full-list-of-events](https://github.com/turbolinks/turbolinks#full-list-of-events)). 

Is this expected behaviour? Otherwise the `turbolinks:load` event would be a better candidate to prevent React from wasting renders.

Judahmeek:
Fixes #1139

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1421)

### DIFF
--- a/docs/api/javascript-api.md
+++ b/docs/api/javascript-api.md
@@ -61,7 +61,8 @@ The best source of docs is the main [ReactOnRails.js](https://github.com/shakaco
   /**
    * Set options for ReactOnRails, typically before you call ReactOnRails.register
    * Available Options:
-   * `traceTurbolinks: true|false Gives you debugging messages on Turbolinks events
+   * `traceTurbolinks: true|false` Gives you debugging messages on Turbolinks events
+   * `turbo: true|false` enables Turbo, which is not auto-detected like the Turbolinks library is
    */
   setOptions(options)
 
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -241,16 +241,16 @@ function renderInit(): void {
   if (turboInstalled()) {
     debugTurbolinks(
       'USING TURBO: document added event listeners ' +
-      'turbo:before-render and turbo:render.');
-    document.addEventListener('turbo:before-render', reactOnRailsPageUnloaded);
-    document.addEventListener('turbo:render', reactOnRailsPageLoaded);
+      'turbo:before-visit and turbo:load.');
+    document.addEventListener('turbo:before-visit', reactOnRailsPageUnloaded);
+    document.addEventListener('turbo:load', reactOnRailsPageLoaded);
     reactOnRailsPageLoaded();
   } else if (turbolinksVersion5()) {
     debugTurbolinks(
       'USING TURBOLINKS 5: document added event listeners ' +
-      'turbolinks:before-render and turbolinks:render.');
-    document.addEventListener('turbolinks:before-render', reactOnRailsPageUnloaded);
-    document.addEventListener('turbolinks:render', reactOnRailsPageLoaded);
+      'turbolinks:before-visit and turbolinks:load.');
+    document.addEventListener('turbolinks:before-visit', reactOnRailsPageUnloaded);
+    document.addEventListener('turbolinks:load', reactOnRailsPageLoaded);
     reactOnRailsPageLoaded();
   } else {
     debugTurbolinks(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 Changes since last non-beta release.
 
 ### Fixed
-- Double rendering of components with using Turbo or Turbolinks. [PR 1421](https://github.com/shakacode/react_on_rails/pull/1421) by  [judahmeek](https://github.com/judahmeek).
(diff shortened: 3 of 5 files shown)


docs/api/javascript-api.md
```diff
@@ -61,7 +61,8 @@ The best source of docs is the main [ReactOnRails.js](https://github.com/shakaco
   /**
    * Set options for ReactOnRails, typically before you call ReactOnRails.register
    * Available Options:
-   * `traceTurbolinks: true|false Gives you debugging messages on Turbolinks events
+   * `traceTurbolinks: true|false` Gives you debugging messages on Turbolinks events
+   * `turbo: true|false` enables Turbo, which is not auto-detected like the Turbolinks library is
```
templeman15, line 65: @Judahmeek do we need to update this here as it doesn't seem to include `turbo`?

Screen Shot 2022-01-14 at 3 39 50 PM

justin808: we could allow for either string, but it's not totally required.

templeman15: Yeah I was thinking `turbo?: boolean`

Or `turbo: boolean | null` or however it's actually defined. 

I think it'd help increase visibility because my IDE only shows `traceTurboLinks`. 

I'm speaking specifically about the d.ts file.

justin808: if we could easily just have the shorter traceTurbo, that's better for the long-term.

justin808: @Judahmeek, yes, we need to update the d.ts file.

node_package/src/clientStartup.ts
```diff
@@ -241,16 +241,16 @@ function renderInit(): void {
...
     reactOnRailsPageLoaded();
   } else if (turbolinksVersion5()) {
     debugTurbolinks(
       'USING TURBOLINKS 5: document added event listeners ' +
-      'turbolinks:before-render and turbolinks:render.');
-    document.addEventListener('turbolinks:before-render', reactOnRailsPageUnloaded);
-    document.addEventListener('turbolinks:render', reactOnRailsPageLoaded);
+      'turbolinks:before-visit and turbolinks:load.');
+    document.addEventListener('turbolinks:before-visit', reactOnRailsPageUnloaded);
+    document.addEventListener('turbolinks:load', reactOnRailsPageLoaded);
```
justin808, line 253: @Judahmeek did you test on Turbolinks 5?

docs/api/javascript-api.md
templeman15: @justin808 my interpretation of the code seems slightly different.

For Turbo to work I need:

`turbo: true`

To see the logging I need:

`traceTurboLinks: true`

Both have to be on for it to work and log with Turbo. 

--

Turbolinks seems to work without opting into it. `traceTurboLinks: true` just prints to the logs in that scenario.

node_package/src/clientStartup.ts
Judahmeek: I did.

Judahmeek:
@justin808 

CI was failing due to https://github.com/webpack-contrib/mini-css-extract-plugin/pull/899.

I can put the fix commit in a separate PR if you'd prefer, but I'm expecting a new version from mini-css-extract-plugin's maintainers pretty quick since they already merged the fix.

justin808 (changes requested):
> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1421)* status: all files reviewed, 3 unresolved discussions (waiting on @Judahmeek)

___
*[CHANGELOG.md, line 19 at r2](https://reviewable.io/reviews/shakacode/react_on_rails/1421#-MtfmLY37caq6TJuTIC4:-MtfmLY425-ZCZRa1Svj:b-u6f23j) ([raw file](https://github.com/shakacode/react_on_rails/blob/8cd4b57024925ecab817a5471abd3f008b9535a4/CHANGELOG.md#L19)):*
> ```Markdown
> 
> ### Fixed
> - Double rendering of components with using Turbo or Turbolinks. [PR 1421](https://github.com/shakacode/react_on_rails/pull/1421) by  [judahmeek](https://github.com/judahmeek).
> ```

and TS type for turbo option
___
*[node_package/src/clientStartup.ts, line 253 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1421#-MtPTqO9Kaqucq8CKfeV-r1-253:-MtfmCQnEXubUmcTPdP1:b-8fx2u7) ([raw file](https://github.com/shakacode/react_on_rails/blob/a6cc1e2272426e1e93208207997525e333ebc8f5/node_package/src/clientStartup.ts#L253)):*

I did.
</blockquote></details>

:+1:

justin808 (commented):
@judahmeek up to you. After you make the changelog change, you can merge, squashing commits with a nice commit message.

> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1421#-:-MtfmPyK5EKWs7kFfKxo:b4uaxr6)* status: all files reviewed, 3 unresolved discussions (waiting on @Judahmeek)

Judahmeek:
@justin808, @templeman15 says that this fix created a new problem with flashing content. We're exploring alternative solutions, such as #1425.

Also, I don't have permissions to merge PRs to `master`.

PR #1427, opened 2022-01-26
Update shakapacker for spec dummy

gscarv13:
The goal of this PR is to replace Webpacker with Shakapacker for `spec/dummy`

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1427)


justin808:
See #1428.

Issue #1429, opened 2022-02-09
Fixing failing tests on master


justin808:
Tests are failing on master...

Here's some research on this this:

To repro:

```
bundle exec rake run_rspec:examples
cd gen-examples/examples/basic-server-rendering
rails s -e test
tail -f log/test.log
```

The problem is that the server-bundle.js is getting generated with chunks. The SSR only pulls in the top level file. It doesn't deal with chunks.

```json
{
  "entrypoints": {
    "hello-world-bundle": {
      "assets": {
        "js": [
          "/packs-test/js/runtime.js",
          "/packs-test/js/134.js",
          "/packs-test/js/hello-world-bundle.js"
        ],
        "css": [
          "/packs-test/css/hello-world-bundle.css"
        ]
      }
    },
    "server-bundle": {
      "assets": {
        "js": [
          "/packs-test/js/runtime.js",
          "/packs-test/js/134.js",
          "/packs-test/js/server-bundle.js"
        ],
        "css": [
          "/packs-test/css/server-bundle.css"
        ]
      }
    }
  },
  "hello-world-bundle.css": "/packs-test/css/hello-world-bundle.css",
  "hello-world-bundle.js": "/packs-test/js/hello-world-bundle.js",
  "js/134.js": "/packs-test/js/134.js",
  "js/134.js.LICENSE.txt": "/packs-test/js/134.js.LICENSE.txt",
  "runtime.js": "/packs-test/js/runtime.js",
  "server-bundle.css": "/packs-test/css/server-bundle.css",
  "server-bundle.js": "/packs-test/js/server-bundle.js"
}
```

justin808:
The root cause was the updated bin/webpacker binstub causes a search for config/webpack/test.js which didn't exist.

Then the default webpack config was used which used code-splitting.

We really need a better SSR failure message when this happens!



```
Puma caught this error: wrong number of arguments (given 1, expected 0) (ActionView::Template::Error)
/Users/justin/shakacode/react-on-rails/react_on_rails_2/lib/react_on_rails/prerender_error.rb:12:in `initialize'
/Users/justin/shakacode/react-on-rails/react_on_rails_2/lib/react_on_rails/helper.rb:486:in `exception'
/Users/justin/shakacode/react-on-rails/react_on_rails_2/lib/react_on_rails/helper.rb:486:in `raise'
[...]

PR #1430, opened 2022-02-09
Fixed generator to support Shakapacker

justin808:
* Also fixed incorrect error handling

Addresses #1429

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1430)

### DIFF
--- a/lib/generators/react_on_rails/templates/base/base/config/webpack/test.js.tt
+++ b/lib/generators/react_on_rails/templates/base/base/config/webpack/test.js.tt
@@ -1,0 +1,6 @@
+<%= add_documentation_reference(config[:message], "// https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/config/webpack/development.js") %>
+
+// Use dev
+const config = require('./development');
+
+module.exports = config;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
+### [13.0.0] - 2022-02-08
+#### Breaking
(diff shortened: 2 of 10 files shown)


justin808:
@tomdracz, please take a look.

lib/generators/react_on_rails/templates/base/base/config/webpack/test.js.tt
```diff
@@ -0,0 +1,6 @@
+<%= add_documentation_reference(config[:message], "// https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/config/webpack/development.js") %>
```
tomdracz, line 1: @justin808 Minor one, should the doc reference link to https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/config/webpack/test.js instead or is development.js fine because we're essentially loading it as is? Might need to see if any relevant updates are needed in the tutorial repo

tomdracz (approved):
Looks good, one minor comment

lib/generators/react_on_rails/templates/base/base/config/webpack/test.js.tt
justin808: @tomdracz I'm wondering if we can refactor the sort of convoluted current setup a bit. Think about what would be the simplest JS code to get the job done, maybe depending on a small NPM helper package, like react-on-webpacker.

justin808: @tomdracz we actually need some tests in that sample repo, so then we'd have a file called test.js.

Having a file named `webpack/<ENV>.js` is the old webpacker style, ditched in favor of a simpler webpack.config.js.

justin808: I updated! for the 13.0.1 release!

PR #1432, opened 2022-02-09
POC for creating a registered components hash for Turbolinks cached components

templeman15:
POC for creating a registered components hash for Turbolinks cached components to prevent re-mounting of cached components.

Ideally there would be a unique key that exists on react_on_rails components that does not change between page visits. With that we could just simply mark that component as cached instead of the additional complexity I've added for window.location.

### DIFF
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -135,6 +135,10 @@ function domNodeIdForEl(el: Element): string {
  * delegates to a renderer registered by the user.
  * @param el
  */
+
+type ComponentCache = { [key: string]: boolean };
+let components: { [key: string]: ComponentCache; } = {};
+
 function render(el: Element, railsContext: RailsContext): void {
   const context = findContext();
   // This must match lib/react_on_rails/helper.rb
@@ -142,6 +146,20 @@ function render(el: Element, railsContext: RailsContext): void {
   const domNodeId = domNodeIdForEl(el);
   const props = (el.textContent !== null) ? JSON.parse(el.textContent) : {};
   const trace = el.getAttribute('data-trace') === "true";
+  const path = window.location.pathname;
+  const locationMounted = components[path];
+
+  if (locationMounted) {
+    const componentMounted = locationMounted[props.componentName];
+    if (componentMounted) return;
+
+    components[path] = {
+      ...components[path],
+      [props.componentName]: true,
+    }
+  } else {
+    components[path] = { [props.componentName]: true };
+  }
 
   try {
     const domNode = document.getElementById(domNodeId);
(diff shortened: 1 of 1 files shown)


node_package/src/clientStartup.ts
```diff
@@ -135,13 +135,31 @@ function domNodeIdForEl(el: Element): string {
...
+let components: { [key: string]: ComponentCache; } = {};
+
 function render(el: Element, railsContext: RailsContext): void {
   const context = findContext();
   // This must match lib/react_on_rails/helper.rb
   const name = el.getAttribute('data-component-name') || "";
   const domNodeId = domNodeIdForEl(el);
   const props = (el.textContent !== null) ? JSON.parse(el.textContent) : {};
   const trace = el.getAttribute('data-trace') === "true";
+  const path = window.location.pathname;
```
templeman15, line 149: With a unique key all of the code below goes away and could become:

```
const uniqueKeyThatDoesntChangeBetweenVisits = keyHere;
const componentMounted = components[uniqueKeyThatDoesntChangeBetweenVisits];
if (componentMounted) return;

components = {
  ...components,
  [uniqueKeyThatDoesntChangeBetweenVisits]: true,
}
```

node_package/src/clientStartup.ts
```diff
@@ -135,13 +135,31 @@ function domNodeIdForEl(el: Element): string {
...
+  const locationMounted = components[path];
+
+  if (locationMounted) {
+    const componentMounted = locationMounted[props.componentName];
+    if (componentMounted) return;
```
templeman15, line 154: returning seems to work great for us. I'm not positive if we do in fact need to continue execution of some of the code in the remainder of this function that already existed. But we can't run all of it.

justin808:
@templeman15 Awesome! 🎉 👏 

Can you take a look at this PR: https://github.com/shakacode/react-webpack-rails-tutorial/pull/492

Can you reproduce the issue on https://github.com/shakacode/react-webpack-rails-tutorial (maybe using the branch above for stimulus)?

Could this code have any backward compatibility issues?

@tomdracz @Judahmeek Please review.

templeman15:
> @templeman15 Awesome! 🎉 👏

🥳 

> Can you take a look at this PR: [shakacode/react-webpack-rails-tutorial#492](https://github.com/shakacode/react-webpack-rails-tutorial/pull/492)

Yes I can, what would you like me to review in there, I'm not sure if stimulus is related to the problem we are having?

> Can you reproduce the issue on https://github.com/shakacode/react-webpack-rails-tutorial (maybe using the branch above for stimulus)?

Reproduce the turbolinks / react problem we are having? It's definitely only visibly present on a large monolith that is a little slow when rendering it's server rendered views. If it's a lightweight app that is just communicating with a JSON API, you won't see the problem, the page renders too fast. I've never used stimulus so I'm not sure what I'd be looking out for. 

Although, with how react_on_rails was initially set up it would trigger the components to mount an unmount twice on every cached visit. Triggering two API requests back to back (if an API request occurred on mount).

> Could this code have any backward compatibility issues?

I think it definitely could. If someone was doing this, as we were:

```
document.addEventListener("turbolinks:before-visit", () => {
  if (window.Turbolinks) {
    window.Turbolinks.clearCache();
  }
});
```

I'm not positive what the result would be. The snippet above does prevent the double mounting but the back button no longer works. The solution in this PR allow us to remove the above snippet and get back the back button. 

--

This PR as is should definitely not go in as is. If you know of a unique key that we could use that for each react on rails component that does not change between Turbolinks visits, we can use that and I think it'd work for other users of react on rails. I used `props.componentName` as a cache key which is a special prop we use in our app, it would not exist for any other users of react_on_rails.

--

There is one small issue but it's one we are willing to live with. When you visit an uncached page or server rendered page it looks perfect. 

When you visit a react client rendered page that is cached you see the data that was already loaded and then the fresh version follows which triggers the API request and shows loading state and then the results come back. You definitely need to reload data when visiting a cached page, but it does show content for a split second -> loading -> content.

node_package/src/clientStartup.ts
```diff
@@ -135,13 +135,31 @@ function domNodeIdForEl(el: Element): string {
...
+
+    components[path] = {
+      ...components[path],
+      [props.componentName]: true,
```
tomdracz, line 158: @templeman15 Does this mean you can render each component on the given path only once?
Can't imagine you would be rendering same thing multiple times often, but good to watch out for.

templeman15: @tomdracz that's a good point. We don't do that but I definitely see what you're saying. You could technically call react on rails with the same component multiple times on the page..

I was hoping you all would know a unique key that persists across page visits for the rendered components. In our app I see the react dom ID but it changes even on the same component when visiting a cached page.

templeman15:
We have done two things to fix the double rendering problem  and broken back button we encountered:
Use turbolinks to do all page navigation.

1) Use a tags appropriately and applicable
     a) Navigate with Turbolinks.visit instead of window.location.href =
2) Added redux and rtk query which stops cached requests from firing (this was our biggest concern) as the cached page would switch to loading state twice and do two API requests

PR #1433, opened 2022-02-17
Stop checking encoding names

casperisfine:
Comparing the names is much less efficient than comparing the instance directly.

It may also change in the future: https://bugs.ruby-lang.org/issues/18576

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1433)

### DIFF
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -279,8 +279,9 @@ module ReactOnRails
           # URLs as UTF-8. This situation can occur in browsers that do not encode the
           # entire URL as UTF-8 already, mostly on the Windows platform (IE11 and lower).
           original_url_normalized = request.original_url
-          if original_url_normalized.encoding.to_s == "ASCII-8BIT"
-            original_url_normalized = original_url_normalized.force_encoding("ISO-8859-1").encode("UTF-8")
+          if original_url_normalized.encoding == Encoding::BINARY
+            original_url_normalized = original_url_normalized.force_encoding(Encoding::ISO_8859_1)
+                                                             .encode(Encoding::UTF_8)
           end
 
           # Using Addressable instead of standard URI to better deal with


justin808:
Hi @casperisfine, the fix looks fine. 

Could you add a small test to confirm this fix and a changelog entry?

How did you discover this? 

Will older versions of Ruby or Rails have issues with the fix? Do we need and backwards compatibility code?

casperisfine:
> Could you add a small test to confirm this fix and a changelog entry?

I assume it's already covered, and it's really a small implementation detail.

> How did you discover this?

When we discuss about making change to ruby, we tend to grep across public gems to see what may break.

> Will older versions of Ruby or Rails have issues with the fix?

Not at all, you can ship as it and it will work fine all the way back to 1.9.3.

lib/react_on_rails/helper.rb
```diff
@@ -279,7 +279,7 @@ def rails_context(server_side: true)
           # URLs as UTF-8. This situation can occur in browsers that do not encode the
           # entire URL as UTF-8 already, mostly on the Windows platform (IE11 and lower).
           original_url_normalized = request.original_url
-          if original_url_normalized.encoding.to_s == "ASCII-8BIT"
+          if original_url_normalized.encoding == Encoding::BINARY
             original_url_normalized = original_url_normalized.force_encoding("ISO-8859-1").encode("UTF-8")
```
justin808: @casperisfine should this next line change to use some constants?

casperisfine: It definitely could. Note that I opened over 30 PRs like this one, so I really only focused on what would break if we made this change. `force_encoding("ISO-8859-1")` would still work, so I didn't modify it.

But if it was me, yes I'd change it for `force_encoding(Encoding::ISO_8859_1)`.

justin808: If you can make that change and CI passes, I'll merge!

justin808: Thanks for the fix!

casperisfine: Done.

lib/react_on_rails/helper.rb
```diff
@@ -279,8 +279,9 @@ def rails_context(server_side: true)
           # URLs as UTF-8. This situation can occur in browsers that do not encode the
           # entire URL as UTF-8 already, mostly on the Windows platform (IE11 and lower).
           original_url_normalized = request.original_url
-          if original_url_normalized.encoding.to_s == "ASCII-8BIT"
-            original_url_normalized = original_url_normalized.force_encoding("ISO-8859-1").encode("UTF-8")
+          if original_url_normalized.encoding == Encoding::BINARY
+            original_url_normalized.force_encoding(Encoding::ISO_8859_1)
+            original_url_normalized = original_url_normalized.encode(Encoding::UTF_8)
```
justin808: @casperisfine any reason to put this on 2 separate lines?

casperisfine: To reduce line length.

justin808: Please keep as one line. The line is not long.

casperisfine: Done.

justin808: Thanks.

casperisfine:
```
lib/react_on_rails/helper.rb:283:121: C: Layout/LineLength: Line is too long. [122/120]
            original_url_normalized = original_url_normalized.force_encoding(Encoding::ISO_8859_1).encode(Encoding::UTF_8)
```

justin808: My bad!

```
Inspecting 11[5](https://github.com/shakacode/react_on_rails/runs/5315703112?check_suite_focus=true#step:10:5) files
......................C............................................................................................

Offenses:

lib/react_on_rails/helper.rb:2[8](https://github.com/shakacode/react_on_rails/runs/5315703112?check_suite_focus=true#step:10:8)3:121: C: Layout/LineLength: Line is too long. [122/120]
            original_url_normalized = original_url_normalized.force_encoding(Encoding::ISO_885[9](https://github.com/shakacode/react_on_rails/runs/5315703112?check_suite_focus=true#step:10:9)_1).encode(Encoding::UTF_8)
```

How about just breaking into 2 lines with the dot on the end of the line?

casperisfine: Done.

justin808:
@casperisfine 
Can you run `rubocop -a`.


```
[build (2.7, 14)](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#logs)
failed [1](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:1) hour ago in 1m [4](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:4)s
Search logs
3s
1s
1s
0s
0s
3[5](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:5)s
5s
11s
1s
[7](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:7)s
Run bundle exec rubocop
Inspecting 115 files
......................C............................................................................................

Offenses:

lib/react_on_rails/helper.rb:2[8](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:8)3:[9](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:9)9: C: [Correctable] Layout/DotPosition: Place the . on the next line, together with the method name.
            original_url_normalized = original_url_normalized.force_encoding(Encoding::ISO_8859_1).
                                                                                                  ^
lib/react_on_rails/helper.rb:284:15: C: [Correctable] Layout/MultilineMethodCallIndentation: Align encode with original_url_normalized.force_encoding(Encoding::ISO_8859_1). on line 283.
              encode(Encoding::UTF_8)
              ^^^^^^

[11](https://github.com/shakacode/react_on_rails/runs/5316818390?check_suite_focus=true#step:10:11)5 files inspected, 2 offenses detected, 2 offenses auto-correctable
Error: Process completed with exit code 1.
```

casperisfine:
Done.

justin808:
Thanks, @casperisfine!

PR #1437, opened 2022-03-31
Not pusing mini_racer in default install. Closes #1438

### LINKED ISSUE #1438: Suggestion. Not pushing `mini_racer` in new installs but use the standard Node.js that is faster

In the past I have had several issues installing `mini_racer` because it depends on the gem `libv8-node` that depends on `libv8` that needs to be installed in the Operating System, and the version of all of them must correspond.

In the present I have issues installing `mini_racer` in adJ/OpenBSD 7.1a1 (because recent libv8 has conflicts with the `clang` compiler in that platform), and  as well as in Ubuntu 21.10, where I couldn't compile. 

However in the later case I could install version 0.6.2.

That `react_on_rails` don push `mini_racer` in new installs, but just using the standard Node.js used by default by rails.

The benchmarks in the `mini_racer` github page are outdated, so I  I wondered if  it was worthy the effort to install and use `mini_racer` in a Ruby on Rails project instead of the default Node.js, so I run the benchmarks included in `mini_racer`

**Additional context**

# Platform

* Processor: Intel(R) Core(TM) i5-8265U 8 cores.
* RAM: 8GB
* Operating System: Ubuntu 21.10
* Node.js: 16.14.0 installed with nvm
* Ruby: 3.0.0 installed with rvm with gemsets in user space

# Preparation of the `mini_racer` benchmark

I cloned the `master` branch of the `mini_racer` repository and run
```
bundle
rake
```

It produced the error:
```
cp tmp/x86_64-linux/mini_racer_extension/3.0.0/mini_racer_extension.so tmp/x86_64-linux/stage/lib/mini_racer_extension.so
/home/vtamara/comp/ruby/mini_racer/lib/mini_racer.rb:6:in `<top (required)>': undefined method `require_paths' for nil:NilClass (NoMethodError)
```

However I was able to install the gem (I guess precompiled) with:
```
% gem install mini_racer
Fetching mini_racer-0.6.2.gem
Building native extensions. This could take a while...
Successfully installed mini_racer-0.6.2
```

To run the babel benchmark I did
```
cd mini_racer/bechmark/babel
```
And in the `Gemfile` I changed the line:
```
gem 'mini_racer', path: '../../'
```
by removing `, path: '../../'` to force the benchmark to use the installed gem.

I run benchmark with `mini_racer`:
```
% bundle exec ruby bench.rb 
```

And then with `Node.js` 
```
% node bench_node.js 
```

To run the `uglify` benchmark I did analogous steps in the directory `benchmark/exec_js_uglify`

# Results of babel benchmark

```
% bundle exec ruby bench.rb 
107.638547 load babel
mini racer 245.63389999999998 transform
mini racer 136.00871 transform
mini racer 107.751951 transform
mini racer 91.45205200000001 transform
mini racer 92.916148 transform
mini racer 78.54188099999999 transform
mini racer 84.429853 transform
mini racer 76.42903199999999 transform
mini racer 89.70468500000001 transform
mini racer 72.84783999999999 transform
mini racer 10x 716.3434669999999 transform
mini racer 10x 672.19839 transform
mini racer 10x 656.803133 transform
mini racer 10x 644.338204 transform
mini racer 10x 669.813773 transform
mini racer 10x 632.5552 transform
mini racer 10x 637.2754170000001 transform
mini racer 10x 616.257543 transform
mini racer 10x 621.600582 transform
mini racer 10x 626.336146 transform
```

----

```
% node bench_node.js 
Running V8 version 9.4.146.24-node.20
91 requiring babel
190 transpile
114 transpile
93 transpile
79 transpile
77 transpile
70 transpile
73 transpile
67 transpile
69 transpile
67 transpile
658 transpile 10 times
642 transpile 10 times
621 transpile 10 times
619 transpile 10 times
630 transpile 10 times
594 transpile 10 times
611 transpile 10 times
574 transpile 10 times
590 transpile 10 times
576 transpile 10 times
```

Cleaning and comparing this results, `mini_racer` in all the cases is slower than Node.js.  In its fastest case it is 2.6% slower and its slowest case it is 19.5% slower with an average of 10% slower.

# Results of uglify benchmark

```
exec_js_uglify master % bundle exec ruby bench.rb node
Benching with node
node minify discourse_app.js 11000.563412ms
node minify discourse_app_minified.js 9899.891416999999ms
node minify discourse_app.js twice (2 threads) 13247.968232ms
```

---

```
exec_js_uglify master % bundle exec ruby bench.rb mini_racer
Benching with mini_racer
mini_racer minify discourse_app.js 13102.9107ms
mini_racer minify discourse_app_minified.js 12039.30939ms
mini_racer minify discourse_app.js twice (2 threads) 16506.145027ms
```

Here `mini_racer` was between 19,1% slower and 24,5% slower with an average of 21.7%.

# Conclusion

For short javascript tasks `mini_racer` is 10% slower than Node.js, for longer tasks `mini_racer` is 21.7% slower than Node.js.  

Since `mini_racer` is so difficult to install and slower than the default Node.js of Rails projects, what for to recommend or push it in new installs of `react_on_rails` ?

Blessings

vtamara:
node is used by default and usually it has less problems than mini_racer.

react_on_rails shouldn't push `mini_racer` in new installations. This patch removes such behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1437)

### DIFF
--- a/docs/guides/client-vs-server-rendering.md
+++ b/docs/guides/client-vs-server-rendering.md
@@ -4,7 +4,7 @@
 
 In most cases, you should use the `prerender: false` (default behavior) with the provided helper method to render the React component from your Rails views. In some cases, such as when SEO is vital, or many users will not have JavaScript enabled, you can enable server-rendering by passing `prerender: true` to your helper, or you can simply change the default in `config/initializers/react_on_rails`.
 
-Now the server will interpret your JavaScript. The default is to use [ExecJS](https://github.com/rails/execjs) and pass the resulting HTML to the client. We recommend using [mini_racer](https://github.com/discourse/mini_racer) as ExecJS's runtime.
+Now the server will interpret your JavaScript. The default is to use Node.js.
 
 If you want to maximize the perfomance of your server rendering, then you want to use React on Rails Pro which uses NodeJS to do the server rendering. See the [docs for React on Rails Pro](https://github.com/shakacode/react_on_rails/wiki).
 
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -137,14 +137,14 @@ ReactOnRails.configure do |config|
   ################################################################################
   # Server Renderer Configuration for ExecJS
   ################################################################################
-  # The default server rendering is ExecJS, probably using the mini_racer gem
+  # The default server rendering is Node.js
   # If you wish to use an alternative Node server rendering for higher performance,
   # contact justin@shakacode.com for details.
   #
   # For ExecJS:
   # You can configure your pool of JS virtual machines and specify where it should load code:
-  # On MRI, use `mini_racer` for the best performance
-  # (see [discussion](https://github.com/reactjs/react-rails/pull/290))
+  # On MRI, use `Node.js` for good performance
+  # (see [issue](https://github.com/shakacode/react_on_rails/issues/1438))
   # On MRI, you'll get a deadlock with `pool_size` > 1
   # If you're using JRuby, you can increase `pool_size` to have real multi-threaded rendering.
   config.server_renderer_pool_size = 1 # increase if you're on JRuby
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -73,7 +73,6 @@ module ReactOnRails
       end
 
       def add_base_gems_to_gemfile
-        gem "mini_racer", platforms: :ruby
         run "bundle"
       end
 
--- a/Gemfile.development_dependencies
+++ b/Gemfile.development_dependencies
@@ -21,8 +21,6 @@ gem "sprockets", "~> 4.0"
 
 gem "amazing_print"
 
-gem "mini_racer", ">= 0.6.2"
-
 group :development, :test do
   gem "listen"
   gem "pry"
--- a/docs/guides/tutorial.md
+++ b/docs/guides/tutorial.md
@@ -91,8 +91,6 @@ Install React on Rails: `rails generate react_on_rails:install`. You need to fir
 Note, using `redux` is no longer recommended as the basic installer uses React Hooks.
 If you want the redux install: `rails generate react_on_rails:install --redux`
 
-The generator will add `mini_racer`'s latest version. If you're using linux & encounter issues installing `libv8`, here's [a common solution](https://github.com/rubyjs/mini_racer/issues/218).
-
 ```
 bundle exec rails generate react_on_rails:install
 ```
--- a/docs/javascript/server-rendering-tips.md
+++ b/docs/javascript/server-rendering-tips.md
@@ -2,9 +2,6 @@
 
 For the best performance with Server Rendering, consider using [React on Rails Pro]
 
-Be sure to use mini_racer. See [issues/428](https://github.com/shakacode/react_on_rails/issues/428).
-
-
 
 ## General Tips
 - Your code can't reference `document`. Server side JS execution does not have access to `document`,


justin808:
@alexeyr @tomdracz Please take a look at this. 

1. Is this a breaking change in any way? 
2. Recommended under what circumstances?
3. Any downsides?

I don't quite understand why recently this change is needed.

tomdracz:
From my perspective:
- agree that `mini_racer` creates some headaches. Had a lot of time spent debugging and swearing when CI wasn't playing ball with `libv8` versions
- Breaking change? Possibly, maybe. You might hit some implementation differences but considering Node itself is built on top of v8, I wouldn't expect much difference apart from some edge cases
- Having `mini_racer` as dependency means sneakily changing ExecJS runtime from Node to `mini_racer` if it wasn't previously specified. Following principle of least surprise, I would rather have the app dictate what they use for the JS engine. If people want to use `mini_racer` for whatever circumstances, we should let them, but they should do it in the main app Gemfile.
- Can't see downsides here really. Considering you will need Node either way and it's faster, I see no reason to rely on `mini_racer`
- `mini_racer` is only useful if you want to use your app in an environment without existing JS engine. But you will need Node to compile assets so this point is moot for anyone using `react_on_rails`

I've ran a few tests on my app that is using `mini_racer` gem - yanking it out resulted in no visible changes, app still works, test suite still passes etc.

Happy for this one to go, will add some minor comments.

tomdracz:
From my perspective:
- agree that `mini_racer` creates some headaches. Had a lot of time spent debugging and swearing when CI wasn't playing ball with `libv8` versions
- Breaking change? Possibly, maybe. You might hit some implementation differences but considering Node itself is built on top of v8, I wouldn't expect much difference apart from some edge cases
- Having `mini_racer` as dependency means sneakily changing ExecJS runtime from Node to `mini_racer` if it wasn't previously specified. Following principle of least surprise, I would rather have the app dictate what they use for the JS engine. If people want to use `mini_racer` for whatever circumstances, we should let them, but they should do it in the main app Gemfile.
- Can't see downsides here really. Considering you will need Node either way and it's faster, I see no reason to rely on `mini_racer`
- `mini_racer` is only useful if you want to use your app in an environment without existing JS engine. But you will need Node to compile assets so this point is moot for anyone using `react_on_rails`

I've ran a few tests on my app that is using `mini_racer` gem - yanking it out resulted in no visible changes, app still works, test suite still passes etc.

Happy for this one to go, will add some minor comments.

docs/guides/client-vs-server-rendering.md
```diff
@@ -4,7 +4,7 @@
 
 In most cases, you should use the `prerender: false` (default behavior) with the provided helper method to render the React component from your Rails views. In some cases, such as when SEO is vital, or many users will not have JavaScript enabled, you can enable server-rendering by passing `prerender: true` to your helper, or you can simply change the default in `config/initializers/react_on_rails`.
 
-Now the server will interpret your JavaScript. The default is to use [ExecJS](https://github.com/rails/execjs) and pass the resulting HTML to the client. We recommend using [mini_racer](https://github.com/discourse/mini_racer) as ExecJS's runtime.
+Now the server will interpret your JavaScript. The default is to use Node.js.
```
tomdracz, line 7: Think ExecJS will still be used behind the scenes but with Node.js rather than `mini_racer`. Maybe change this to reflect this and tell people they've got choice of alternative runtimes, as documented in ExecJS readme?

docs/guides/configuration.md
```diff
@@ -137,14 +137,14 @@ ReactOnRails.configure do |config|
   ################################################################################
   # Server Renderer Configuration for ExecJS
   ################################################################################
-  # The default server rendering is ExecJS, probably using the mini_racer gem
+  # The default server rendering is Node.js
```
tomdracz, line 140: As above, still will be ExecJS but with Node runtime

tomdracz, line 140: As above, still will be ExecJS but with Node runtime

docs/guides/configuration.md
```diff
@@ -137,14 +137,14 @@ ReactOnRails.configure do |config|
...
   # If you wish to use an alternative Node server rendering for higher performance,
   # contact justin@shakacode.com for details.
   #
   # For ExecJS:
   # You can configure your pool of JS virtual machines and specify where it should load code:
-  # On MRI, use `mini_racer` for the best performance
-  # (see [discussion](https://github.com/reactjs/react-rails/pull/290))
```
tomdracz, line 147: The discussion there has shown that mini_racer performed better than Node. That was quite a few years back, would be good to link to any benchmarks supporting the fact that Node is more peformant

tomdracz, line 147: The discussion there has shown that mini_racer performed better than Node. That was quite a few years back, would be good to link to any benchmarks supporting the fact that Node is more peformant

lib/generators/react_on_rails/base_generator.rb
```diff
@@ -73,7 +73,6 @@ def copy_webpacker_config
       end
 
       def add_base_gems_to_gemfile
-        gem "mini_racer", platforms: :ruby
```
tomdracz, line 76: We should have a changelog entry reflecting that. If users already have `mini_racer` in their Gemfiles, nothing will change. We should note that it's been removed and ask users to also remove and test their apps for any changes

tomdracz, line 76: We should have a changelog entry reflecting that. If users already have `mini_racer` in their Gemfiles, nothing will change. We should note that it's been removed and ask users to also remove and test their apps for any changes

tomdracz (changes requested):

docs/guides/client-vs-server-rendering.md
alexeyr-c: That would be my suggestion too.

docs/guides/configuration.md
alexeyr-c: Yes, but since it already says "for ExecJS" just 2 lines above I don't know if it needs to be said again.

docs/guides/client-vs-server-rendering.md
alexeyr-c: Actually, can anything other than ExecJS be used? If not, the correct phrasing could be something like "Now the server will interpret your JavaScript using ExecJS. You can pick the engine as described in the documentation." If yes, this is a good place to explain how.

PR #1439, opened 2022-03-31
Not checking presence of i18n directory at startup. Closes #1319

### LINKED ISSUE #1319: Remove check for i18n directory at startup
The i18n directory might be deleted to save on the Heroku slug size.

React on Rails checks that the i18n directory exists and errors out on startup

However, that’s supposed to be an informative message at runtime. It’s only used during precompile.

We can check that value lazily rather than in the config setup.

vtamara:
Checking just when running rake task `react_on_rails:locale`.

Fixes #1319. 

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1439)

### DIFF
--- a/lib/react_on_rails/locales/base.rb
+++ b/lib/react_on_rails/locales/base.rb
@@ -4,7 +4,34 @@ require "erb"
 
 module ReactOnRails
   module Locales
+
+    def self.check_i18n_directory_exists
+      return if ReactOnRails.configuration.i18n_dir.nil?
+      return if Dir.exist?(ReactOnRails.configuration.i18n_dir)
+
+      msg = <<-MSG.strip_heredoc
+      Error configuring /config/initializers/react_on_rails.rb: invalid value for `config.i18n_dir`.
+      Directory does not exist: #{ReactOnRails.configuration.i18n_dir}. Set to value to nil or comment it
+      out if not using the React on Rails i18n feature.
+      MSG
+      raise ReactOnRails::Error, msg
+    end
+
+    def self.check_i18n_yml_directory_exists
+      return if ReactOnRails.configuration.i18n_yml_dir.nil?
+      return if Dir.exist?(ReactOnRails.configuration.i18n_yml_dir)
+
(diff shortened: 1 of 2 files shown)


lib/react_on_rails/locales/base.rb
```diff
@@ -4,7 +4,34 @@
 
 module ReactOnRails
   module Locales
+
+    def self.check_i18n_directory_exists
+      return if ReactOnRails.configuration.i18n_dir.nil?
```
alexeyr, line 9: I'd prefer to extract `ReactOnRails.configuration.i18n_dir` to a variable or call after https://github.com/shakacode/react_on_rails/blob/21d79dece18547c8b4e709724178a79bcc6ed481/lib/react_on_rails/locales/base.rb#L94 instead of repeating. Same in `self.check_i18n_yml_directory_exists`.

lib/react_on_rails/locales/base.rb
```diff
@@ -4,7 +4,34 @@
...
+      Error configuring /config/initializers/react_on_rails.rb: invalid value for `config.i18n_yml_dir`.
+      Directory does not exist: #{ReactOnRails.configuration.i18n_yml_dir}. Set to value to nil or comment it
+      out if not using this i18n with React on Rails, or if you want to use all translation files.
+      MSG
+      raise ReactOnRails::Error, msg
+    end
+
     def self.compile
+      self.check_i18n_directory_exists
+      self.check_i18n_yml_directory_exists
```
alexeyr, line 34: I would suggest unifying the two `check...` functions: `def self.check_directory_exists(directory, key_name)` and then 
```
self.check_directory_exists(ReactOnRails.configuration.i18n_dir, "config.i18n_dir")
self.check_directory_exists(ReactOnRails.configuration.i18n_yml_dir, "config.i18n_yml_dir")
```

justin808: @alexeyr can you please finish up this one? Feel free to open a new PR.

Issue #1441, opened 2022-04-10
React 18 Warning: You are importing hydrateRoot from "react-dom" [...] You should instead import it from "react-dom/client"


gastonmorixe:
```
Warning: You are importing hydrateRoot from "react-dom" which is not supported.
You should instead import it from "react-dom/client".
```

https://github.com/shakacode/react_on_rails/blob/0e96cf8a4c6779daf2c671974e93598134162587/node_package/src/reactHydrate.ts#L1

justin808:
Hi @gastonmorixe, Ari Summer is working on this as well. 

We can chat at the [Slack Discusssion](https://join.slack.com/t/reactrails/shared_invite/enQtNjY3NTczMjczNzYxLTlmYjdiZmY3MTVlMzU2YWE0OWM0MzNiZDI0MzdkZGFiZTFkYTFkOGVjODBmOWEyYWQ3MzA2NGE1YWJjNmVlMGE).

summera:
I took a stab at fixing this in https://github.com/shakacode/react_on_rails/pull/1448

smeijer:
Is there a status update for this?

justin808:
I'll soon be trying to do some upgrades.

justin808:
@gastonmorixe @summera @smeijer Completed! I'll ship an updated version soon!

smeijer:
Thanks! Appreciate the ping.

PR #1448, opened 2022-04-23
Conditionally require react-dom/client in reactHydrate/reactRender if React version >= 18

summera:
### Summary
This resolves https://github.com/shakacode/react_on_rails/issues/1441. 

With React 18, `hydrateRoot` and `createRoot` have been moved to `react-dom/client`. If you try to access these methods on `react-dom`, you'll see a warning that looks like this:

>  Warning: You are importing createRoot from “react-dom” which is not supported. You should instead import it from “react-dom/client”.

React 18 support was added in https://github.com/shakacode/react_on_rails/pull/1409 but also resulted in this warning since `createRoot` and `hydrateRoot` were being called on `react-dom`.

In order to prevent the warning, I added a conditional require using a ternary statement in `reactHydrate.ts` and `reactRender.ts`. I also had to disable the `import/no-unresolved` rule as it was throwing an error when `react-dom/client` wasn't present (in React < 18). I _think_ this should do it but I struggled making eslint and typescript happy so let me know if I'm missing anything.

### DIFF
--- a/node_package/src/reactHydrate.ts
+++ b/node_package/src/reactHydrate.ts
@@ -1,10 +1,11 @@
-import ReactDOM from 'react-dom';
 import { ReactElement, Component } from 'react';
 import supportsReactCreateRoot from './supportsReactCreateRoot';
 
+// eslint-disable-next-line import/no-unresolved
+const ReactDOM = supportsReactCreateRoot ? require("react-dom/client") : require("react-dom");
+
 export default function reactHydrate(domNode: Element, reactElement: ReactElement): void | Element | Component {
   if (supportsReactCreateRoot) {
-    // @ts-expect-error potentially present if React 18 or greater
     return ReactDOM.hydrateRoot(domNode, reactElement);
   }
 
--- a/node_package/src/reactRender.ts
+++ b/node_package/src/reactRender.ts
@@ -1,10 +1,11 @@
-import ReactDOM from 'react-dom';
 import { ReactElement, Component } from 'react';
 import supportsReactCreateRoot from './supportsReactCreateRoot';
 
+// eslint-disable-next-line import/no-unresolved
+const ReactDOM = supportsReactCreateRoot ? require("react-dom/client") : require("react-dom");
+
 export default function reactRender(domNode: Element, reactElement: ReactElement): void | Element | Component {
   if (supportsReactCreateRoot) {
-    // @ts-expect-error potentially present if React 18 or greater
     const root = ReactDOM.createRoot(domNode);
     root.render(reactElement);
     return root


Judahmeek (changes requested):
@summera Thanks for your contribution!

Unfortunately, it's failing CI.

Looks like the `require("react-dom/client")` even though `/spec/dummy/yarn.lock`'s version of r react-dom is v16.

I'm not sure what the solution is.

Maybe

```javascript
const reactDomSrc = supportsReactCreateRoot ? "react-dom/client" : "react-dom";
require(reactDomSrc);
```
???

summera:
@Judahmeek unfortunately doing that results in the following errors:

>   5:18  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
>   5:18  error  Calls to require() should use string literals   import/no-dynamic-require

If the approach taken by @tomdracz in https://github.com/shakacode/react_on_rails/pull/1449 works, I suggest we close this out in favor of that PR.

tomdracz:
> @Judahmeek unfortunately doing that results in the following errors:
> 
> 
> 
> >   5:18  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
> 
> >   5:18  error  Calls to require() should use string literals   import/no-dynamic-require
> 
> 
> 
> If the approach taken by @tomdracz in https://github.com/shakacode/react_on_rails/pull/1449 works, I suggest we close this out in favor of that PR.

Think mine might suffer from similar issues, although the warnings are perhaps ok to ignore.

With the way React made it quite hard to support migrations due to API changes, I think the only clean ways to do it are:
- require react 18 and make new create root API optional. If not using the new API, the app will behave as before. Will still emit annoying warnings though and it's a bit of sledgehammer approach.
- create "adapter" subpackages, one for React < 18 and one for React 18+. They would encapsulate the logic that's added in my PR. This will create additional installation step (adding correct package for your react version), but we could have clear React dependency specified in those rather than React in Rails core

summera:
@Judahmeek would you mind kicking off the workflows again? I _may_ have gotten this passing with my latest commit, but I'd like to verify.

justin808:
I just updated with master. This might pass CI now.

Which should we favor? This or #1449? @tomdracz @summera

tomdracz:
I think this one has better chance of working than mine. Using require rather than dynamic import is more likely to work without issues.

Looks like we are still getting some spec fails so need to investigate that

node_package/src/reactHydrate.ts
```diff
@@ -1,10 +1,13 @@
-import ReactDOM from 'react-dom';
 import { ReactElement, Component } from 'react';
 import supportsReactCreateRoot from './supportsReactCreateRoot';
 
+const reactDomClient = "react-dom/client";
+const reactDom = "react-dom";
+// eslint-disable-next-line import/no-dynamic-require
+const ReactDOM = require(supportsReactCreateRoot ? reactDomClient : reactDom);
```
tomdracz: Try 

```
const ReactDOM = require(`react-dom${supportsReactCreateRoot ? '/client' : ''}`)
```

Might solve resolution error

@summera @justin808

justin808:
@summera do you want to finish this one up? I'd love to get it merged!

node_package/src/reactHydrate.ts
summera: @tomdracz I can certainly give this a try if you think it'll resolve the load error. I've tried a bunch of combinations without much luck. Do you mind explaining why this would solve the issue?

tomdracz: I can't quite find the article where I've read about it and it's just a hunch but I think that gives webpack bit more information how to handle import.

If you pass just the variable, I think it goes bit wonky and doesn't quite know what to do - potentially including any possible module. Whereas if passed template string:
```
`react-dom${supportsReactCreateRoot ? '/client' : ''}`
```

I think webpack will create require.context for anything matching react-dom. I'm thinking that since the react-dom is required and we can assume this folder will be there in node_modules, it will be bit more keen to grab the correct thing. Only one way to find out!

More info https://webpack.js.org/guides/dependency-management/#require-with-expression

summera: @tomdracz ah ok, got it. Thanks for the explanation! Hope it works. I just pushed up the change in [e68ad9b](https://github.com/shakacode/react_on_rails/pull/1448/commits/e68ad9b86c944dd4fcf1edf750f2db82544b14e2).

tomdracz:
I'm at loss with this one. Found reference to template strings in require https://stackoverflow.com/questions/42797313/webpack-dynamic-module-loader-by-require/42804941#42804941

I can see in the error we're getting empty context error. This leads me to believe that either:
- package is not there
- we're looking it up in a wrong directory (more likely)

Need to dig into this and see what is the context returned by that dynamic require statement. Might bring us closer to figuring out what's wrong (maybe some webpack setting?). It's as if it's not looking up module inside the node_modules folder (which I THINK  it should by default)

alexeyr-ci1:
#1460 has been merged instead.

PR #1449, opened 2022-04-25
React 18 conditional support

tomdracz:
Another stab at #1441

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1449)


node_package/src/clientStartup.ts
```diff
@@ -213,6 +211,7 @@ function unmount(el: Element): void {
   const domNode = document.getElementById(domNodeId);
   if(domNode === null){return;}
   try {
+    // Might need updating? https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html
```
tomdracz, line 227: `unmountComponentAtNode` is considered legacy https://reactjs.org/docs/react-dom.html#unmountcomponentatnode but would need to store references to root instead if we were to update

node_package/src/helpers/renderHelper.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as modernRootHandler from './modernRootHandler';
+
+
+let toExport;
+
+if (supportsReactCreateRoot) {
+  toExport = legacyRootHandler;
+} else {
+  toExport = modernRootHandler;
+}
```
tomdracz: Struggling to find a nicer way to get this going, can't think of a way to do a dynamic import and then reexport

node_package/src/helpers/renderHelper.ts
```diff
@@ -0,0 +1,19 @@
+import supportsReactCreateRoot from './supportsReactCreateRoot';
+import * as legacyRootHandler from './legacyRootHandler';
+import * as modernRootHandler from './modernRootHandler';
+
+
+let toExport;
+
+if (supportsReactCreateRoot) {
```
tomdracz: This could be good based not only on React version but also some configuration option so it's an opt-in

tomdracz:
Also worth noting https://github.com/reactwg/react-18/discussions/125#discussioncomment-2383200 and multiple roots (which is what we're doing) being discouraged, although not a great info on "why"

node_package/src/helpers/renderHelper.ts
tomdracz: Would be fairly trivial with top-level await though 🤔

summera:
> Also worth noting https://github.com/reactwg/react-18/discussions/125#discussioncomment-2383200 and multiple roots (which is what we're doing) being discouraged, although not a great info on "why"

I also noticed this while working on https://github.com/shakacode/react_on_rails/pull/1448. @justin808 would switching to [portals](https://reactjs.org/docs/portals.html) be an option for react_on_rails?

tomdracz:
> > Also worth noting https://github.com/reactwg/react-18/discussions/125#discussioncomment-2383200 and multiple roots (which is what we're doing) being discouraged, although not a great info on "why"
> 
> 
> 
> I also noticed this while working on https://github.com/shakacode/react_on_rails/pull/1448. @justin808 would switching to [portals](https://reactjs.org/docs/portals.html) be an option for react_on_rails?

Don't think portals are suitable here, would normally use them for some modals and overlays rather than full component trees. It doesn't seem like the docs discourage it anywhere, so apart from that one comment, there's no clear rationale and options from React team. Probably good to leave as is for now

tomdracz:
@Judahmeek @justin808 any of you able to trigger the tests to see the output?

justin808 (approved):
Looks fantastic with a quick code review!

@tomdracz I like your coding style a lot!

> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1449#-:-N0itj9y1bBGUnwVTeXG:b-11o7nf)* status: all files reviewed, 3 unresolved discussions (waiting on @tomdracz)

___
*[`node_package/src/clientStartup.ts` line 214 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1449#-N0VOHHx6EDAAKsFFT-b-r1-214:-N0isiZBE8Vnr3RVEovu:b-h2fey8) ([raw file](https://github.com/shakacode/react_on_rails/blob/af52b8aa838ca1d1ecefcd684a28e051bab5b92f/node_package/src/clientStartup.ts#L214)):*
<details><summary><i>Previously, tomdracz (Tom Dracz) wrote…</i></summary><blockquote>

`unmountComponentAtNode` is considered legacy https://reactjs.org/docs/react-dom.html#unmountcomponentatnode but would need to store references to root instead if we were to update
</blockquote></details>

Agree. It looks like there is more to do: https://github.com/reactwg/react-18/discussions/5.

Consequently, I'm not sure we should use the new `createRoot` API until we're confident that ReactOnRails is ready, as that is why the old behavior is available. 
___
*[`node_package/src/helpers/renderHelper.ts` line 12 at r1](https://reviewable.io/reviews/shakacode/react_on_rails/1449#-N0VOHHx6EDAAKsFFT-i-r1-12:-N0itcL59PH1L4QRN9R_:b-avngyo) ([raw file](https://github.com/shakacode/react_on_rails/blob/af52b8aa838ca1d1ecefcd684a28e051bab5b92f/node_package/src/helpers/renderHelper.ts#L12)):*
<details><summary><i>Previously, tomdracz (Tom Dracz) wrote…</i></summary><blockquote>

Would be fairly trivial with top-level await though 🤔 
</blockquote></details>

Looks great so far!
___
*[`node_package/src/helpers/renderHelper.ts` line 8 at r2](https://reviewable.io/reviews/shakacode/react_on_rails/1449#-N0VOHHx6EDAAKsFFT-i-r2-8:-N0itSdi68mIV686KNhO:b27iims) ([raw file](https://github.com/shakacode/react_on_rails/blob/47912e2c89ff979508fc765c22b89b5e9bdc16bb/node_package/src/helpers/renderHelper.ts#L8)):*
<details><summary><i>Previously, tomdracz (Tom Dracz) wrote…</i></summary><blockquote>

This could be good based not only on React version but also some configuration option so it's an opt-in
</blockquote></details>

100% agree! Given what I wrote in the previous comment of us not being 100% sure that we're fully ready!

justin808:
@tomdracz we've got CI fixed. Is this one ready to go?

node_package/src/clientStartup.ts
alexeyr-ci1: Please see #1466.

alexeyr-ci1:
#1460 has been merged instead.

PR #1458, opened 2022-06-20
Only check the existence of i18n directory before locale generation

### LINKED ISSUE #1319: Remove check for i18n directory at startup

alexeyr:
Fixes #1319. Replaces #1439.


alexeyr:
If I add 
```
        expect do
          ReactOnRails.Locales.compile
        end.not_to raise_error
```
to https://github.com/shakacode/react_on_rails/blob/305957fd659944b7be9ea6ce4a45456651d95e2a/spec/react_on_rails/configuration_spec.rb#L99-L105, the test fails (because `locale_files` returns `nil`). Should it pass?

PR #1460, opened 2022-06-22
Require createRoot/hydrateRoot from 'react-dom/client' under react-dom 18+

### LINKED ISSUE #1441: React 18 Warning: You are importing hydrateRoot from "react-dom" [...] You should instead import it from "react-dom/client"

alexeyr:
Fixes #1441

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1460)


alexeyr:
@tomdracz @summera Can you take a look at this one? Especially if you have actual React 18 apps to test it with? OTOH, some generated examples use React 18 and the warning doesn't seem to be in the logs https://github.com/shakacode/react_on_rails/runs/7024950174?check_suite_focus=true

justin808:
@alexeyr LGTM!

node_package/src/reactHydrateOrRender.ts
```diff
@@ -0,0 +1,44 @@
...
+type HydrateOrRenderType = (domNode: Element, reactElement: ReactElement) => RenderReturnType;
+const supportsReactCreateRoot = parseInt(ReactDOM.version.split('.')[0], 10) >= 18;
+
+// TODO: once React dependency is updated to >= 18, we can remove this and just
+// import ReactDOM from 'react-dom/client';
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let reactDomClient: any;
+if (supportsReactCreateRoot) {
+  // eslint-disable-next-line camelcase
+  if (__webpack_require__) {
```
tomdracz: Wonder if this is the magic thing that was missing in the other attempts!

tomdracz (approved):
LGTM, not entirely sure still why this one works where the other approaches didn't but 🤷‍♂️

Don't have React 18 app on hand to test though, if we have test cases that use it, it's good enough

node_package/src/reactHydrateOrRender.ts
alexeyr-ci1: No, it isn't :) If I understand correctly (Webpack documentation is bad), https://github.com/shakacode/react_on_rails/pull/1448 was actually close to working, but Webpack needs a base directory for dynamic imports/requires. If it had 
```
require(`react-dom/${supportsReactCreateRoot ? 'client' : 'index'}`);
``` 
it would recognize `react-dom` as the base directory and work.

alexeyr-ci1: This `if` is needed because `yarn run test` in react_on_rails doesn't run under Webpack and `require.context` is not available.

alexeyr-ci1: > Don't have React 18 app on hand to test though, if we have test cases that use it, it's good enough

I've tried to update `spec/dummy` to React 18 and it doesn't work after all. Maybe the generated examples don't use Webpack?

justin808: How about if we just use a different import for react 18 with react on rails?

Check source for react-rails?

Google to see what other libraries do which have react as a dependency?

tomdracz: Some of the libraries I've seen that got direct dependency react-dom seem to have just released major versions that only support React 18 and left it at that. For all intents and purposes, React 18 components not initialized with the new root API should function the same as they do in React 17.

I do feel there's bit more at play here, think I've noted some other changes that might be required at one of the other PRs trying to tackle this.

alexeyr-ci1:
@tomdracz @justin808 The actual solution turns out to be entirely different. Please review again. See https://github.com/shakacode/react_on_rails/pull/1463 for the React 18 test.

justin808 (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,16 @@ Changes since last non-beta release.
...
+- Fixed the `You are importing hydrateRoot from "react-dom" [...] You should instead import it from "react-dom/client"` warning under React 18 ([#1441](https://github.com/shakacode/react_on_rails/issues/1441)). [PR 1460](https://github.com/shakacode/react_on_rails/pull/1460) by [alexeyr](https://github.com/alexeyr).
+
+  In exchange, you may see a warning like this when building a Webpack bundle under React 16:
+  ```
+  WARNING in ./node_modules/react-on-rails/node_package/lib/reactHydrateOrRender.js19:25-52
+  Module not found: Error: Can't resolve 'react-dom/client' in '/home/runner/work/react_on_rails/react_on_rails/spec/dummy/node_modules/react-on-rails/node_package/lib'
+   @ ./node_modules/react-on-rails/node_package/lib/ReactOnRails.js 34:45-78
+   @ ./client/app/packs/client-bundle.js 5:0-42 32:0-23 35:0-21 59:0-26
+  ```
+  It can be safely [suppressed](https://webpack.js.org/configuration/other-options/#ignorewarnings) in your Webpack configuration.
```
justin808, line 36: @ahangarha Let's better document this.

ahangarha: I am concerning about version 16 or 18. I think there is a mistake there. I asked Alexey for clarification.

PR #1463, opened 2022-06-27
Upgrade spec/dummy to React 18

alexeyr-ci1:


alexeyr-ci1:
Depends on #1460.

alexeyr-ci1:
`build-and-test` CI is a lot slower, not clear if it's permanent or corresponds to an actual slow-down.

PR #1475, opened 2022-08-04
Enable server rendered promises to resolve to react elements

Judahmeek:
...in addition to strings & objects.

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1475)

### DIFF
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -18,7 +18,7 @@ calls renderToString, that takes one parameter. You need to add an extra unused 
 as a renderFunction and not a simple React Function Component.`);
     throw error;
   }
-};
+}
 
 export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
   const { name, domNodeId, trace, props, railsContext, renderingReturnsPromises, throwJsErrors } = options;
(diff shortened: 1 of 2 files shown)


Judahmeek:
We can add an additional check like https://github.com/shakacode/react_on_rails/blob/master/node_package/src/createReactOutput.ts#L59 if you want.

Judahmeek:
I also should add a test case specifically for a promise that resolves into a react element

node_package/src/serverRenderReactComponent.ts
```diff
@@ -110,8 +110,12 @@ as a renderFunction and not a simple React Function Component.`);
       let promiseResult;
 
       try {
+        let awaitedResult = await renderResult;
+        if(typeof awaitedResult !== 'string' && typeof awaitedResult !== 'object'){
```
justin808, line 114: typeof a function is surely not object, correct?

is there any check that's more specific and positive (check `===` rather than `!==`)?

Judahmeek: No, typeof a function is `'function'`

I was thinking more of an object like `{ componentHtml: ..., some other react-helmet key-pairs }`

justin808:
Closing this because we're going to require that the value returned from a promise used for SSR is a String, and thus `ReactDOMServer.renderToString(element)` has already been called.

PR #1479, opened 2022-08-26
Removed keyword parameters warning. This will raise error in Ruby 3+ version.

alkesh26:
With the Ruby 2.7 version we are facing the below warning:

Screenshot 2022-08-26 at 3 25 53 PM

When using Ruby 3.1.2 the test will break with the below error:

Screenshot 2022-08-26 at 3 26 29 PM

Raised this PR to fix the warning in the `render_options_spec` file. We won't face any issues in Ruby 3+ versions.

Screenshot 2022-08-26 at 3 27 00 PM

Reference article: https://blog.saeloun.com/2019/10/07/ruby-2-7-keyword-arguments-redesign.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1479)

### DIFF
--- a/spec/react_on_rails/react_component/render_options_spec.rb
+++ b/spec/react_on_rails/react_component/render_options_spec.rb
@@ -22,7 +22,7 @@ describe ReactOnRails::ReactComponent::RenderOptions do
     attrs = the_attrs
 
     expect do
-      described_class.new(attrs)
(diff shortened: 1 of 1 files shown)


justin808:
LGTM. Thanks @alkesh26.

PR #1491, opened 2022-10-26
Add dev scripts to bin directory on install

### LINKED ISSUE #1482: Install script should create `./bin/dev`
- [ ] For parity with JSBundling, the installer script should create a simple shell script at `./bin/dev` with this contents:

```
#!/usr/bin/env bash
if ! command -v foreman &> /dev/null
then
  echo "Installing foreman…"
  gem install foreman
fi
foreman start -f Procfile.dev-static
```
- [ ] remember shell script must be set to be executable (`chmod 0755 ./bin/dev`)

- [ ] use the post-install message to direct people to this option

TODO: @jasonfb says he will PR this.

ahangarha:
Add handy script in `bin` directory to run Procfiles

Closes #1482

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1491)


ahangarha:
Some tests fail but not because of my changes.

docs/guides/tutorial.md
```diff
@@ -45,58 +45,57 @@ Then we need to create a fresh Rails application as following.
...
-# This is for Rails 7
+# For Rails 7.x
 rails new test-react-on-rails --skip-turbolinks --skip-javascript
 
 cd test-react-on-rails
 ```
 
 ## Add the shakapacker and react_on_rails gems
-To avoid issues regarding inconsistent gem and npm versions, you should specify the exact versions
+We recomment to use latest version of these gems. Otherwise, specify the exact versions
```
justin808: @ahangarha please spellcheck in your editor! 

`recommend`

lib/generators/react_on_rails/bin/dev
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+if ! command -v foreman &> /dev/null
+then
+  echo "Installing foreman…"
+  gem install foreman && echo "Foreman Installed."
+fi
+
+foreman start -f Procfile.dev
```
justin808: See my comment: https://github.com/rails/jsbundling-rails/issues/136

lib/generators/react_on_rails/install_generator.rb
```diff
@@ -75,6 +76,11 @@ def missing_node?
         GeneratorMessages.add_error(error)
         true
       end
+
+      def add_bin_scripts
+        directory "#{__dir__}/bin", "bin"
+        chmod "bin", 0o755 & ~File.umask, verbose: false
```
justin808: Why is this needed?

docs/guides/tutorial.md
ahangarha: My bad. Sorry for making the commit without proof check. Will update the word and even the sentence in a while.

lib/generators/react_on_rails/install_generator.rb
ahangarha: If you mean the chmod line, I got the idea from the code in shakapacker [binstubs](https://github.dev/shakacode/shakapacker/blob/30044b5a610ce9d0edb5eeb3de04903851c6e3ad/lib/install/binstubs.rb#L4) file. I thought about it and the only guess I was about issues might occur in Windows machines by loosing execution flag. But that also doesn't make much sense since `rails generate react_on_rails:install` runs once. And I believe it is a bad idea to run a command that can affect other irrelevant files.

I tried to go with a simpler solution of adding `bin/dev` to the root of the gem (as I read in the rubygem documentation) but it didn't work. So I took a look at shakapacker code and tried to get the working solution from there.

ahangarha: I removed `chmod` since I couldn't find any good reason to use it in this case. The file in the `bin` directory is already executable.

justin808:
Please add a CHANGELOG entry and I'll merge.

justin808:
@Judahmeek and @ahangarha, I'd like to merge, but we need to fix CI failures first.

```

An error occurred while installing psych (5.0.0), and Bundler cannot continue.

In Gemfile:
  sdoc was resolved to 2.4.0, which depends on
    rdoc was resolved to 6.5.0, which depends on
      psych
Error: Process completed with exit code 5.
```

ahangarha:
> I'd like to merge, but we need to fix CI failures first.

@justin808 The issue comes from `psych v5+` and in particular from Related this PR: https://github.com/ruby/psych/pull/541.

Should we install `libyaml-dev` directly in our CI?

I opened [an issue](https://github.com/ruby/psych/issues/602) on their repo to get their insight.

lib/generators/react_on_rails/bin/dev
justin808: Let's do this: https://github.com/rails/webpacker/blob/6bd3b4b25a69288b0e9408c848f56b1cf50946da/lib/tasks/webpacker/check_yarn.rake#L6

then we use overmind first or foreman second

then display message to either install overmind or foreman with the command for each...

then once we merge, we can submit a PR to jsbundling-rails

ahangarha:
@justin808 
Now not only we prefer Overmind over Foreman, but also the script is written in Ruby so it is expected to work cross platform.

ahangarha:
@justin808 
Shouldn't we add these scripts to the dummy app as well?
I think we should.

Judahmeek:
@ahangarha can you squash this & rebase it?

We should be able to merge it after that.

ahangarha:
@Judahmeek 
It was challenging to fix the mess I made in the commit history, but I think I did it well.
I think this PR is ready to be merged.

Judahmeek (approved):
> Reviewed 1 of 4 files at r1, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1491)* status: all files reviewed, 3 unresolved discussions (waiting on @ahangarha and @justin808)

justin808 (approved):

PR #1495, opened 2022-12-05
Update instruction on an existing Rails app

ahangarha:
Updated the instruction to refer to `shakapacker` rather `webpacker`.

I recommend merging #1491 before this instruction and updating this guide based on those changes otherwise, we need to make a new PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1495)

### DIFF
--- a/docs/guides/installation-into-an-existing-rails-app.md
+++ b/docs/guides/installation-into-an-existing-rails-app.md
@@ -15,13 +15,13 @@
 
    ```bash
    bundle add react_on_rails --version=13.1.0 --strict
-   bundle add shakapacker
+   bundle add shakapacker --strict
    ```
 
 2. Run the following 2 commands to install Webpacker with React. Note, if you are using an older version of Rails than 5.1, you'll need to install webpacker with React per the instructions [here](https://github.com/rails/webpacker).
 
    ```bash
-   bundle exec rails webpacker:install
+   rails webpacker:install
    ```
 
 3. Commit this to git (or else you cannot run the generator unless you pass the option `--ignore-warnings`).
--- a/docs/guides/installation-into-an-existing-rails-app.md
+++ b/docs/guides/installation-into-an-existing-rails-app.md
@@ -37,15 +37,16 @@
    ```bash
    rails generate react_on_rails:install --help
    ```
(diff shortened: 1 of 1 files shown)


Judahmeek (approved):

docs/guides/installation-into-an-existing-rails-app.md
```diff
@@ -7,48 +7,45 @@
...
-_Use the latest version for react_on_rails._
+   Or use `bundle add`:
 
-```
-bundle add webpacker                 
-bundle add react_on_rails --version=12.0.4 --strict
-```
+   ```bash
+   $ bundle add react_on_rails --version=13.1.0 --strict
+   $ bundle add shakapacker
```
justin808: @ahangarha what is the `$` for? your personal prompt?

ahangarha: Thanks for bringing this up. This was one of my concerns and since it was minor, I didn't bring it up. In our docs, sometimes we have this `$` sign (which represents the shell of a non-root user). To me, not only this is not necessary (the user should know which code is to run in the shell and which one to is to add to a file), but it is annoying especially when one wants to copy/paste the code block.

It should be enough to have syntax highlighting.

Now I am confident to remove `$` from all such cases.

ahangarha: @justin808 Done

justin808:
@ahangarha please resolve conflicts.

spec/dummy/Gemfile.lock
```diff
@@ -11,67 +11,73 @@ PATH
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (7.0.0)
-      actionpack (= 7.0.0)
-      activesupport (= 7.0.0)
+    actioncable (7.0.4.2)
```
justin808: gem version items should be in separate PR!

justin808 (changes requested):

spec/dummy/Gemfile.lock
ahangarha: @justin808 I removed the commit. I am not sure why I pushed that commit.

PR #1496, opened 2022-12-14
Use RSpec's when_first_matching_example_defined if available

mcls:
This RSpec method was added in v3.5.0 to with expensive setup logic. It results in similar behavior as just using `before(:example, ...)`, but it will only run once instead of for each example. So the behavior remains the same, but with the small optimization of not having to check for compiled assets on every example.

Related documentation:
- https://relishapp.com/rspec/rspec-core/v/3-12/docs/hooks/when-first-matching-example-defined-hook

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1496)


justin808:
@mcls 
1. What is the difference between doing this and a `before(:suite)`?
2. Assuming this is needed, please update the CHANGELOG.md and update the docs, https://github.com/shakacode/react_on_rails/blob/master/docs/guides/rspec-configuration.md,  to show how to use this change.

mcls:
Hi Justin,

> 1. What is the difference between doing this and a before(:suite)?

`before(:suite)` does not support metadata filtering and thus doesn't support the same conditional of running setup code like `when_first_matching_example_defined` does. I was initially confused by this too, but you can find some additional context here in the original RSpec PR which introduces the feature: https://github.com/rspec/rspec-core/pull/2175

> 2. Assuming this is needed, please update the CHANGELOG.md and update the docs, https://github.com/shakacode/react_on_rails/blob/master/docs/guides/rspec-configuration.md, to show how to use this change.

I have updated the `CHANGELOG.md` and pushed an updated commit. Regarding the docs, I just checked but I don't think they need to be updated because this is just an internal change to `ReactOnRails::TestHelper.configure_rspec_to_compile_assets`.

justin808 (approved):
Looks fine. 

Just want to check that somebody tested this with the versions where `when_first_matching_example_defined` is not available.

And we need to resolve conflict in the changelog.

justin808:
@Judahmeek or @ahangarha Please merge this after resolving the conflict.

ahangarha:
@justin808 I don't have write access to this repo so I cannot resolve the conflict and merge.

justin808:
@mcls  @ahangarha can I get confirmation:

> Just want to check that somebody tested this with the versions where when_first_matching_example_defined is not available.

and then I'll merge.

PR #1498, opened 2022-12-21
Fix reactOnRailsPageUnloaded when there is no component on the page

NhanHo:
In `reactOnRailsPageLoaded`, if we finds no react component, the function [return ](https://github.com/NhanHo/react_on_rails/blob/f121906bf94a7f84ad4dde20c1cd7e93fa75f550/node_package/src/clientStartup.ts#L209) and `context.roots` is not initialized. This cause `findContext().roots` in `reactOnRailsPageUnloaded` to be undefined, and `for .. of` throws an exception.

This should mostly affect app using both hotwire and react_on_rails, as the recommended way to load js is to do the same file everywhere, otherwise the issue can be avoided by just not loading react_on_rails

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1498)

PR #1499, opened 2022-12-23
Fix reactOnRailsPageUnloaded when there is no component on the page

justin808:
From https://github.com/shakacode/react_on_rails/pull/1498.

In `reactOnRailsPageLoaded`, if we finds no react component, the function [return ](https://github.com/NhanHo/react_on_rails/blob/f121906bf94a7f84ad4dde20c1cd7e93fa75f550/node_package/src/clientStartup.ts#L209) and `context.roots` is not initialized. This cause `findContext().roots` in `reactOnRailsPageUnloaded` to be undefined, and `for .. of` throws an exception.

This should mostly affect app using both hotwire and react_on_rails, as the recommended way to load js is to do the same file everywhere, otherwise the issue can be avoided by just not loading react_on_rails

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1499)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,14 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+### [13.2.0] - 2022-12-23
+                                           
+### Fixed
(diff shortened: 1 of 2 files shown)


justin808:
Merged #1498.

PR #1506, opened 2023-01-16
Generate Pack if Missing Generated Pack for Registered Component

pulkitkkr:
## Problem
Currently, The Generate Pack Script handles the on-demand generation of packs only for existing components by checking m_time to verify if the generated packs are stale. If a user creates a new Component under `components_subdirectory,` It's not tracked by `stale_generated_component_packs`; 

## Solution and Changes in PR
This PR updates the FS-based Pack generation feature by adding a check for missing_packs. The system will generate a new pack when either the components are stale, or a new component is created.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1506)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,14 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
-- Added `./bin/dev` and `./bin/dev-static` executables to ease and standardize running the dev server. [PR 1491](https://github.com/shakacode/react_on_rails/pull/1491) by [ahangarha](https://github.com/ahangarha)
+
+### [13.3.0] - 2022-01-28
+### Fixed
 - Fixed pack not found warning while using `react_component` and `react_component_hash` helpers, even when corresponding chunks are present. [PR 1511](https://github.com/shakacode/react_on_rails/pull/1511) by [pulkitkkr](https://github.com/pulkitkkr)
 - Fixed FS-based packs generation functionality to trigger pack generation  on the creation of a new react component inside `components_subdirectory`. [PR 1506](https://github.com/shakacode/react_on_rails/pull/1506) by [pulkitkkr](https://github.com/pulkitkkr)
+
+### Added
+- Added `./bin/dev` and `./bin/dev-static` executables to ease and standardize running the dev server. [PR 1491](https://github.com/shakacode/react_on_rails/pull/1491) by [ahangarha](https://github.com/ahangarha)
 
 ### [13.2.0] - 2022-12-23
                                            
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Changes since last non-beta release.
 *Please add entries here for your pull requests that are not yet released.*
 - Added `./bin/dev` and `./bin/dev-static` executables to ease and standardize running the dev server. [PR 1491](https://github.com/shakacode/react_on_rails/pull/1491) by [ahangarha](https://github.com/ahangarha)
 - Fixed pack not found warning while using `react_component` and `react_component_hash` helpers, even when corresponding chunks are present. [PR 1511](https://github.com/shakacode/react_on_rails/pull/1511) by [pulkitkkr](https://github.com/pulkitkkr)
+- Fixed FS-based packs generation functionality to trigger pack generation  on the creation of a new react component inside `components_subdirectory`. [PR 1506](https://github.com/shakacode/react_on_rails/pull/1506) by [pulkitkkr](https://github.com/pulkitkkr)
 
 ### [13.2.0] - 2022-12-23
                                            
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -93,27 +93,6 @@ module ReactOnRails
       end
     end
 
-    def load_pack_for_component(component_name)
-      is_component_pack_present = File.exist?(generated_components_pack_path(component_name))
-      is_development = ENV["RAILS_ENV"] == "development"
-
-      if is_development && !is_component_pack_present
-        ReactOnRails::PacksGenerator.generate
-        raise_missing_pack_error(component_name)
-      end
-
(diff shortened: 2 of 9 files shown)


justin808 (changes requested):
Looks OK, but where are the tests that break without the fix and pass with the fix?

pulkitkkr:
> Looks OK, but where are the tests that break without the fix and pass with the fix?

This is a feature that Get's Triggered Only If someone creates a new component while Rails Server is still running. I don't know if we have test cases to cover such scenarios or if we should plan to add them.

lib/react_on_rails/helper.rb
```diff
@@ -556,8 +556,7 @@ def initialize_redux_stores
 
     def raise_generated_missing_pack_warning(component_name)
       msg = <<~MSG
-        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please refresh the webpage \
-        once webpack has finished generating the bundles. If the problem persists
+        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please restart the webpack dev server to ensure webpack generates the chunks corresponding to #{component_name} component. If the problem persists
```
justin808: or webpack in watch mode

lib/react_on_rails/packs_generator.rb
```diff
@@ -23,13 +23,11 @@ def self.raise_nested_entries_disabled
...
 
-      raise_webpacker_not_installed unless ReactOnRails::WebpackerUtils.using_webpacker?
-      raise_shakapacker_version_incompatible unless shackapacker_version_requirement_met?
-      raise_nested_entries_disabled unless ReactOnRails::WebpackerUtils.nested_entries?
+      verify_configuration
 
       is_generated_directory_present = Dir.exist?(generated_packs_directory_path)
 
-      return if is_generated_directory_present && webpack_assets_status_checker.stale_generated_component_packs.empty?
+      return if is_generated_directory_present && stale_or_missing_packs.empty?
```
justin808: is the stale_or_missing_packs list used to optimize what's generated?

justin808: If not, then why keep going after one is detected?

justin808 (changes requested):
where are the tests?

lib/react_on_rails/helper.rb
pulkitkkr: We can not run webpack in watch mode as creating a new Component makes a new entry point. To accommodate this new entry point, the server needs to get restarted as setEntry to webpack is configured on initialization of webpack.

lib/react_on_rails/packs_generator.rb
pulkitkkr: This optimization should be a separate PR as it requires additional testing and might introduce new bugs if not done correctly. Considering the time sensitivity of this work, I would suggest creating a new Issue for optimization, which we can work on later.

lib/react_on_rails/helper.rb
justin808: that's my point

the error says webpack dev server only

lib/react_on_rails/packs_generator.rb
justin808: then exit early on the first found stale or missing entry

pulkitkkr: We will use this list later to generate packs just for them. I can exit it early by making the `stale_or_missing_packs` return boolean. But since we are going to use it very soon, I suggest keeping it this way for now.

lib/react_on_rails/helper.rb
pulkitkkr: Fixed

lib/react_on_rails/helper.rb
```diff
@@ -556,8 +556,7 @@ def initialize_redux_stores
 
     def raise_generated_missing_pack_warning(component_name)
       msg = <<~MSG
-        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please refresh the webpage \
-        once webpack has finished generating the bundles. If the problem persists
+        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please restart the webpack dev server or webpack in watch mode, to ensure webpack generates the chunks corresponding to #{component_name} component. If the problem persists
```
justin808, line 560: Message says ERROR but method is named warning.

spec/dummy/spec/packs_generator_spec.rb
```diff
@@ -4,6 +4,8 @@
 
 # rubocop:disable Metrics/ModuleLength
 module ReactOnRails
+  GENERATED_PACKS_OUTPUT_REGEX = /Generated Packs:/.freeze
```
justin808: Why a regex? Seems like a simple string match.

justin808 (changes requested):

spec/dummy/spec/packs_generator_spec.rb
pulkitkkr: Updated to String

lib/react_on_rails/helper.rb
```diff
@@ -555,10 +555,9 @@ def initialize_redux_stores
       result
     end
 
-    def raise_generated_missing_pack_warning(component_name)
+    def raise_missing_pack_error(component_name)
       msg = <<~MSG
-        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please refresh the webpage \
-        once webpack has finished generating the bundles. If the problem persists
+        **ERROR** ReactOnRails: Generated missing pack for Component: #{component_name}. Please restart the webpack dev server or webpack in watch mode, to ensure webpack generates the chunks corresponding to #{component_name} component. If the problem persists
```
justin808, line 560: should this error message be detected in some tests?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Changes since last non-beta release.
 *Please add entries here for your pull requests that are not yet released.*
 - Added `./bin/dev` and `./bin/dev-static` executables to ease and standardize running the dev server. [PR 1491](https://github.com/shakacode/react_on_rails/pull/1491) by [ahangarha](https://github.com/ahangarha)
```
justin808: 1. Let's make this 13.3.0
2. Add a `Fixed` and `Added` sections. See below, line 24 and 28.

pulkitkkr: Done! Added a commit

PR #1509, opened 2023-01-22
Separate pack generation helper methods into separate module

Judahmeek:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1509)


docs/guides/configuration.md
```diff
@@ -166,6 +166,8 @@ ReactOnRails.configure do |config|
   # For automated component registry, `render_component` view helper method tries to load bundle for component from
   # generated directory. default is false, you can pass option at the time of individual usage or update the default
   # in the following line
+
+  # Enabling this setting without including `ReactOnRails::WebpackerHelper` in your `ApplicationHelper` will result in a fatal error
```
justin808: What does this mean?

docs/guides/file-system-based-automated-bundle-generation.md
```diff
@@ -36,6 +36,13 @@ You can change the value in `config/initializers/react_on_rails` by updating it
 config.auto_load_bundle = true
 ```
 
+### Update `ApplicationHelper`
+Due to being dependent on `shakapacker`, pack generation logic is stored in a separate module, which must be included in your `ApplicationHelper`:
+
+```rb
+  include ReactOnRails::WebpackerHelper
+```
+
```
justin808: Completely unnecessary because we can check if constants are defined in Ruby.

lib/react_on_rails/webpacker_helper.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "webpacker"
+
+module ReactOnRails
+  module WebpackerHelper
+    include Webpacker::Helper
```
justin808: Call this Shakapacker

spec/dummy/app/helpers/application_helper.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  include ReactOnRails::WebpackerHelper
```
justin808: Need to remove.

justin808 (changes requested):
Overall, the refactoring should be:

1. Conditional inclusion of functionality based on updating this utility file: https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/webpacker_utils.rb (not helper!)
2. Consider moving code related to auto-packs generation to its own module as much as possible.

If one tries to use new feature but has incompatible Shakapacker version, throw an error.

See code of checking React on Rails gem version for example of how to check shakapacker version, version checker.

lib/react_on_rails/configuration.rb
```diff
@@ -123,6 +124,18 @@ def setup_config_values
 
     private
 
+    def check_auto_load_requirements_if_auto_load
+      return if auto_load_bundle != true
```
pulkitkkr: ```return unless auto_load_bundle```

docs/guides/file-system-based-automated-bundle-generation.md
```diff
@@ -180,6 +180,8 @@ The default value of the `auto_load_bundle` parameter can be specified by settin
 
 If server rendering is enabled, the component will be registered for usage both in server and client rendering. In order to have separate definitions for client and server rendering, name the component files as `ComponentName.server.jsx` and `ComponentName.client.jsx`. The `ComponentName.server.jsx` file will be used for server rendering and the `ComponentName.client.jsx` file for client rendering. If you don't want the component rendered on the server, you should only have the `ComponentName.client.jsx` file.
 
+Once generated, all server entrypoints will be imported into a file named `[ReactOnRails.configuration.server_bundle_js_file]-generated.js`, which in turn will be imported into a source file named the same as `ReactOnRails.configuration.server_bundle_js_file`. If your server bundling logic is such that your server bundle source entrypoint is not named the same as your `ReactOnRails.configuration.server_bundle_js_file` & changing it would be difficult, please let us know.
+
```
pulkitkkr, line 184: For conditions when server bundle source entrypoint is not named the same as your `ReactOnRails.configuration.server_bundle_js_file`. What should be the solution? Should we document that? I know @frullah faced this issue and had a working solution.

spec/react_on_rails/configuration_spec.rb
```diff
@@ -213,12 +213,12 @@ module ReactOnRails
       expect(ReactOnRails.configuration.auto_load_bundle).to eq(false)
     end
 
-    it "changes the configuration of the gem, such as setting the auto_load_bundle option to true" do
+    it "changes the configuration of the gem, such as setting the components_subdirectory option to 'something" do
```
pulkitkkr: Why are we removing the test for `auto_load_bundle `? How about having both?

pulkitkkr (changes requested):

docs/guides/file-system-based-automated-bundle-generation.md
justin808: There should be no reason that the name needs to be different, AFAIK.

@frullah?

lib/react_on_rails/configuration.rb
```diff
@@ -243,6 +256,16 @@ def configure_skip_display_none_deprecation
...
     end
+
+    def raise_missing_components_subdirectory
+      msg = <<~MSG
+        **ERROR** ReactOnRails: auto_load_bundle is set to true, yet components_subdirectory is unconfigured.\
+        Please set components_subdirectory to the desired directory.  For more information, please see \
+        https://www.shakacode.com/react-on-rails/docs/guides/file-system-based-automated-bundle-generation.md
+      MSG
+
+      raise ReactOnRails::Error, msg
```
justin808, line 267: @pulkitkkr, this error looks helpful!

spec/react_on_rails/configuration_spec.rb
justin808: Could leave test for both.

justin808 (changes requested):
Looks mainly OK.

lib/react_on_rails/configuration.rb
justin808: This is checking the wrong value. 

Should check if `components_subdirectory` `present?`
```
  # components_subdirectory is the name of the subdirectory matched to detect and register components automatically
  # The default is nil. You can enable the feature by updating it in the next line.
  config.components_subdirectory = "ror_components"
```

justin808 (changes requested):

lib/react_on_rails/configuration.rb
```diff
@@ -123,6 +124,18 @@ def setup_config_values
 
     private
 
+    def check_auto_load_requirements_if_auto_load
+      return if auto_load_bundle != true
+
+      raise_missing_components_subdirectory unless components_subdirectory.present?
```
justin808: Only continue checks if this one is present.

Throw if auto_load_bundle is true.

order matters, b/c it's fine for auto_load_bundle **DEFAULT** to be false and used only in a few places.

lib/react_on_rails/packs_generator.rb
```diff
@@ -6,44 +6,30 @@ module ReactOnRails
...
+    MINIMUM_SHAKAPACKER_VERSION = [6, 5, 1].freeze
 
-    def self.raise_nested_entries_disabled
-      packs_generator = PacksGenerator.new
-      packs_generator.raise_nested_entries_disabled
+    def self.instance
+      @instance ||= PacksGenerator.new
     end
 
     def verify_setup_and_generate_packs
```
justin808: 1. change name to generate_packs_if_stale
2. remove checks. those are done at startup
3. move check logic to configuration file

spec/dummy/.prettierignore
```diff
@@ -2,3 +2,7 @@ node_modules
 public/webpack/
 *.bs.js
 lib/bs/**
+
+# File Generated by ROR FS-based Registry
+/client/app/packs/server-bundle-generated.js
+/client/app/packs/generated
```
justin808: missing LF

justin808 (approved):

docs/guides/file-system-based-automated-bundle-generation.md
frullah: @pulkitkkr my issue wasn't related to this, it was different: RoR requires _public_ `server_bundle_js_file` to be present before generating packs. I was able to use generated `server_bundle_js_file` as a server bundle entry point.

PR #1520, opened 2023-02-08
Reduce the dependencies on rails

lephyrius:
Reduce the dependencies so we don't need the full rails like action cable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1520)


justin808:
We need documentation as to why this is beneficial.

And linting failed.

justin808:
I'll reopen if there are good reasons.

PR #1518, opened 2023-02-02
Changes required for RoR Pro caching

Judahmeek:
Changes required for https://github.com/shakacode/react_on_rails_pro/pull/306

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1518)

PR #1527, opened 2023-02-24
Revert "Changes required for RoR Pro caching (#1518)"

justin808:
This reverts commit 6753c84da3e981a0d49d39d1fa88f00a8df3e2c7.

This is reverted to push out the release without this change.

### DIFF
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -313,6 +313,10 @@ module ReactOnRails
       @rails_context.merge(serverSide: server_side)
     end
 
+    # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity
+
+    private
+
     def load_pack_for_component(component_name)
       is_component_pack_present = File.exist?(generated_components_pack_path(component_name))
       is_development = ENV["RAILS_ENV"] == "development"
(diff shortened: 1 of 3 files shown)


lib/react_on_rails/helper.rb
```diff
@@ -327,7 +327,8 @@ def load_pack_for_component(component_name)
       end
 
       ReactOnRails::WebpackerUtils.raise_nested_entries_disabled unless ReactOnRails::WebpackerUtils.nested_entries?
-
+      puts "ZYXXYZ" if @self.nil?
```
justin808: why `@self` and not `self`?

Judahmeek: Because I only want line 330~331 to run the debug statement once.

.gitignore
```diff
@@ -19,8 +19,6 @@
 
 node_modules
 
-/node_package/lib
```
justin808: Why are you removing this?

justin808: These are the source files: https://github.com/shakacode/react_on_rails/tree/master/node_package/src

lib is generated.

Judahmeek: So I can test the branch without getting you to make a new release.

justin808:
Please advise me if this is should be merged soon.

PR #1531, opened 2023-03-18
Judahmeek/move generated out of packs

Judahmeek:
### Summary

Moves `generated-server-bundle` outside of packs so there isn't a redundant server bundle entrypoint in packs.


justin808:
@Judahmeek, the  `generated-server-bundle` is something that we'll have run through webpack, so it's appropriate in the packs directory.

Judahmeek:
> @Judahmeek, the `generated-server-bundle` is something that we'll have run through webpack, so it's appropriate in the packs directory.

There's no reason for it to be in the packs directory though as it is completely redundant.

It's only created if server-bundle exists & it is imported by server-bundle.

Placing it in the packs directory simply means that the user must delete another key within their client configuration:
```
  delete clientConfig.entry['server-bundle'];
  delete clientConfig.entry['server-bundle-generated'];
```

spec/dummy/.gitignore
```diff
@@ -37,5 +37,5 @@ yarn-debug.log*
 *.bs.js
 
 # File Generated by ROR FS-based Registry
-/client/app/packs/server-bundle-generated.js
+/client/app/server-bundle-generated.js
```
justin808: `/client/app/generated/server-bundle-generated.js`

and the generated directory is ignored...

lib/react_on_rails/packs_generator.rb
```diff
@@ -93,9 +95,11 @@ def add_generated_pack_to_server_bundle
     end
 
     def generated_server_bundle_file_path
+      return server_bundle_entrypoint if ReactOnRails.configuration.make_generated_server_bundle_the_entrypoint
+
       generated_server_bundle_file_path = server_bundle_entrypoint.sub(".js", "-generated.js")
       generated_server_bundle_file_name = component_name(generated_server_bundle_file_path)
-      source_entry_path = ReactOnRails::WebpackerUtils.webpacker_source_entry_path
+      source_entry_path = Pathname(ReactOnRails::WebpackerUtils.webpacker_source_entry_path).parent
```
justin808: subdir of `generated`

lib/react_on_rails/helper.rb
```diff
@@ -324,8 +324,10 @@ def load_pack_for_component(component_name)
```
justin808: we should rename to `load_pack_for_generated_component`

changes to support non-generated don't make sense b/c you can have many components defined in one bundle if doing by hand.

lib/react_on_rails/helper.rb
```diff
@@ -324,8 +324,10 @@ def load_pack_for_component(component_name)
 
       ReactOnRails::WebpackerUtils.raise_nested_entries_disabled unless ReactOnRails::WebpackerUtils.nested_entries?
 
-      append_javascript_pack_tag "generated/#{component_name}"
-      append_stylesheet_pack_tag "generated/#{component_name}"
+      prefix = is_component_pack_present ? "generated/" : ""
+
+      append_javascript_pack_tag "#{prefix}#{component_name}"
```
justin808: see above comment.

spec/dummy/.gitignore
justin808: this is inconsistent with https://github.com/shakacode/react_on_rails/pull/1531/files#diff-cb9a4038d8838c04798d3da7dc2b785dffe109b7f9b903e354601e5cd5a9efe6R39

justin808 (changes requested):

rakelib/dummy_apps.rake
```diff
@@ -16,7 +16,12 @@ namespace :dummy_apps do
     bundle_install_in(dummy_app_dir)
   end
 
-  task dummy_apps: %i[dummy_app node_package] do
+  task :generate_packs do
+    dummy_app_dir = File.join(gem_root, "spec/dummy")
+    sh_in_dir(dummy_app_dir, "bundle exec rake react_on_rails:generate_packs")
```
justin808, line 21: can we document this should be done for test environments before the bundles are built?

spec/dummy/client/app/components/ImageExample/ImageExample.module.scss
```diff
@@ -3,7 +3,7 @@
 }
 
 .background {
-  background-image: url(assets/images/section-title-bg.png);
+background-image: url(assets/images/section-title-bg.png);
```
justin808: bad change?

spec/dummy/app/views/layouts/application.html.erb
```diff
@@ -10,7 +10,7 @@
   <%= yield :head %>
 
   <!--  NOTE: Must use defer and not async to keep async scripts loading in correct order  -->
-  <%= javascript_pack_tag('client-bundle', 'data-turbolinks-track': true, defer: true) %>
+  <%= javascript_pack_tag('client-bundle', 'data-turbolinks-track': true, defer: false) %>
```
justin808, line 13: why is this changing?

see comment on line 12

spec/dummy/client/app/startup/ImageExample.jsx
```diff
@@ -1,39 +1,40 @@
...
-import blueprintIcon from './blueprint_icon.svg';
-
-const TestComponent = (_props) => (
+const TestComponent = (props) => (
   <div>
     <h1 className={css.red}>This is a test of CSS module color red.</h1>
     <hr />
     <h1 className={css.background}>Here is a label with a background-image from the CSS modules imported</h1>
-    <img src={logo} alt="logo" />
+    <img src={logo} />
```
justin808: why change these? alt is for accessibility

spec/dummy/config/initializers/react_on_rails.rb
```diff
@@ -36,4 +36,5 @@ def self.adjust_props_for_client_side_hydration(component_name, props)
   config.rendering_props_extension = RenderingPropsExtension
   config.components_subdirectory = "startup"
   config.auto_load_bundle = true
+  config.defer_generated_component_packs = false
```
justin808, line 39: where is the documentation for this?

spec/dummy/config/webpacker.yml
```diff
@@ -56,7 +56,7 @@ development:
 
 test:
   <<: *default
-  compile: true
+  compile: false
```
justin808, line 59: did you switch to the react on rails helper?

if so, we should encourage that in the docs

justin808 (changes requested):
Looks really good. I had a couple of questions and comments.

rakelib/dummy_apps.rake
Judahmeek: Done

spec/dummy/app/views/layouts/application.html.erb
Judahmeek: I canceled `defer` on all the javascript_pack_tags. No async, no defer.

spec/dummy/client/app/components/ImageExample/ImageExample.module.scss
Judahmeek: Yes

spec/dummy/client/app/startup/ImageExample.jsx
Judahmeek: Because I copied the file from https://github.com/shakacode/react_on_rails_pro/blob/master/spec/dummy/client/app/components/ImageExample/ImageExample.js

spec/dummy/config/initializers/react_on_rails.rb
Judahmeek: Added

spec/dummy/config/webpacker.yml
Judahmeek: I did not.

Until I run some tests, I don't trust either mtime checker to work with CI right.

Of course, I can't say that it doesn't work. I just remember seeing weird behavior that might have been related.

docs/guides/configuration.md
```diff
@@ -178,17 +178,21 @@ ReactOnRails.configure do |config|
...
 
-  # default is false
+  # Default is false
   # Set this to true & instead of trying to import the generated server components into your existing
   # server bundle entrypoint, the PacksGenerator will create a server bundle entrypoint using
   # config.server_bundle_js_file for the filename.
   config.make_generated_server_bundle_the_entrypoint = false
 
+  # Default is false
+  # Set this to true to have `defer: false` added to your `append_javascript_pack` calls for generated entrypoints.
```
justin808: This comment is very confusing.

Default is for `defer` is true or false if not used?

Setting to `true` gives you the default defer?

docs/guides/file-system-based-automated-bundle-generation.md
```diff
@@ -185,6 +185,14 @@ Once generated, all server entrypoints will be imported into a file named `[Reac
 
 *Note: If specifying separate definitions for client and server rendering, please make sure to delete the generalized `ComponentName.jsx` file.*
 
+### Integrating auto-bundling into CI workflows
+
+Currently, ReactOnRails contains conditional logic that checks for the existence of generated entrypoint files whenever `react_component` or one of its derivative methods are called. This conditional logic is used to determine the relative file path to provide `append_javascript_pack` as an argument.
```
justin808: This should not be true in production. Production should assume it's built.

The conditional is to generate the pack if needed during development, not to determine the path.

docs/guides/file-system-based-automated-bundle-generation.md
```diff
@@ -185,6 +185,14 @@ Once generated, all server entrypoints will be imported into a file named `[Reac
...
+
+This means that `rake react_on_rails:generate_packs` or its programmatic equivalent must be run as a prerequisite to any sort of test or spec that would result in `react_component` or one of its derivative methods being called, even if the generated entrypoint files have already been bundled in a previous workflow/job.
+
+Caching of the generated entrypoints would likely also resolve this issue.
```
justin808: we need to mention the react on rails spec helper will do this automatically

thus, it should be used and not `compile: true` in the shakapacker setup.

justin808 (changes requested):

docs/guides/configuration.md
Judahmeek: My mistake. I went off memory instead of looking at the code I had wrote. Corrected.

docs/guides/file-system-based-automated-bundle-generation.md
Judahmeek: You're partially correct. I was wrong that the conditional logic is for determining the path.

To support "legacy" components, I removed that conditional logic & replaced it with conditional logic that just skips adding an `append_javascript_pack` call if requested component doesn't match any generated entrypoint.

That's why `load_pack_for_component` was renamed to `load_pack_for_generated_component`.

docs/guides/file-system-based-automated-bundle-generation.md
Judahmeek: While the `TestHelper` does generate packs if they are stale, it only uses mtime to check & doesn't have a digest strategy available, like Shakapacker does.

Before we push users to switch, would you like me to investigate having `TestHelper` use Shakapacker's stake-checking strategies so that we can reuse code & specs?

The [comment documentation looks to be out-of-date](https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/test_helper.rb#L56) (method names are wrong), but looks like there's already a way to add a custom asset_checker.

We'll just probably need an adapter module.

docs/guides/configuration.md
```diff
@@ -189,9 +189,9 @@ ReactOnRails.configure do |config|
   # config.server_bundle_js_file for the filename.
   config.make_generated_server_bundle_the_entrypoint = false
 
-  # Default is false
-  # Set this to true to have `defer: false` added to your `append_javascript_pack` calls for generated entrypoints.
-  config.defer_generated_component_packs = false
+  # Default is true, which matches Webpacker/Shakapacker's defer default for `append_javascript_pack`
+  # Set this to false to have `defer: false` added to your `append_javascript_pack` calls for generated entrypoints.
+  config.defer_generated_component_packs = true
```
justin808, line 194: double-check that the docs reflect this...

justin808 (approved):

docs/guides/configuration.md
Judahmeek: That is the docs

PR #1539, opened 2023-05-25
Stop using defer_generated_component_packs in spec/dummy

Judahmeek:
Comments out `config.defer_generated_component_packs = false` in `spec/dummy/config/initializers/react_on_rails.rb` since it is apparently no longer neccessary for CI to pass.

Superseded by #1540

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1539)


spec/dummy/config/initializers/react_on_rails.rb
```diff
@@ -36,5 +36,5 @@ def self.adjust_props_for_client_side_hydration(component_name, props)
   config.rendering_props_extension = RenderingPropsExtension
   config.components_subdirectory = "startup"
   config.auto_load_bundle = true
-  config.defer_generated_component_packs = false
+  # config.defer_generated_component_packs = false
```
justin808, line 39: Commented out should be the default with an explanation of why change this. Typically, that is b/c we're loading jQuery and inline HTML scripts need the JS code loaded first.

justin808 (changes requested):

PR #1540, opened 2023-05-27
Make sure renderInit only runs after DOMContentLoaded

Judahmeek:
Fixed the race condition revealed by https://github.com/shakacode/react_on_rails/pull/1539

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1540)

### DIFF
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -309,7 +309,7 @@ export function clientStartup(context: Context): void {
   // but sub-resources such as images, stylesheets and frames are still loading.
   // If lazy asynch loading is used, such as with loadable-components, then the init
   // function will install some handler that will properly know when to do hyrdation.
-  if (document.readyState == 'complete') {
+  if (document.readyState === 'complete') {
     window.setTimeout(renderInit);
   } else {
     document.addEventListener('DOMContentLoaded', renderInit);
--- a/node_package/src/clientStartup.ts
+++ b/node_package/src/clientStartup.ts
@@ -286,6 +286,17 @@ function isWindow(context: Context): context is Window {
   return (context as Window).document !== undefined;
 }
 
+function onPageReady(callback: () => void) {
+  if (document.readyState === "complete") {
+    callback();
+  } else {
+    document.addEventListener("readystatechange", function onReadyStateChange() {
+        onPageReady(callback);
+        document.removeEventListener("readystatechange", onReadyStateChange);
+    });
+  }
+}
+
 export function clientStartup(context: Context): void {
   // Check if server rendering
   if (!isWindow(context)) {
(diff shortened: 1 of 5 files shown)


node_package/src/clientStartup.ts
```diff
@@ -309,7 +309,7 @@ export function clientStartup(context: Context): void {
   // but sub-resources such as images, stylesheets and frames are still loading.
```
justin808: We need to change lines 307-309 to reflect the new information.

node_package/src/clientStartup.ts
```diff
@@ -304,11 +304,6 @@ export function clientStartup(context: Context): void {
 
   debugTurbolinks('Adding DOMContentLoaded event to install event listeners.');
 
-  // So  long as the document is not loading, we can assume:
-  // The document has finished loading and the document has been parsed
-  // but sub-resources such as images, stylesheets and frames are still loading.
-  // If lazy asynch loading is used, such as with loadable-components, then the init
-  // function will install some handler that will properly know when to do hyrdation.
```
justin808, line 311: Should we put back in a comment reflecting your research as to why this changed?

node_package/src/clientStartup.ts
```diff
@@ -304,12 +304,7 @@ export function clientStartup(context: Context): void {
...
-  if (document.readyState !== 'loading') {
+  if (document.readyState === 'complete') {
```
justin808: This is the way the code used to be, which I changed in https://github.com/shakacode/react_on_rails/commit/7b301f9e87b4f27b9197207f8e210c38095c551b.

We need to ensure that we don't break the use case of loadable-components.

Maybe we should allow an opt-in to loading  scripts so long as the `readState !== 'loading'`?

CC: @Romex91 @AbanoubGhadban

node_package/src/clientStartup.ts
```diff
@@ -304,12 +304,7 @@ export function clientStartup(context: Context): void {
...
     window.setTimeout(renderInit);
   } else {
     document.addEventListener('DOMContentLoaded', renderInit);
```
Romex91: From MDN:
1. [`The DOMContentLoaded event fires when the initial HTML document has been completely loaded and parsed, without waiting for stylesheets, images, and subframes to finish loading.`](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event)
2. [`readyState === 'complete': The document and all sub-resources have finished loading. The state indicates that the load event is about to fire.`](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState#readystatechange_as_an_alternative_to_load_event)

The event listener and the if statement correspond to different stages of page load.

Should it be something like the code below?
```
function onPageReady(callback) {
  if (document.readyState === "complete") {
    callback();
  } else {
    document.addEventListener("readystatechange", function onReadyStateChange() {
        onPageReady(callback);
        document.removeEventListener("readystatechange", onReadyStateChange);
    });
  }
}
```

Judahmeek: Yes, you're right, @Romex91.

Do you think this change, with your suggestions included, could cause any sort of performance regression for our clients?

Judahmeek:
@Judahmeek this PR is ready to merge.

PR #1544, opened 2023-06-12
remove extra js eval call

wyattades:
### Summary

The current implementation calls `ExecJS::Runtime#eval` twice during every react server-side render, but we can refactor it to only call `eval` once.

This causes a 2x speedup (e.g. 800ms -> 400ms) with the Node.js runtime, and other runtimes with slow IO.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1544)

### DIFF
--- a/lib/react_on_rails/server_rendering_js_code.rb
+++ b/lib/react_on_rails/server_rendering_js_code.rb
@@ -40,13 +40,17 @@ module ReactOnRails
           var railsContext = #{rails_context};
         #{redux_stores}
           var props = #{props_string};
-          return ReactOnRails.serverRenderReactComponent({
-            name: '#{react_component_name}',
-            domNodeId: '#{render_options.dom_id}',
-            props: props,
-            trace: #{render_options.trace},
-            railsContext: railsContext
-          });
+          try {
+            return ReactOnRails.serverRenderReactComponent({
+              name: '#{react_component_name}',
+              domNodeId: '#{render_options.dom_id}',
+              props: props,
+              trace: #{render_options.trace},
+              railsContext: railsContext
+            });
+          } finally {
+            console.history = [];
+          }
         })()
         JS
       end
--- a/lib/react_on_rails/server_rendering_js_code.rb
+++ b/lib/react_on_rails/server_rendering_js_code.rb
@@ -40,17 +40,13 @@ module ReactOnRails
           var railsContext = #{rails_context};
         #{redux_stores}
           var props = #{props_string};
-          try {
-            return ReactOnRails.serverRenderReactComponent({
-              name: '#{react_component_name}',
-              domNodeId: '#{render_options.dom_id}',
-              props: props,
-              trace: #{render_options.trace},
-              railsContext: railsContext
-            });
-          } finally {
-            console.history = [];
-          }
+          return ReactOnRails.serverRenderReactComponent({
+            name: '#{react_component_name}',
+            domNodeId: '#{render_options.dom_id}',
+            props: props,
+            trace: #{render_options.trace},
+            railsContext: railsContext
+          });
         })()
         JS
       end
--- a/node_package/src/serverRenderReactComponent.ts
+++ b/node_package/src/serverRenderReactComponent.ts
@@ -9,7 +9,7 @@ import buildConsoleReplay from './buildConsoleReplay';
 import handleError from './handleError';
 import type { RenderParams, RenderResult, RenderingError } from './types/index';
 
-export default function serverRenderReactComponent(options: RenderParams): null | string | Promise<RenderResult> {
+function serverRenderReactComponentInternal(options: RenderParams): null | string | Promise<RenderResult> {
   const { name, domNodeId, trace, props, railsContext, renderingReturnsPromises, throwJsErrors } = options;
 
   let renderResult: null | string | Promise<string> = null;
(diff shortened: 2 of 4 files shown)


lib/react_on_rails/server_rendering_js_code.rb
```diff
@@ -40,13 +40,17 @@ def render(props_string, rails_context, redux_stores, react_component_name, rend
...
+          try {
+            return ReactOnRails.serverRenderReactComponent({
+              name: '#{react_component_name}',
+              domNodeId: '#{render_options.dom_id}',
+              props: props,
+              trace: #{render_options.trace},
+              railsContext: railsContext
+            });
+          } finally {
+            console.history = [];
```
wyattades: alternatively, we could move this into `serverRenderReactComponent`. I'll let you all decide

justin808:
Awesome work @wyattades. Can you document how you profiled this? Maybe you could add such information to CONTRIBUTING.md?

lib/react_on_rails/server_rendering_js_code.rb
justin808: This is a good solution.

justin808 (changes requested):
Good work.

To merge, I'd like to see:
1. Tests for the changes if easy to do.
2. We need to deprecate or prevent eval_js from being called directly, as it won't clear the console.history any more.
3. We need a changelog.md entry.

wyattades:
> Tests for the changes if easy to do.
> We need a changelog.md entry.

Can do, probably by next week.

> Can you document how you profiled this? Maybe you could add such information to CONTRIBUTING.md?

I don't have anything reproducible at the moment, since I tested it against a private project. But this would be as simple as running a benchmark on request durations on the dummy project. Maybe it'd be good to make a `bin/benchmark` script that tests a few different execjs runtimes and React SSR scenarios?

> We need to deprecate or prevent eval_js

I don't see any documentation that it's public, do we need deprecation?

lib/react_on_rails/server_rendering_js_code.rb
justin808: BTW, @Judahmeek @ahangarha @KhaledEmaraDev This line would be a great place to add a configurable cleanup code for SSR. For example, in the past, I saw weird side effects from MobX on SSR.

justin808: @wyattades I see your point that we could move `console.history = [];` to here: https://github.com/shakacode/react_on_rails/blob/master/node_package/src/serverRenderReactComponent.ts#L12

One can argue that it makes sense to keep the generated JS as small as possible, so I like that solution.

@Judahmeek @alexeyr-ci what do you guys think?

I vote we:

1. rename the [current function `serverRenderReactComponent`](https://github.com/shakacode/react_on_rails/blob/master/node_package/src/serverRenderReactComponent.ts#L12) to something like `serverRenderReactComponentInternal` and not export it.
2. Create `serverRenderReactComponent` in the same file with the try/finally block.

justin808:
@wyattades 
> I don't see any documentation that it's public, do we need deprecation?
Correct! we don't!

lib/react_on_rails/server_rendering_js_code.rb
justin808: Easier to read.  Function is already too long.

Judahmeek:
@wyattades can you also rebase on master?

CI won't pass until you do since we had some of the jobs set to trigger on `push` only, which meant that they didn't trigger for PRs from forks.

This oversight was corrected by #1553

.prettierrc
```diff
@@ -6,7 +6,6 @@ singleQuote: true
 trailingComma: all
 bracketSpacing: true
 jsxBracketSameLine: false
-parser: flow
```
wyattades: maybe out of scope here, but prettier was essentially disabled

node_package/src/serverRenderReactComponent.ts
```diff
@@ -33,37 +32,42 @@ See https://github.com/shakacode/react_on_rails#renderer-functions`);
...
-        // We let the client side handle any redirect
-        // Set hasErrors in case we want to throw a Rails exception
-        hasErrors = !!(reactRenderingResult as {routeError: Error}).routeError;
-
-        if (hasErrors) {
-          console.error(
-            `React Router ERROR: ${JSON.stringify((reactRenderingResult as {routeError: Error}).routeError)}`,
-          );
-        }
+      // We let the client side handle any redirect
```
wyattades: whitespace changes caused by prettier

wyattades:
> Can you document how you profiled this? Maybe you could add such information to CONTRIBUTING.md?

Made a separate PR for this, since I ran into some issues with the dummy app https://github.com/shakacode/react_on_rails/pull/1555

node_package/src/serverRenderReactComponent.ts
justin808: Is this unavoidable?

Can we have a separate pr for prettier?

justin808 (changes requested):
Prefer to put formatting in separate PR.

node_package/src/serverRenderReactComponent.ts
wyattades: reverted

justin808:
Nice and small update!

PR #1545, opened 2023-06-13
Always generate append_X_pack calls if not in development

Judahmeek:
### Summary

Instead of creating an `append_X_pack` call  conditional on the existence of the generated pack entrypoint, the new logic will always create the `append_X_pack` call for the component as long as the component's render option's `auto_load_bundle` is `true` in `test` or `production` environments.

In the development environment, however, a missing generated pack entrypoint will result in a raised error message.

### Pull Request checklist

- [x] Update documentation
- [ ] Update CHANGELOG file

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1545)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Changes since last non-beta release.
 
 *Please add entries here for your pull requests that are not yet released.*
 
+### Removed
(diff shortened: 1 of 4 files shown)


justin808:
@Judahmeek any changelog entry?

PR #1546, opened 2023-06-13
WIP - Upgrade Shakapacker from 6 to 7

ahangarha:
### Summary

Upgrade Shakapacker to version 7

### Pull Request checklist

- [ ] Add/update test to cover these changes
- [ ] Update documentation
- [ ] Update CHANGELOG file

### Other Information

Todos:
- [ ] Change all spelling from `webpacker` to `shakapacker`

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1546)

PR #1547, opened 2023-06-14
Use adapter method for Shakapacker/Webpacker constant

Judahmeek:
### Summary

This change decouples ReactOnRails from the Webpacker constant, enabling ReactOnRails to support both Webpacker v5 & Shakapacker v7 without any breaking changes.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1547)

### DIFF
--- a/.github/workflows/lint-js-and-ruby.yml
+++ b/.github/workflows/lint-js-and-ruby.yml
@@ -1,6 +1,6 @@
 name: Lint JS and Ruby
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
--- a/.github/workflows/package-js-tests.yml
+++ b/.github/workflows/package-js-tests.yml
@@ -1,6 +1,6 @@
 name: JS unit tests for Renderer package
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
--- a/.github/workflows/rspec-package-specs.yml
+++ b/.github/workflows/rspec-package-specs.yml
@@ -1,6 +1,6 @@
 name: Rspec test for gem
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
--- a/lib/generators/USAGE
+++ b/lib/generators/USAGE
@@ -21,4 +21,4 @@ Then you may run
 
 More Details:
 
-    `https://github.com/shakacode/react_on_rails/blob/master/docs/basics/generator-details.md`
+    `https://github.com/shakacode/react_on_rails/blob/master/docs/additional-details/generator-details.md`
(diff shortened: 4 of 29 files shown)


ahangarha:
This is an interesting implementation. It is a learning opportunity for me.

But I wonder if we need to add such complexity to this project while migrating to Shakapacer 7 is easy and straightforward. Adding an adapter would be good if we have a dependency with a complex upgrade, but in this case, encouraging migration is a better approach.

Judahmeek:
> But I wonder if we need to add such complexity to this project while migrating to Shakapacer 7 is easy and straightforward. Adding an adapter would be good if we have a dependency with a complex upgrade, but in this case, encouraging migration is a better approach.

You're probably right, although I don't think these changes actually add that much more complexity to our code base.

What this change does do is enable any ReactOnRails users who are still using Webpacker 5 (which I believe we're still supporting? We never formally dropped support for it) to continue using Webpacker & enjoy the latest bug fixes, such as https://github.com/shakacode/react_on_rails/pull/1540 or https://github.com/shakacode/react_on_rails/pull/1544.

If we don't support Webpacker 5, then I would agree that the transition from Shakapacker v6 to Shakapacker v7 should be smooth enough that these changes are unnecessary.

justin808:
Seems like this can work, but is this worth the extra complexity? Can we really have tests for both shakapacker and webpacker dependencies? 

Could we just change to Shakapacker for the next version? Why or why not?

Judahmeek:
@justin808

> Seems like this can work, but is this worth the extra complexity?

The only benefit over a simple replace & major version bump is that this would enable users still using Webpacker to enjoy any non-Shakapacker specific fixes or functionality that we add in the future.

Since we don't intend to support any other webpack wrapper, this complexity is likely overkill.

> Can we really have tests for both shakapacker and webpacker dependencies?

We could, although that would probably require overhauling the test suite to be similar to react-rails' test suite & possibly even the generators for ReactOnRails.

If we were going to merge this PR, I'd probably go with not changing the test suite & handling any webpacker regressions as clients report them.

> Could we just change to Shakapacker v7 for the next version?

It'll require a major version bump for RoR since we're changing the "optional" dependency, but yes.

Then we'll just tell users to upgrade Shakapacker & ReactOnRails together.

ahangarha:
I am not in a right position to give opinion on this. I prefer to learn from the decisions.

But I think we don't need to do this unless there is a necessity.

Judahmeek:
Closing in preference for #1546

ahangarha:
I think #1548 satisfies our needs for now.

Issue #1558, opened 2023-07-18
Client rendering crashes when configuring `optimization.runtimeChunk` to `multiple`


rubenochiavone:
Configuring webpack to embed the runtime in each chunk and calling `react_component` multiple times in a rails view/partial causes the client render to crash with the following error:

```
Could not find component registered with name XXX. Registered component names include [ YYY ]. Maybe you forgot to register the component?
```

```
VM4859 clientStartup.js:132 Uncaught Error: ReactOnRails encountered an error while rendering component: XXX. See above error message.
    at Object.get (ComponentRegistry.js:40:15)
    at Object.getComponent (ReactOnRails.js:211:44)
    at render (VM4859 clientStartup.js:103:53)
    at forEachReactOnRailsComponentRender (VM4859 clientStartup.js:138:9)
    at reactOnRailsPageLoaded (VM4859 clientStartup.js:164:5)
    at renderInit (VM4859 clientStartup.js:205:9)
    at onPageReady (VM4859 clientStartup.js:234:9)
    at HTMLDocument.onReadyStateChange (VM4859 clientStartup.js:238:13)
```

### Webpack configuration

```js
optimization: {
  runtimeChunk: 'multiple'
},
```

### Rails view

```haml
= react_component("XXX", props: @props)
= yield
= react_component("YYY", props: @props)
```

## Environment

1. Ruby version: 3.1
2. Rails version: 7.0.6
3. Shakapacker version: 6.6.0
4. React on Rails version: 13.3.5

## Expected behavior

Should render both components.

## Actual behavior

Crashes on the client with the aforementioned error/stacktrace.

## Small, reproducible repo

Checkout the branch [rubenochiavone/repro-issue-1558](https://github.com/shakacode/react_on_rails/tree/rubenochiavone/repro-issue-1558) and open the dummy/app in the following URL: http://localhost:3000/render_with_layouts_wrapper. You should see the following errors in the dev-tools console:

```
createReactOutput.js:53 Could not find component registered with name ReduxApp. Registered component names include [ HeaderWrapper, HelloWorld, FooterWrapper ]. Maybe you forgot to register the component?
```

```
render @ createReactOutput.js:53
forEachReactOnRailsComponentRender @ createReactOutput.js:61
reactOnRailsPageLoaded @ handleError.js:15
renderInit @ handleError.js:49
onPageReady @ isRenderFunction.js:25
onReadyStateChange @ isRenderFunction.js:25
[...]

PR #1562, opened 2023-07-20
Throw error when attempting to redefine ReactOnRails

rubenochiavone:
### Summary

Webpack configuration could cause react-on-rails to be imported multiple times in each runtime chunk, leading to the global ReactOnRails variable being overridden and client rendering crashes. This should be avoided by relying on a single runtime chunk.

Read more at https://github.com/shakacode/react_on_rails/issues/1558.

### Pull Request checklist

- [x] Update documentation
- [x] Update CHANGELOG file  

### Screenshots

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1562)


docs/javascript/troubleshooting-when-using-webpacker.md
justin808, line 1: shakapacker and not webpacker

node_package/src/ReactOnRails.ts
```diff
@@ -30,6 +30,16 @@ if (ctx === undefined) {
   throw new Error("The context (usually Window or NodeJS's Global) is undefined.");
 }
 
+if (ctx.ReactOnRails !== undefined) {
+  throw new Error(`
+    The ReactOnRails value exists in the ${ctx} scope, it may not be safe to overwrite it.
+    
+    This could be caused by setting Webpack's optimization.runtimeChunk to "true" or "multiple" - please check your Webpack configuration.
```
justin808: This could be caused by setting Webpack's optimization.runtimeChunk to "true" or "multiple," rather than "single." Check your Webpack configuration.

justin808 (changes requested):

docs/javascript/troubleshooting-when-using-webpacker.md
rubenochiavone: I moved the new troubleshooting docs to `docs/javascript/troubleshooting-when-using-shakapacker.md`, but kept the old one since it is related to webpacker. See c6d7da99abe8780efe612a7f3e23496df24c0651.

node_package/src/ReactOnRails.ts
rubenochiavone: Done in cbde473a203dbc230e4431a1df69c28ddc2bfe26.

PR #1567, opened 2023-07-26
Packs generator subdirectory check

### LINKED ISSUE #1566: Pack generation runs with nil components_subdirectory
## Environment

1. Ruby version: `2.7.4p191`
2. Rails version: `6.0.4.1`
3. Shakapacker/Webpacker version: `shakapacker@7.0.2`
4. React on Rails version: `13.3.5`

#### React-On-Rails Config
- components_subdirectory: `nil`
- auto_load_bundle: `false`

## Expected behavior

The config indicates that for `config.components_subdirectory`
> The default is nil. You can enable the feature by updating it in the next line.

This gives off the impression that the pack generation can be disabled by leaving the default (or setting `components_subdirectory` to `nil`).

When running `./bin/rake assets:precompile`, i'd expect the `react_on_rails:generate_packs` task to not generate any packs. Maybe there is some configuration setting i'm missing, but following the code it doesn't seem to be the case.

## Actual behavior

Running the `./bin/rake assets:precompile` task with `components_subdirectory = nil`, causes every file (that doesn't match `*.server.*`) to have a pack generated for it.

Also it seems that file names with non-alphanumeric characters in it or reserved JS keywords, do not get generated correctly in the generated server bundle file (like `something.client`, `default`, `my-file`, etc). But this is most likely because it generated packs for files that don't match the expected `ComponentName.(client|server).js` syntax.

## Initial Debugging

This seems to happen because the `PacksGenerator#components_search_path` resolves to `{source_path}/**/` and then has a `*` appended, making it `{source_path}/**/*`.

Also appears to be related to #1509 (and release [13.3.0](https://github.com/shakacode/react_on_rails/releases/tag/13.3.0)) which removed the `verify_setup_and_generate_packs` function that checked if `components_subdirectory.present?` and subsequently did not generate the packs if this directory was not set ([ref](https://github.com/shakacode/react_on_rails/pull/1509/files#diff-ce958005f1f8ae7bb40ae163bea1570934d090b1267a0c89004605bb8c088059L25)).

Adding `return unless ReactOnRails.configuration.components_subdirectory.present?` to the top of `generate_packs_if_stale` seems to prevent this issue.

blackjack26:
### Summary

Verifies that the `component_subdirectory` is set before generating packs. This is especially important when running the `assets:precompile` task.

Fixes #1566 

### Pull Request checklist
- [x] ~Update documentation~
- [x] Update CHANGELOG file

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1567)

### DIFF
--- a/spec/dummy/spec/packs_generator_spec.rb
+++ b/spec/dummy/spec/packs_generator_spec.rb
@@ -260,7 +260,7 @@ module ReactOnRails
       end
     end
 
-    context "when components subdirectory is not set" do
+    context "when components subdirectory is not set", :focus do
       before do
           ReactOnRails.configuration.components_subdirectory = nil
           ReactOnRails.configuration.auto_load_bundle = false
--- a/spec/dummy/spec/packs_generator_spec.rb
+++ b/spec/dummy/spec/packs_generator_spec.rb
@@ -19,7 +19,6 @@ module ReactOnRails
 
     let(:old_server_bundle) { ReactOnRails.configuration.server_bundle_js_file }
     let(:old_subdirectory) { ReactOnRails.configuration.components_subdirectory }
-    let(:old_auto_load_bundle) { ReactOnRails.configuration.auto_load_bundle }
 
     before do
       ReactOnRails.configuration.server_bundle_js_file = server_bundle_js_file
@@ -257,28 +256,6 @@ module ReactOnRails
         path = "#{webpacker_source_path}/components/#{component_name}/#{components_subdirectory}/#{name}.jsx"
 
         File.write(path, "// Empty Test Component\n")
-      end
-    end
-
-    context "when components subdirectory is not set" do
-      before do
-        ReactOnRails.configure do |config|
-          config.components_subdirectory = nil
-          config.auto_load_bundle = false
-        end
-      end
-
-      after do
-        ReactOnRails.configure do |config|
-          config.components_subdirectory = old_subdirectory
-          config.auto_load_bundle = old_auto_load_bundle
-        end
-      end
-
-      it "does not generate packs" do
-        expect do
-          described_class.instance.generate_packs_if_stale
-        end.not_to output(GENERATED_PACKS_CONSOLE_OUTPUT_REGEX).to_stdout
       end
     end
 
--- a/spec/dummy/spec/packs_generator_spec.rb
+++ b/spec/dummy/spec/packs_generator_spec.rb
@@ -260,7 +260,7 @@ module ReactOnRails
       end
     end
 
-    context "when components subdirectory is not set & auto_load_bundle is false", :focus do
+    context "when components subdirectory is not set & auto_load_bundle is false" do
       it "does not generate packs" do
         puts "YYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY"
         puts "ReactOnRails.configuration.components_subdirectory: #{ReactOnRails.configuration.components_subdirectory}"
--- a/spec/dummy/spec/packs_generator_spec.rb
+++ b/spec/dummy/spec/packs_generator_spec.rb
@@ -262,7 +262,8 @@ module ReactOnRails
 
     context "when components subdirectory is not set & auto_load_bundle is false" do
       it "does not generate packs" do
-        old_sub, old_auto = old_subdirectory, old_auto_load_bundle
+        old_sub = old_subdirectory
+        old_auto = old_auto_load_bundle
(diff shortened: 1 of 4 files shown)


spec/dummy/spec/packs_generator_spec.rb
```diff
@@ -259,6 +259,18 @@ def create_new_component(name)
       end
     end
 
+    context "when components subdirectory is not set" do
+      before do
+        ReactOnRails.configuration.components_subdirectory = nil
+      end
```
ahangarha: I believe you need to set it back to the original value after the test.

ahangarha:
@blackjack26 Please let me know if you are still working on it.

justin808 (approved):
LGTM!

spec/dummy/spec/packs_generator_spec.rb
```diff
@@ -259,6 +260,28 @@ def create_new_component(name)
       end
     end
 
+    context "when components subdirectory is not set" do
+      before do
+        ReactOnRails.configure do |config|
+          config.components_subdirectory = nil
+          config.auto_load_bundle = false
+        end
+      end
```
ahangarha: I am still trying to figure out this part.
Does it mean if one sets `components_subdirectory` to `nil`, they should set `auto_load_bundle` also to false? If so, we should document it. If not, we should automate it.

Otherwise, we may need to implement tests differently.

I am waiting for updates.

blackjack26: @ahangarha was trying different things, but can't quite figure out why that suite is failing. (Also having a hard time running all the tests locally).

There is already some check that throws an error if `components_subdirectory` is `nil` and `auto_load_bundle` is `true` (in `Configuration#check_autobundling_requirements_if_configured`).

ahangarha: Thanks for your work so far.

So, while you are working on the issue, I also try to fix the problem. Please update the PR if you made any progress.

Judahmeek:
I've got some insight into the CI failures:

The X.2 errors all say `Could not find component registered with name X`, but the names of those components shouldn't be `X`, they should all be `generated/X`.

Since they are not, that means that [load_pack_for_generated_component](https://github.com/shakacode/react_on_rails/blob/16a4c7d16f6072e682ce91398c03c22500bc21d4/lib/react_on_rails/helper.rb#L316) isn't doing what is expected, probably because `render_options.auto_load_bundle` is falsy.

justin808:
Sounds like we definitely need CI fixed!

Judahmeek:
So... this is really weird, but it looks like adding debugging statements forces ReactOnRails configuration to initialize sooner (with the correct values) & this fixed the tests?

ahangarha:
I think we need to fix the issue at this part of the code:
https://github.com/shakacode/react_on_rails/blob/e0d5d0e368c5bf5badd188abed5367a07c72e5f8/lib/react_on_rails/packs_generator.rb#L186-L190

ahangarha:
But also, I have a different interpretation of the issue. I think the reported bug is actually a feature here. `component_subdirectory` is the name of exclusive directories from which we want to get files to process. When it is set to `nil`, we don't have any exclusive directory to look into, so we include all files under `source_path`.

Let's not hurry for merge.

Judahmeek:
@ahangarha although `components_search_path` is where we use `configuration.components_subdirectory`, waiting until we call `components_search_path` to check if `configuration.components_subdirectory` has been set wastes processing time & complicates the code more than immediately checking if `configuration.components_subdirectory` has been set as soon as `PacksGenerator.instance.generate_packs_if_stale` is called.

[Guard clauses](https://betterprogramming.pub/refactoring-guard-clauses-2ceeaa1a9da) are not just band-aids. They are an attempt to simplify code, by eliminating edge cases.

spec/dummy/spec/packs_generator_spec.rb
Judahmeek: @ahangarha if `auto_load_bundle` is true & `components_subdirectory` is not set, then [the Configuration throws an error here](https://github.com/shakacode/react_on_rails/blob/16a4c7d16f6072e682ce91398c03c22500bc21d4/lib/react_on_rails/configuration.rb#L128)

This PR is to stop a bug, in which [the PacksGenerator runs during the `assets:precompile` rake task](https://github.com/shakacode/react_on_rails/blob/16a4c7d16f6072e682ce91398c03c22500bc21d4/lib/react_on_rails/configuration.rb#L160), even though `auto_load_bundle` is `false` & `components_subdirectory` is not set.

PR #1568, opened 2023-07-27
Add ruby 3 and node 16 & 18 in CI

### LINKED ISSUE #1565: Include Ruby 3 and node 18 to CI workflows
Currently, we only run CI with ruby 2.7 and node: 14.

Besides the fact Node 14 is not supported anymore, we need to ensure the gem works with the recent version of Ruby and Node.

MY proposal:
- ruby: `[2.7, 3.2]`
- node: `[16, 18]`

@justin808 @Judahmeek

ahangarha:
### Summary
Update Ruby and Node versions in CI

### Other Information
Closes #1565

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1568)


justin808:
Not passing CI

Judahmeek (approved):

lib/react_on_rails/git_utils.rb
ahangarha, line 1: Though I don't have any proposal for fixing the issue, this solution seems a bit hacky. It fixes our failing tests but it doesn't make sense to me. We can pass `git_installed = false` while we execute `git status ...`.

**Update:**
I think there is another issue as well. We set `git_installed` to `$CHILD_STATUS.success?` before we run `git status --porcelain`. This means we do not capture `suscess` of the `git` command execution.

Maybe we should extract the entire git logic out of this function.

Judahmeek: This is a form of mocking the git status for testing, because without it, we get a `Cannot proxy frozen objects, rspec-mocks relies on proxies for method stubbing and expectations.` error.

PR #1580, opened 2023-11-04
Remove JsonOutput.escape workaround for Rails < 4.1.1

wwahammy:
### Summary

I noticed that JsonOutput.escape had a workaround for Rails < 4.1.1. Since the minimum supported Rails version for the gem is greater than that, it doesn't seem like this needs to be there. 

Addititionally, I removed the `Utils.rails_version_less_than_4_1_1` method since that was the only place it was used

### Pull Request checklist

- [x] ~Update documentation~
- [x] Update CHANGELOG file 

### Other Information

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1580)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ Please follow the recommendations outlined at [keepachangelog.com](http://keepac
 ### [Unreleased]
 Changes since the last non-beta release.
 
-*Please add entries here for your pull requests that are not yet released.*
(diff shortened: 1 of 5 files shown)


justin808:
@wwahammy Nice catch. How did you find this?

wwahammy:
> @wwahammy Nice catch. How did you find this?

I was curious how the JSON output code worked because I needed something like it for a non-React on Rails project. I just noticed it when I got there.

Judahmeek (approved):

Judahmeek:
@wwahammy thank you for your contribution.

PR #1581, opened 2023-11-13
CI Fixes

Judahmeek:
This PR upgrades Puma to v6 to make it compatible with Rack v3.

This PR also adds new workflows so that all workflows required by our repository settings actually run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1581)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,9 @@ on:
   push:
     branches:
       - 'master'
-  pull_request
+  pull_request:
+    paths:
+      - '**.*'
 
 jobs:
   build-dummy-app-webpack-test-bundles:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         ruby: [2.7, 3.2]
-        node: [14, 16, 18]
+        node: [14, 20]
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
@@ -84,7 +84,7 @@ jobs:
     strategy:
       matrix:
         ruby: [2.7, 3.2]
-        node: [14, 16, 18]
+        node: [14, 20]
         rake_task: ['run_rspec:all_but_examples', 'run_rspec:examples']
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -1,13 +1,17 @@
...
+      - 'master'
+  pull_request:
 
 jobs:
   build-dummy-app-webpack-test-bundles:
     strategy:
       matrix:
         ruby: [2.7, 3.2]
-        node: [16, 18]
+        node: [14, 16, 18]
```
ahangarha: Is there any particular reason for bringing back Node 14?
Before we decided to drop support for Node 14. (Though it seems I missed it in some other places)

Judahmeek: If we dropped support for it, then we need to tell Justin to remove any workflows for Node v14 from the list of required workflows.

Screenshot 2023-11-14 at 10 26 22 AM

.github/workflows/main.yml
```diff
@@ -79,8 +83,8 @@ jobs:
     needs: build-dummy-app-webpack-test-bundles
     strategy:
       matrix:
-        ruby: [2.7]
-        node: [14]
+        ruby: [2.7, 3.2]
+        node: [14, 16, 18]
```
justin808: newer node should be added

.github/workflows/package-js-tests.yml
```diff
@@ -1,12 +1,17 @@
...
+    branches:
+      - 'master'
+  pull_request:
 
 jobs:
   build:
     strategy:
       matrix:
-        node: [16, 18]
+        node: [14, 16, 18]
```
justin808: newer node

.github/workflows/rspec-package-specs.yml
```diff
@@ -1,13 +1,17 @@
...
+      - 'master'
+  pull_request:
 
 jobs:
   build:
     strategy:
       matrix:
         ruby: [2.7, 3.2]
-        node: [16, 18]
+        node: [14, 16, 18]
```
justin808: same

justin808 (changes requested):

ahangarha (approved):
I agree with this decision of testing with the oldest (that we care about) and the newest Node version.

justin808 (approved):

PR #1592, opened 2024-01-05
Prevent Shakapacker v7 deprecation messages

ahangarha:
### Summary

This PR prevents displaying the Shakapcker 7+ deprecation message for:
- `webpacker_precompile?`
- `webpacker:clean` rake task

### Pull Request checklist
- [x] ~Update documentation~
- [x] Update CHANGELOG file

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1592)

### DIFF
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -141,7 +141,7 @@ module ReactOnRails
 
       return if skip_react_on_rails_precompile || build_production_command.blank?
 
-      raise(ReactOnRails::Error, shakapacker_precompile_true_message) if shakapacker_precompile?
+      raise(ReactOnRails::Error, compile_command_conflict_message) if shakapacker_precompile?
 
       precompile_tasks = lambda {
         Rake::Task["react_on_rails:generate_packs"].invoke
@@ -152,10 +152,8 @@ module ReactOnRails
         # removes files older than 1 hour.
         versions = 100_000
 
-        clean_task = using_shakapacker_6? ? "webpacker:clean" : "shakapacker:clean"
-
-        puts "Invoking task #{clean_task} from React on Rails"
-        Rake::Task[clean_task].invoke(versions)
+        puts "Invoking task #{shakapacker_clean_task} from React on Rails"
+        Rake::Task[shakapacker_clean_task].invoke(versions)
       }
 
       if Rake::Task.task_defined?("assets:precompile")
@@ -257,20 +255,18 @@ module ReactOnRails
       raise ReactOnRails::Error, msg
     end
 
-    def using_shakapacker_6?
-      shakapacker_major_version = ReactOnRails::WebpackerUtils.shakapacker_version_as_array[0]
-
-      shakapacker_major_version == 6
-    end
-
     def shakapacker_precompile?
-      return Webpacker.config.webpacker_precompile? if using_shakapacker_6?
+      return Webpacker.config.webpacker_precompile? if ReactOnRails::WebpackerUtils.using_shakapacker_6?
 
       Webpacker.config.shakapacker_precompile?
     end
 
-    def shakapacker_precompile_true_message
-      packer = using_shakapacker_6? ? "webpacker" : "shakapacker"
+    def shakapacker_clean_task
+      ReactOnRails::WebpackerUtils.using_shakapacker_6? ? "webpacker:clean" : "shakapacker:clean"
+    end
+
+    def compile_command_conflict_message
+      packer = ReactOnRails::WebpackerUtils.using_shakapacker_6? ? "webpacker" : "shakapacker"
 
       <<~MSG
 
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -141,7 +141,7 @@ module ReactOnRails
 
       return if skip_react_on_rails_precompile || build_production_command.blank?
 
-      if Webpacker.config.webpacker_precompile?
+      if shakapacker_precompile?
         msg = <<~MSG
 
           React on Rails and Shakapacker error in configuration!
@@ -265,6 +265,14 @@ module ReactOnRails
       MSG
 
       raise ReactOnRails::Error, msg
+    end
+
+    def shakapacker_precompile?
+      config = Webpacker.config
+
+      return config.webpacker_precompile? unless config.respond_to?(:shakapacker_precompile?)
+
+      config.shakapacker_precompile?
     end
   end
 end
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -146,13 +146,16 @@ module ReactOnRails
       precompile_tasks = lambda {
         Rake::Task["react_on_rails:generate_packs"].invoke
         Rake::Task["react_on_rails:assets:webpack"].invoke
-        puts "Invoking task webpacker:clean from React on Rails"
 
         # VERSIONS is per the shakacode/shakapacker clean method definition.
         # We set it very big so that it is not used, and then clean just
         # removes files older than 1 hour.
         versions = 100_000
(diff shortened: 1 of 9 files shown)


justin808 (changes requested):
see suggestion

justin808 (commented):
Should there be any test?

lib/react_on_rails/configuration.rb
```diff
@@ -266,6 +266,14 @@ def raise_missing_components_subdirectory
 
       raise ReactOnRails::Error, msg
     end
+
+    def shakapacker_precompile?
+      config = Webpacker.config
```
justin808: In the future, won't Webpacker be unavailable?

ahangarha: Sure.
But still, we need to provide support for the old spelling.

As soon as we force the upgrade to Shakapacker 7+, we need to have support for webpacker spelling.

lib/react_on_rails/configuration.rb
```diff
@@ -266,6 +256,34 @@ def raise_missing_components_subdirectory
...
+      shakapacker_major_version == 6
+    end
+
+    def shakapacker_precompile?
+      return Webpacker.config.webpacker_precompile? if using_shakapacker_6?
+
+      Webpacker.config.shakapacker_precompile?
+    end
+
+    def shakapacker_precompile_true_message
```
Judahmeek: We should rename this to something like "compile_command_conflict_message".

spec/dummy/spec/rake/assets_precompile_rake_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require "rake"
+require "fileutils"
+
+require_relative "../rails_helper"
+
+describe "rake assets:precompile task" do
+  it "doesn't show deprecation message for using webpacker:clean task" do
+    ENV["SHAKAPACKER_PRECOMPILE"] = "false"
```
ahangarha: I couldn't find any other way to disable shakapacker_precompile. I tried mocking `ENV[]` and `ENV.fetch` but didn't have success.

justin808: There's a clear way to mock `ENV[]` in Ruby. please use it.

lib/react_on_rails/configuration.rb
justin808: At some point, we release an update to Shakapacker.

Somebody has not yet updated React on Rails.

`config = Webpacker.config` crashes.

Check that the constant `Webpacker` exists before calling it.

Other code spots might have the same issue. Check.

justin808:
Please fix conflicts and see my comments.

spec/dummy/spec/rake/assets_precompile_rake_spec.rb
ahangarha: I have done that. I mocked both `[]` and `fetch` methods, but it didn't work here. At least the tests failed locally.

I am more concerned about the cause of the current failure.

justin808:
@ahangarha please finish up so I can merge.

lib/generators/react_on_rails/base_generator.rb
```diff
@@ -91,7 +91,8 @@ def add_yarn_dependencies
             babel-plugin-macros"
 
         puts "Adding CSS handlers"
-        run "yarn add css-loader css-minimizer-webpack-plugin mini-css-extract-plugin style-loader"
+        # css-minimizer-webpack-plugin v6.0.0 requires Node 18+
+        run "yarn add css-loader 'css-minimizer-webpack-plugin@<6.0.0' mini-css-extract-plugin style-loader"
```
ahangarha: This is added in a [separate PR](https://github.com/shakacode/react_on_rails/pull/1598) with an updated changelog, with this in mind that it will be merged before this PR. I needed this to be fixed here as well so I added the same code here too.

lib/react_on_rails/configuration.rb
ahangarha: Let's fix this issue in another dedicated PR.

I think at some point, we should drop support Shakapacker 6. I think it is important to push projects to move to use the new API for shakapacker and version 7 was the transitional version for that migration.

justin808 (commented):
See comment in #1598. Otherwise this is OK.

.github/workflows/main.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         ruby: [3.0, 3.3]
-        node: [16, 20]
+        node: [18, 20]
```
ahangarha: As per the discussion [here](https://github.com/shakacode/react_on_rails/pull/1598#discussion_r1461220802), we bump the Node version to 18+ to be able to run our generator.

spec/dummy/spec/rake/assets_precompile_rake_spec.rb
ahangarha: Fixed.

lib/react_on_rails/configuration.rb
ahangarha: Done

.github/workflows/main.yml
ahangarha: This needs to be reverted after the #1598 merge.

lib/react_on_rails/configuration.rb
```diff
@@ -266,6 +254,32 @@ def raise_missing_components_subdirectory
 
       raise ReactOnRails::Error, msg
     end
+
+    def shakapacker_precompile?
+      return Webpacker.config.webpacker_precompile? if ReactOnRails::WebpackerUtils.using_shakapacker_6?
+
+      Webpacker.config.shakapacker_precompile?
```
justin808, line 261: @ahangarha @Judahmeek should this be 

```ruby
Shakapacker.config.shakapacker_precompile?
```

lib/react_on_rails/webpacker_utils.rb
```diff
@@ -141,5 +141,11 @@ def self.raise_shakapacker_not_installed
     def self.semver_to_string(ary)
       "#{ary[0]}.#{ary[1]}.#{ary[2]}"
     end
+
+    def self.using_shakapacker_6?
+      shakapacker_major_version = shakapacker_version_as_array[0]
+
+      shakapacker_major_version == 6
```
justin808, line 148:
```rb
shakapacker_version[:major] == 6
```

isn't that nicer? you can change the shakapacker_version method very easily.

spec/react_on_rails/generators/dev_tests_generator_spec.rb
```diff
@@ -25,16 +25,16 @@
 
     it "changes package.json to use local react-on-rails version of module" do
       assert_file("package.json") do |contents|
-        assert_match('"react-on-rails"', contents)
-        assert_match('"postinstall"', contents)
+        expect(contents).to match('"react-on-rails"')
```
justin808, line 28: Ideally, such changes are in a separate PR.

lib/react_on_rails/webpacker_utils.rb
ahangarha: Well, if it is okay to modify that method, your suggestion is much much better.
Can we be sure that nobody uses that method?

spec/react_on_rails/generators/dev_tests_generator_spec.rb
ahangarha: Sure. But it was needed to pass Rubocop.

lib/react_on_rails/configuration.rb
ahangarha: I think we should do this when we drop support for Webpacker/Shakapacker 6.

This way, we still know we are supporting the old interface. Otherwise, changing the name is not a big issue and can be done in a separate PR.

PR #1596, opened 2024-01-10
Update node dependencies

ahangarha:
We drop support for Node v14 and v17 because of bumping [`@typescript-eslint/parser` version to 6+](https://github.com/typescript-eslint/typescript-eslint/blob/main/CHANGELOG.md#600-2023-07-10) and [js-dom to v22](https://github.com/jsdom/jsdom/blob/main/Changelog.md#2200)

Remaining upgrades:
- Bump React to version 18
- Bump Eslint-related dependencies and fix missing peer dependency issues for `eslint-config-shakacode`

Can we do these upgrades in another PR?

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1596)


justin808:
please rebase and update

ahangarha:
@justin808 
This PR still needs more work for
- updating React to v18
- Eslint

node_package/src/buildConsoleReplay.ts
```diff
@@ -1,6 +1,8 @@
 import RenderUtils from './RenderUtils';
 import scriptSanitizedVal from './scriptSanitizedVal';
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
```
ahangarha, line 5: This part needs to be improved. I couldn't find a way other than ignoring the linter rule.

node_package/tests/ReactOnRails.test.js
```diff
@@ -21,7 +21,7 @@ describe('ReactOnRails', () => {
 
     document.body.innerHTML = '<div id="root"></div>';
     // eslint-disable-next-line no-underscore-dangle
-    const actual = ReactOnRails.render('R1', {}, 'root')._reactInternalFiber.type;
+    const actual = ReactOnRails.render('R1', {}, 'root')._reactInternals.type;
```
ahangarha, line 24: This is internal logic in React, and there is no documentation about it. But in React 17, `ReactOnRails.render('R1', {}, 'root')` only has `._reactInternals` In React 18, this is changed again.

node_package/src/buildConsoleReplay.ts
Judahmeek: Some times there's no better way to do things than to ignore linter rules.

That's because some linter rules are just for identifying possible code smells.

webpackConfigLoader.js
```diff
@@ -51,7 +51,7 @@ const configLoader = (configPath) => {
   // Some test environments might not have the NODE_ENV set, so we'll have fallbacks.
   const configEnv = env.NODE_ENV || env.RAILS_ENV || 'development';
   const ymlConfigPath = join(configPath, 'webpacker.yml');
-  const settings = safeLoad(readFileSync(ymlConfigPath, 'utf8'))[configEnv];
+  const settings = load(readFileSync(ymlConfigPath, 'utf8'))[configEnv];
```
Judahmeek, line 54: @ahangarha can you explain why we had to stop using safeLoad?

ahangarha: Sure. That is because of the [new changes in `js-yaml` v4.0.0](https://github.com/nodeca/js-yaml/blob/master/CHANGELOG.md#400---2021-01-03).

> Breaking: removed `safe*` functions. Use `load`, `loadAll`, `dump` instead which are all now safe by default.

They are all safe by this version.

Judahmeek (approved):

package.json
```diff
@@ -14,45 +14,45 @@
...
   "dependencies": {
     "@babel/runtime-corejs3": "^7.12.5"
   },
   "peerDependencies": {
-    "js-yaml": ">= 3.0.0",
-    "react": ">= 16",
-    "react-dom": ">= 16"
+    "js-yaml": ">= 4.1.0",
+    "react": ">= 17",
+    "react-dom": ">= 17"
```
justin808, line 55: @Judahmeek @ahangarha can we under the bump of React?

webpackConfigLoader.js
justin808: Figure out why we still have this file and if we can delete this.

package.json
ahangarha: Currently, no logic has changed in the codebase regarding React API (except in tests, which is not our concern here).

I think we can keep the minimum version of React to 16.

webpackConfigLoader.js
ahangarha: @justin808 @Judahmeek 

We don't use this file anymore, but I am not sure if no other project uses this as config loader. It is very unlikely but still...

It is included in out npm package and removing it is a breaking change.

What about putting this file in the queue for removal in the next major release and for now, we show a deprecation message for using it?

PR #1598, opened 2024-01-21
Restrict css-minimizer-webpack-plugin version in the generator unless using Node v18+

ahangarha:
### Summary

Recently `css-minimizer-webpack-plugin` [released v6.0.0](https://github.com/webpack-contrib/css-minimizer-webpack-plugin/releases/tag/v6.0.0) in which the minimum required Node version is 18.

We install this package in our generator, and we need this PR to ensure our generator doesn't break in an environment with an older Node.

~⚠️  We may need to make a release after this merge.~

~Alternatively, we can bump our Nodejs support to v18 (the current LTS)~

I have change the generator so that it limits the version of this package unless we use Node v18+.

### Pull Request checklist

- [x] ~Add/update test to cover these changes~
- [x] ~Update documentation~
- [x] ~Update CHANGELOG file~  

### Other Information
-

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1598)

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -91,8 +91,7 @@ module ReactOnRails
             babel-plugin-macros"
 
         puts "Adding CSS handlers"
-        # css-minimizer-webpack-plugin v6.0.0 requires Node 18+
-        run "yarn add css-loader 'css-minimizer-webpack-plugin@<6.0.0' mini-css-extract-plugin style-loader"
+        run "yarn add css-loader css-minimizer-webpack-plugin mini-css-extract-plugin style-loader"
 
         puts "Adding dev dependencies"
         run "yarn add -D @pmmmwh/react-refresh-webpack-plugin react-refresh"
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -91,7 +91,15 @@ module ReactOnRails
             babel-plugin-macros"
 
         puts "Adding CSS handlers"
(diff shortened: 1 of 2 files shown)


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -91,7 +91,8 @@ def add_yarn_dependencies
             babel-plugin-macros"
 
         puts "Adding CSS handlers"
-        run "yarn add css-loader css-minimizer-webpack-plugin mini-css-extract-plugin style-loader"
+        # css-minimizer-webpack-plugin v6.0.0 requires Node 18+
```
justin808: This is 100% the wrong thing to do. We should not be doing things like this to fix testing...Instead, just take Node 16 out of the test matrix. 

We're not requiring an update to Node.  Node 16 should still work with React on Rails.

Please confirm that the error is easy to debug if somebody uses the generator with old Node.

justin808 (changes requested):

ahangarha:
I close this PR regarding [this comment](https://github.com/shakacode/react_on_rails/pull/1598#discussion_r1461220802). The required change is applied in #1592

lib/generators/react_on_rails/base_generator.rb
Judahmeek: @ahangarha you could actually make the version of `css-minimizer-webpack-plugin` installed conditional on the version of `node` that is being used.

Have the generator check the result of `node -v` & then install the latest version of `css-minimizer-webpack-plugin` compatible with that `node` version.

ahangarha: Sounds better 👍🏾

Judahmeek (approved):

justin808 (approved):

PR #1606, opened 2024-04-05
Fix rubocop to v1.61

ahangarha:
Over time, Rubycup gets a new version and that causes CI failure for some new linting rules.

The easiest approach is to fix the version of Rubocup.

I couldn't find any reason why without fixing the version in other repos, we don't face such issues there. So my proposal here is to fix and freeze the version.

⚠️ This PR includes a minimal and temporary fix to prevent the installation of `css-loader` v7, which has [breaking changes](https://github.com/webpack-contrib/css-loader/releases/tag/v7.0.0).

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1606)

### DIFF
--- a/lib/react_on_rails/git_utils.rb
+++ b/lib/react_on_rails/git_utils.rb
@@ -4,7 +4,7 @@ require "English"
 
 module ReactOnRails
   module GitUtils
-    def self.uncommitted_changes?(message_handler, git_installed = true)
+    def self.uncommitted_changes?(message_handler, git_installed: true)
       return false if ENV["COVERAGE"] == "true"
 
       status = `git status --porcelain`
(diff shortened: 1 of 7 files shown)


Judahmeek (approved):

lib/react_on_rails/git_utils.rb
```diff
@@ -4,7 +4,7 @@
 
 module ReactOnRails
   module GitUtils
-    def self.uncommitted_changes?(message_handler, git_installed = true)
```
justin808, line 7: Do only tests use this? if so, maybe we shouldn't have the option?

I'm guessing all other usages assume git is installed, and if so, then we don't need the option.

In other words, does React on Rails break if git is not installed?

PR #1611, opened 2024-04-12
Follow-up to #1606

Judahmeek:
This PR adds lock files for the changes made in #1606 & also restricts the version of `style-loader` in rspec examples (`style-loader` v4 requires a higher version of node than we are currently using) so that CI will pass.

### DIFF
--- a/lib/generators/react_on_rails/base_generator.rb
+++ b/lib/generators/react_on_rails/base_generator.rb
@@ -99,7 +99,7 @@ module ReactOnRails
         css_minimizer_webpack_plugin_version = node_major_version >= 18 ? "" : "@<6.0.0"
         css_minimizer_webpack_plugin = "css-minimizer-webpack-plugin#{css_minimizer_webpack_plugin_version}"
 
-        run "yarn add 'css-loader@<7.0.0' '#{css_minimizer_webpack_plugin}' mini-css-extract-plugin 'style-loader<4.0.0'"
+        run "yarn add 'css-loader@<7.0.0' '#{css_minimizer_webpack_plugin}' mini-css-extract-plugin 'style-loader@<4.0.0'" # rubocop:disable Layout/LineLength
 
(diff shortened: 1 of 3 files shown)


lib/generators/react_on_rails/base_generator.rb
```diff
@@ -99,7 +99,7 @@ def add_yarn_dependencies
         css_minimizer_webpack_plugin_version = node_major_version >= 18 ? "" : "@<6.0.0"
         css_minimizer_webpack_plugin = "css-minimizer-webpack-plugin#{css_minimizer_webpack_plugin_version}"
 
-        run "yarn add 'css-loader@<7.0.0' '#{css_minimizer_webpack_plugin}' mini-css-extract-plugin style-loader"
+        run "yarn add 'css-loader@<7.0.0' '#{css_minimizer_webpack_plugin}' mini-css-extract-plugin 'style-loader@<4.0.0'" # rubocop:disable Layout/LineLength
```
justin808: I don't see why we're locking this:

https://github.com/webpack-contrib/style-loader/releases

New projects can update or change the dependency if needed.

justin808 (changes requested):

justin808:
@Judahmeek let's just update the node version for the tests. If new **generated** projects need a newer node version or older style-loader, that's fine.

react_on_rails.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency "addressable"
   s.add_dependency "connection_pool"
   s.add_dependency "execjs", "~> 2.5"
+  s.add_dependency "sqlite3", "~> 1.6"
```
justin808: why make this a dependency? it is needed for tests only?

Judahmeek: Correct. My mistake.

Issue #1635, opened 2024-06-24
Utilize shakacode/package_json for Enhanced Package Management Compatibility


vaukalak:
**Is your feature request related to a problem? Please describe.**
react-on-rails could benefit from syncronized package management with shakapacker.

**Describe the solution you'd like**
I suggest integrating shakacode/package_json to manage react-on-rails package configurations. This would unify dependency management and enhance project consistency.

justin808:
@G-Rath  can you please make a PR from this? I’d like to ship ASAP.

> turns out I still have my old branch: https://github.com/G-Rath/react_on_rails/tree/use-package_json

> I’ve rebased that with the latest stuff on main and I think it’s almost good to go - the main thing missing is if any tests need to be written

PR #1639, opened 2024-06-25
Use package_json gem

vaukalak:
### Summary

Since shakapacker V8+ uses package_json gem, this is now supported in react_on_rails as well. This PR closes [issue #1635.](https://github.com/shakacode/react_on_rails/issues/1635).

### Pull Request checklist

- [x] Update documentation
- [x] Update CHANGELOG file  

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1639)

## Summary by CodeRabbit

- **Improvements**
  - Enhanced dependency management for easier and more reliable package handling.
  - Updated installation documentation for a streamlined setup process.
  - Improved support for modern versions of Shakapacker.

- **Documentation**
  - Simplified installation instructions in the Getting Started guide.

- **Development**
  - Added the `package_json` gem to development dependencies for better package management.


Gemfile.lock
```diff
@@ -119,6 +119,9 @@ GEM
       tins (~> 1.6)
     crass (1.0.6)
     date (3.3.4)
+    debug (1.9.2)
```
vaukalak: I think those deps are introduced by package_json

lib/generators/react_on_rails/generator_helper.rb
```diff
@@ -1,8 +1,13 @@
 # frozen_string_literal: true
 
+require "package_json"
 require "rainbow"
 
 module GeneratorHelper
+  def package_json
+    @package_json ||= PackageJson.read(Rails.root.to_s)
```
vaukalak: @G-Rath I've changed it vs how it was implemented [here](https://github.com/G-Rath/react_on_rails/commit/6c0a29ebff3b584b6446f1065dec23b6bf7ffb50#diff-84d1b667e35dc8d3cf92e40813c3b880c84a1d31dbd6d6367e8851f84418b435R6) cause it was creating a separate bundle context, due to which I couldn't locate the shakapacker gem during the install generator work.

Gemfile.lock
G-Rath: nope, `package_json` has no dependencies outside of core Ruby, and it's not being added as an explicitly dependency anyway (which I think is fine so long as it's only used in the generator - that's why Shakapacker _does_ directly depend on it, but react-rails doesn't)

G-Rath (approved):

lib/generators/react_on_rails/react_with_redux_generator.rb
```diff
@@ -5,6 +5,7 @@
 module ReactOnRails
   module Generators
     class ReactWithReduxGenerator < Rails::Generators::Base
+      include GeneratorHelper
```
justin808: maybe not necessary?

justin808 (changes requested):

PR #1645, opened 2024-07-26
Replace prepare & prerelease with prepublishOnly

Judahmeek:
Prerelease must have been deprecated a long time ago because it is no longer an NPM lifecycle script.

[prepare is a lifecycle script, but triggers more often than is likely desirable](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts).

prepublishOnly only triggers before `npm publish`

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1645)

## Summary by CodeRabbit

- **Chores**
	- Streamlined the release process by changing the build timing to occur only before publishing.
	- Removed the previous sequence of preparatory checks and builds to enhance efficiency.


justin808:
Great job @Judahmeek

PR #1647, opened 2024-08-25
Add support for console replay while streaming component

AbanoubGhadban:
### Summary
This PR introduces the ability to replay console logs that happen in the react components streamed.

It makes every chunk returned from the server has the following format:

```json
{
  "html": "<div>chunk html</div>",
  "consoleReplayScript": "<script>console.log.apply(console, ['hello world']);</script>"
}
```

On the ruby side, the console replay script is returned in the HTML chunk to be replayed on the browser.

### Key Concepts in SSR and Streaming

1. **Server-Side Rendering (SSR):** 
   SSR generates the HTML for React components on the server and sends it to the client. This allows faster initial rendering on the client side because the HTML is already rendered when the page loads.

2. **Streaming:**
   Instead of sending the entire HTML content at once after it has been fully rendered on the server, streaming allows the server to send partial chunks of HTML as they are rendered. This can improve the perceived load time for the client, as the client can begin processing the initial HTML while the server continues rendering the rest.

### Streaming in the Code

The parts related to streaming are primarily in the `streamServerRenderedReactComponent` function and involve streaming the React component's HTML output back to the client.

#### `streamServerRenderedReactComponent`

```ts
export const streamServerRenderedReactComponent = (options: RenderParams): Readable => {
```

- This function handles the server-side streaming of a React component's rendered HTML.
- It takes `RenderParams` as input, which contains the details of the component to be rendered (name, props, DOM node ID, etc.).

#### `ReactDOMServer.renderToPipeableStream`

```ts
ReactDOMServer.renderToPipeableStream(reactRenderingResult)
  .pipe(transformStream);
```

- `ReactDOMServer.renderToPipeableStream` is a new React feature introduced in React 18 for streaming SSR.
- It creates a stream of HTML chunks from the React component as it renders. 
- The `pipe` method connects this stream to another stream (`transformStream`), allowing the rendered chunks to be processed and eventually sent to the client.

#### `Transform Stream`

```ts
const transformStream = new Transform({
  transform(chunk, _, callback) {
    const htmlChunk = chunk.toString();
    const consoleReplayScript = buildConsoleReplay(previouslyReplayedConsoleMessages);
    previouslyReplayedConsoleMessages = console.history?.length || 0;

    const jsonChunk = JSON.stringify({
      html: htmlChunk,
      consoleReplayScript,
    });

    this.push(jsonChunk);
    callback();
  }
});
```

- The `Transform` stream is used to process each chunk of HTML rendered by React.
- It takes the incoming chunk (HTML), converts it to a string, and appends a script to replay any console messages (`consoleReplayScript`).
- After processing the chunk, it pushes the modified chunk (now a JSON string containing the HTML and console messages) to the next stream destination, which could be sent to the client.

#### Error Handling in Streams

```ts
renderResult = stringToStream(handleError({
  e,
  name,
  serverSide: true,
}));
```

- If an error occurs during rendering, the `handleError` function generates an error message in the form of a string.
- This string is then converted into a stream using `stringToStream`, ensuring that even errors are streamed back to the client.

#### `stringToStream`

```ts
const stringToStream = (str: string): Readable => {
  const stream = new PassThrough();
  stream.push(str);
  stream.push(null); // Ends the stream
  return stream;
};
```

- `stringToStream` creates a readable stream from a string. 
- This is useful for sending static error messages or other fixed content back to the client as a stream, ensuring that the streaming approach is maintained even in cases of failure.

### Summary of Streaming Flow

1. The function `streamServerRenderedReactComponent` gets the component to be rendered and begins the rendering process using `ReactDOMServer.renderToPipeableStream`.
2. As React renders the component, it sends chunks of HTML through a stream.
3. Each chunk is processed by a `Transform` stream, which appends console logs and serializes the result as JSON.
4. The transformed HTML and any other required data are streamed back to the client as soon as they are available.
5. In case of an error during rendering, the error is handled and streamed back using a `stringToStream`.

This approach allows for efficient server-side rendering with streaming, which can improve performance and user experience by delivering parts of the content to the client as soon as they are ready, without waiting for the entire component to render.

This Ruby file is part of the `ReactOnRails` gem, which integrates React with Rails and enables server-side rendering (SSR) of React components. The streaming parts of this Ruby file relate closely to the streaming functionality in the TypeScript code you provided. Here's a breakdown of how the Ruby code handles SSR with a focus on streaming:

### Key Components of the Ruby File

- **`react_component` Method:**
  This method renders a React component on the server and returns the HTML to be injected into the Rails view. It supports both client-side rendering (CSR) and server-side rendering (SSR).

- **`stream_react_component` Method:**
  This method handles server-side rendering with streaming, which corresponds to the `streamServerRenderedReactComponent` function in the TypeScript code. Instead of waiting for the entire component to be rendered before sending it to the client, it sends chunks of HTML as they are rendered.

#### `stream_react_component`

```ruby
def stream_react_component(component_name, options = {})
  rendering_fiber = Fiber.new do
    stream = internal_stream_react_component(component_name, options)
    stream.each_chunk do |chunk|
      Fiber.yield chunk
    end
    Fiber.yield nil
  end

  if @rorp_rendering_fibers.nil?
    raise ReactOnRails::Error, "You must call stream_view_containing_react_components to render the view containing the react component"
  end

  @rorp_rendering_fibers << rendering_fiber

  # return the first chunk of the fiber
  rendering_fiber.resume
end
```

- **Fiber Usage:**
  The `stream_react_component` method creates a `Fiber` to control the streaming of HTML chunks. Fibers in Ruby are lightweight, cooperative threads that allow the code to yield and resume execution, which is perfect for handling streaming data in a controlled way.
  
- **Streaming Chunks:**
  The Fiber calls `internal_stream_react_component`, which returns a stream of chunks of HTML. The Fiber iterates through each chunk and sends it back to the Rails view as it becomes available.

  - The first chunk is rendered immediately and returned to the view (corresponding to the initial part of the React component).
  - Subsequent chunks are sent asynchronously as they are processed and rendered, mimicking the behavior in the TypeScript code where `renderToPipeableStream` is used.

- **Fiber Management:**
  The rendered chunks are managed by storing the rendering Fiber in an instance variable (`@rorp_rendering_fibers`). The first chunk is sent by calling `rendering_fiber.resume`.

#### `internal_stream_react_component`

```ruby
def internal_stream_react_component(component_name, options = {})
  options = options.merge(stream?: true)
  result = internal_react_component(component_name, options)
  build_react_component_result_for_server_streamed_content(
    rendered_html_stream: result[:result],
    component_specification_tag: result[:tag],
    render_options: result[:render_options]
  )
end
```

- **Options Modification:**
  This method ensures that the `stream?` option is set to `true`, signaling that the component should be streamed rather than fully rendered before being sent to the client.
  
- **Streamed Content:**
  The result of `internal_react_component` is passed to `build_react_component_result_for_server_streamed_content`, which prepares the HTML chunks to be streamed back to the client.

#### `build_react_component_result_for_server_streamed_content`

```ruby
def build_react_component_result_for_server_streamed_content(
  rendered_html_stream: required("rendered_html_stream"),
  component_specification_tag: required("component_specification_tag"),
  render_options: required("render_options")
)
  is_first_chunk = true
  rendered_html_stream.transform do |chunk_json_result|
    if is_first_chunk
      is_first_chunk = false
      next build_react_component_result_for_server_rendered_string(
        server_rendered_html: chunk_json_result["html"],
        component_specification_tag: component_specification_tag,
        console_script: chunk_json_result["consoleReplayScript"],
        render_options: render_options
      )
    end

    result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
    compose_react_component_html_with_spec_and_console(
      "", chunk_json_result["html"], result_console_script
    )
  end
end
```

- **Transforming Streamed Chunks:**
  This method takes the stream of chunks from the rendered React component and processes them one by one. The first chunk is special because it contains the initial HTML along with the component specification tag (metadata about the component).

- **Subsequent Chunks:**
  For subsequent chunks, the HTML is processed and appended to the response, similar to how the `Transform` stream in the TypeScript code appends the console logs and wraps the HTML in JSON.

- **Console Replay:**
  Each chunk includes the option to replay console messages (`consoleReplayScript`), as in the TypeScript code, where each HTML chunk includes embedded console logs.

### Relation to TypeScript Streaming Logic

- **Rendering Flow:**
  Both the Ruby and TypeScript code follow a similar flow for streaming:
  - The component is rendered in chunks, with the first chunk being sent immediately, followed by subsequent chunks.
  - Errors during rendering are handled gracefully, with error messages being sent as part of the stream.

- **Chunk Processing:**
  The Ruby method `build_react_component_result_for_server_streamed_content` closely mirrors the functionality of the `Transform` stream in the TypeScript code. Both process each chunk of HTML, append necessary metadata (like console logs), and send it to the client.

- **Fiber and Stream Handling:**
  The use of Fibers in Ruby allows the server to stream the chunks in a non-blocking manner, similar to how the Node.js `renderToPipeableStream` function allows React to stream chunks without waiting for the entire component to finish rendering.

### Summary

The Ruby file is responsible for managing the server-side rendering of React components within a Rails app, including support for streaming the rendered HTML. It parallels the TypeScript streaming logic by using Fibers to control the flow of streamed HTML chunks, ensuring that the client receives content as it becomes available. Both the Ruby and TypeScript implementations focus on improving performance by enabling partial responses to be sent during rendering, reducing the time to the first byte for the client.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1647)

## Summary by CodeRabbit

- **New Features**
	- Introduced streaming capabilities for React components, enhancing server-side rendering performance and debugging.
	- Added methods for replaying console logs during server rendering, allowing better tracking of asynchronous operations.
	- Enhanced `ReactOnRails` module with a new helper for processing streamed React components.

- **Bug Fixes**
	- Improved error handling in server rendering logic for more robust operations.

- **Documentation**
	- Updated the changelog to reflect new features, changes, and fixes.

- **Dependency Updates**
	- Upgraded core React and React DOM packages to version 18.2.0 for improved features and compatibility.

- **Chores**
	- Added new entries to the `.gitignore` file for cleaner version control, targeting IDE-related files.

### DIFF
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -435,7 +435,7 @@ module ReactOnRails
       render_options: required("render_options")
     )
       is_first_chunk = true
-      rendered_html_stream = rendered_html_stream.transform do |chunk_json_result|
+      rendered_html_stream.transform do |chunk_json_result|
         if is_first_chunk
           is_first_chunk = false
           next build_react_component_result_for_server_rendered_string(
@@ -447,7 +447,8 @@ module ReactOnRails
         end
 
         result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
-        # No need to prepend component_specification_tag or add rails context again as they're already included in the first chunk
+        # No need to prepend component_specification_tag or add rails context again
+        # as they're already included in the first chunk
         compose_react_component_html_with_spec_and_console(
           "", chunk_json_result["html"], result_console_script
         )
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -430,29 +430,28 @@ module ReactOnRails
     end
 
     def build_react_component_result_for_server_streamed_content(
-      rendered_html_stream: required("rendered_html_stream"),
-      component_specification_tag: required("component_specification_tag"),
-      render_options: required("render_options")
+      rendered_html_stream:,
+      component_specification_tag:,
+      render_options:
     )
       is_first_chunk = true
       rendered_html_stream.transform do |chunk_json_result|
         if is_first_chunk
           is_first_chunk = false
-          next build_react_component_result_for_server_rendered_string(
+          build_react_component_result_for_server_rendered_string(
             server_rendered_html: chunk_json_result["html"],
             component_specification_tag: component_specification_tag,
             console_script: chunk_json_result["consoleReplayScript"],
             render_options: render_options
           )
+        else
+          result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
+          # No need to prepend component_specification_tag or add rails context again
+          # as they're already included in the first chunk
+          compose_react_component_html_with_spec_and_console(
+            "", chunk_json_result["html"], result_console_script
+          )
         end
-
-        result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
-        # No need to prepend component_specification_tag or add rails context again
-        # as they're already included in the first chunk
-        compose_react_component_html_with_spec_and_console(
-          "", chunk_json_result["html"], result_console_script
-        )
-      end
     end
 
     def build_react_component_result_for_server_rendered_hash(
@@ -687,3 +686,4 @@ ReactOnRails.reactOnRailsComponentLoaded('#{render_options.dom_id}');
 end
 # rubocop:enable Metrics/ModuleLength
 # rubocop:enable Metrics/MethodLength
+
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -452,6 +452,7 @@ module ReactOnRails
             "", chunk_json_result["html"], result_console_script
           )
         end
+      end
     end
 
     def build_react_component_result_for_server_rendered_hash(
@@ -686,4 +687,3 @@ ReactOnRails.reactOnRailsComponentLoaded('#{render_options.dom_id}');
 end
 # rubocop:enable Metrics/ModuleLength
 # rubocop:enable Metrics/MethodLength
-
--- a/lib/react_on_rails/helper.rb
+++ b/lib/react_on_rails/helper.rb
@@ -434,26 +434,24 @@ module ReactOnRails
       component_specification_tag: required("component_specification_tag"),
       render_options: required("render_options")
     )
-      content_tag_options_html_tag = render_options.html_options[:tag] || "div"
-      # The component_specification_tag is appended to the first chunk
-      # We need to pass it early with the first chunk because it's needed in hydration
-      # We need to make sure that client can hydrate the app early even before all components are streamed
       is_first_chunk = true
-      rendered_html_stream = rendered_html_stream.transform do |chunk|
+      rendered_html_stream = rendered_html_stream.transform do |chunk_json_result|
         if is_first_chunk
           is_first_chunk = false
-          html_content = <<-HTML
-            #{rails_context_if_not_already_rendered}
-            #{component_specification_tag}
-            <#{content_tag_options_html_tag} id="#{render_options.dom_id}">#{chunk}</#{content_tag_options_html_tag}>
-          HTML
-          next html_content.strip
+          next build_react_component_result_for_server_rendered_string(
+            server_rendered_html: chunk_json_result["html"],
+            component_specification_tag: component_specification_tag,
+            console_script: chunk_json_result["consoleReplayScript"],
+            render_options: render_options
+          )
         end
-        chunk
-      end
 
-      rendered_html_stream.transform(&:html_safe)
-      # TODO: handle console logs
+        result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
+        # No need to prepend component_specification_tag or add rails context again as they're already included in the first chunk
+        compose_react_component_html_with_spec_and_console(
+          "", chunk_json_result["html"], result_console_script
+        )
+      end
     end
 
     def build_react_component_result_for_server_rendered_hash(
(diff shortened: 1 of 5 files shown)


lib/react_on_rails/helper.rb
```diff
@@ -398,26 +398,25 @@ def build_react_component_result_for_server_streamed_content(
       component_specification_tag: required("component_specification_tag"),
```
justin808: Can you check if all of these "required" calls can go away given that we use a more recent version of Ruby?

lib/react_on_rails/helper.rb
```diff
@@ -398,26 +398,25 @@ def build_react_component_result_for_server_streamed_content(
...
 
-      rendered_html_stream.transform(&:html_safe)
-      # TODO: handle console logs
+        result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
+        # No need to prepend component_specification_tag or add rails context again
+        # as they're already included in the first chunk
+        compose_react_component_html_with_spec_and_console(
+          "", chunk_json_result["html"], result_console_script
+        )
+      end
```
justin808, line 455: Why not a simple "else":

```rb
def build_react_component_result_for_server_streamed_content(
  rendered_html_stream: required("rendered_html_stream"),
  component_specification_tag: required("component_specification_tag"),
  render_options: required("render_options")
)
  is_first_chunk = true
  rendered_html_stream.transform do |chunk_json_result|
    if is_first_chunk
      is_first_chunk = false
      build_react_component_result_for_server_rendered_string(
        server_rendered_html: chunk_json_result["html"],
        component_specification_tag: component_specification_tag,
        console_script: chunk_json_result["consoleReplayScript"],
        render_options: render_options
      )
    else
      result_console_script = render_options.replay_console ? chunk_json_result["consoleReplayScript"] : ""
      compose_react_component_html_with_spec_and_console(
        "", chunk_json_result["html"], result_console_script
      )
    end
  end
end
```

justin808 (changes requested):
1. One suggestion for readability
2. I updated the description with the information from ChatGPT on how this works. Please double-check.
3. Should we have any unit tests?

lib/react_on_rails/helper.rb
AbanoubGhadban: Yes, it's not needed. Removed

PR #1650, opened 2024-10-01
Add support for passing `children` in from Rails

rangerscience:
### Summary

Allows you to do:
(rails template)
```
<%= react_component "MyComponent" do %>
  <%= render "events/form", event: Event.new %>
<% end %>
```

(react component)
```
const MyComponent= () => {
    return (
    <div>
      { props.children }
    </div>
  )
}
```

Note that this uses React's `dangerouslySetInnerHtml`, which... maybe isn't a good idea, but might be fine, since the HTML being set is coming in from the Rails template engine.

You _could_ just include the changes to the Ruby code so that the block gets rendered, and that gets passed in a a nice prop (like `children_html`) but then the React components would have to care about whether their `children` are coming in from more React code, or from the Rails side of things.

### Pull Request checklist

- [x] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

As per the contribution guidelines, I'm supposed to run `rake examples:gen_all`, but that task doesn't exist

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1650)

## Summary by CodeRabbit

- **New Features**
	- Enhanced `react_component` and `react_component_hash` methods to accept child HTML elements via block parameters.
	- Updated documentation for `ReactOnRails` to clarify new functionality and usage of render-functions.

- **Bug Fixes**
	- Improved handling of child elements in React components, ensuring proper rendering and functionality.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -121,6 +121,24 @@ issue.
   };
   ```
 
+- You can pass a block to `react_component`, and it will be provided as the `children` prop to the React component, as a React element:
+  ```ruby
+    <%= react_component("YourComponent") do %>
+      <p>Contained HTML<p>
+      <%= render "your/partial" %>
+    <% end %>
+  ```
+  ```js
+  import React from 'react';
+
+  export default (props) => {
+    return () => (
+      <div>{ props.children }</div>
+    );
+  };
+  ```
+  **Note that this is implementing using React's [dangerouslySetInnerHtml](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html)**, so named because of the exposure to cross-site scripting attacks. You should generally be fine since the HTML is coming from your own code via the Rails rendering engine, but this is sharp knife so please exercise caution :)
+
 See the [View Helpers API](https://www.shakacode.com/react-on-rails/docs/api/view-helpers-api/) for more details on `react_component` and its sibling function `react_component_hash`.
 
 ## Globally Exposing Your React Components
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -131,13 +131,13 @@ issue.
   ```js
   import React from 'react';
 
-  export default (props) => {
-    return () => (
-      <div>{ props.children }</div>
-    );
-  };
+  const YourComponent = (props) => (
+    <div>{ props.children }</div>
+  );
+  
+  export default YourComponent;
   ```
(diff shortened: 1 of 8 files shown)


justin808:
@rangerscience Why not just pass the HTML as a prop and let the react component handle the HTML?

rangerscience:
> @rangerscience Why not just pass the HTML as a prop and let the react component handle the HTML?

Sure! So, then, if you want to have a React component that can take children from _both_ Rails _and_ other React components, you'd need to do:
```
const MyComponent= () => {
    return (
    <div>
      { props.children }
      { props.children_html && <div dangerouslySetInnerHtml={{__html: props.children_html}} /> }
    </div>
  )
}
```

By having React on Rails turn the passed HTML into a React element, the React components don't have to "know" whether they're being invoked by the Rails layer or by the React layer - they get to act just like regular React components.

justin808:
Given an easy work around, why expand the API?

rangerscience:
> Given an easy work around, why expand the API?

Sure! IMHO, developer happiness, least-surprise, and (laziness)[https://thethreevirtues.com/]. You *can* workaround both changes:

```
<%= react_component "MyComponent", props: { children_html: render_to_string("events/form") } %>
```
```
const MyComponent = () => <div>{ props.children_html && <div dangerouslySetInnerHtml={{__html: props.children_html}} /> } />
```
but this makes the developer's intention less clear (all we're doing is passing child elements, nothing more), requires boilerplate, and breaks expectations (causing surprise) - Rails templates can take blocks (presuming the template has a `yield`); React components can take `children` (presuming they put `{ children }` somewhere).

The API expansion allows both the Rails and React side of things to work as intuitively expected; in line with the rest of their ecosystems.

Those are some arguments for why _this_ API expansion is worthwhile. If the debate is actually about whether or not APIs should be expanded when workarounds are available in general, that's a different discussion that I'd be happy to have :)

justin808:
Closing for now as this is easily done via the current API. @rangerscience could you provide a doc PR instead?

rangerscience:
No worries, I'll just keep my fork up and running :) I've got some other changes I want to look at, so if I write up docs it'll be there (and after that), and at that point I'll welcome and help out if y'all want to open a PR from me to you.

I would recommend adding a bit to the docs pointing to this PR, if you want a quick and easy way to get someone this info - or, otherwise make sure the GH PR search pulls this up with search terms you'd expect someone to use.

justin808:
I'm not sure if pointing the docs to the implementation of an added API that was not added will help. 

Thanks in advance for docs on how to cleanly accomplish the goal with the existing API.

PR #1656, opened 2024-11-19
[WIP] Hydrate components immediately after downloading chunks

AbanoubGhadban:
### Summary

_Remove this paragraph and provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together._

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1656)


lib/react_on_rails/configuration.rb
```diff
@@ -41,7 +42,9 @@ def self.configuration
       make_generated_server_bundle_the_entrypoint: false,
       defer_generated_component_packs: true,
       # forces the loading of React components
-      force_load: false
+      force_load: false,
+      auto_load_server_components: true,
+      rsc_rendering_url: DEFAULT_RSC_RENDERING_URL
```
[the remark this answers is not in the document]
AbanoubGhadban: I don't think I will need to add more configs like this soon. But it's useful if you want to implement a helper function.

node_package/src/ComponentRegistry.ts
```diff
@@ -1,9 +1,33 @@
-import type { RegisteredComponent, ReactComponentOrRenderFunction, RenderFunction } from './types/index';
+import React from 'react';
+import type { RegisteredComponent, ReactComponentOrRenderFunction, RenderFunction, ReactComponent } from './types/index';
 import isRenderFunction from './isRenderFunction';
 
 const registeredComponents = new Map<string, RegisteredComponent>();
+const registrationCallbacks = new Map<string, Array<(component: RegisteredComponent) => void>>();
```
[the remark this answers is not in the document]
AbanoubGhadban: Defined in case of being needed later

node_package/src/RSCClientRoot.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import RSDWClient from 'react-server-dom-webpack/client';
+
+if (!('use' in React)) {
+  throw new Error('React.use is not defined. Please ensure you are using React 18.3.0-canary-670811593-20240322 or later to use server components.');
+}
+
+// It's not the exact type, but it's close enough for now
+type Use = <T>(promise: Promise<T>) => T;
```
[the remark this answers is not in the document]
AbanoubGhadban: Added
```
<reference types="react/experimental" />
```
to `types.ts`

node_package/src/StoreRegistry.ts
```diff
@@ -21,6 +23,12 @@ export default {
       }
 
       registeredStoreGenerators.set(name, store);
+
+      const callbacks = generatorCallbacks.get(name) || [];
+      callbacks.forEach(callback => {
+        setTimeout(() => callback(store), 0);
+      });
```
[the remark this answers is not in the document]
AbanoubGhadban: Refactored into separate class and used in ComponentRegistery as well

lib/react_on_rails/helper.rb
```diff
@@ -675,17 +709,20 @@ def server_rendered_react_component(render_options) # rubocop:disable Metrics/Cy
...
+      return result unless store_dependencies.present?
 
       declarations = +"var reduxProps, store, storeGenerator;\n"
-      all_stores = (@registered_stores || []) + (@registered_stores_defer_render || [])
+      store_objects = registered_stores_including_deferred.select do |store|
+        store_dependencies.include?(store[:store_name])
+      end
 
-      result << all_stores.each_with_object(declarations) do |redux_store_data, memo|
+      result << store_objects.each_with_object(declarations) do |redux_store_data, memo|
```
Judahmeek, line 722: How do we make sure that registered stores that aren't listed in the `store_dependencies` don't slip through the cracks?

Should we throw a warning if a store is registered that isn't listed in `store_dependencies`?

lib/react_on_rails/packs_generator.rb
```diff
@@ -44,13 +44,72 @@ def create_pack(file_path)
...
+      end
+
+      ""
+    end
+
+    def is_client_entrypoint?(file_path)
+      content = File.read(file_path)
+      # has "use client" directive. It can be "use client" or 'use client'
+      first_js_statement_in_code(content).match?(/^["']use client["'](?:;|\s|$)/)
+    end
```
Judahmeek, line 88: Is all this necessary?

[Client & server entrypoints should be detectable by their filenames.](https://github.com/shakacode/react_on_rails/blob/master/docs/guides/file-system-based-automated-bundle-generation.md#server-rendering-and-client-rendering-components)

lib/react_on_rails/test_helper/webpack_assets_status_checker.rb
```diff
@@ -50,8 +50,8 @@ def stale_generated_files(files)
       def all_compiled_assets
         @all_compiled_assets ||= begin
           webpack_generated_files = @webpack_generated_files.map do |bundle_name|
-            if bundle_name == ReactOnRails.configuration.server_bundle_js_file
-              ReactOnRails::Utils.server_bundle_js_file_path
+            if bundle_name == ReactOnRails.configuration.react_client_manifest_file
+              ReactOnRails::Utils.react_client_manifest_file_path
```
Judahmeek, line 54: This doesn't make sense.

Neither the `ReactOnRails.configuration.react_client_manifest_file` or the `ReactOnRails::Utils.react_client_manifest_file_path` currently exist.

Also, I don't see why you would want to replace the `server_bundle_js_file` reference with a `react_client_manifest_file` reference when the `bundle_js_file_path` below basically does the same thing.

Judahmeek (commented):
Other small issues on the side, the primary implementation of using a CallbackRegistry looks good.

lib/react_on_rails/packs_generator.rb
AbanoubGhadban: React server and client components differ significantly from the server and client files in React on Rails:

- **React on Rails server and client files**:  
  In this system, `packs/*.server.js` files are bundled into the server bundle, while `*.client.js` files are bundled into the client bundle.

- **React server and client components**:  
  This is a new system introduced by React itself. By default, all components are considered server components and are excluded from the client bundle. To designate a component as a client component (an entry point for the client bundle), the `"use client";` directive must be added at the top of the file. This system is part of React’s core functionality, not specific to React on Rails.

This code ensures compatibility between both systems within React on Rails.

lib/react_on_rails/test_helper/webpack_assets_status_checker.rb
AbanoubGhadban: Reverted. This code has been moved to PR #1663, which introduced `ReactOnRails.configuration.react_client_manifest_file`.

I did not replace the `server_bundle_js_file` reference with the `react_client_manifest_file` reference because the original code appears to have been written a long time ago when the server bundle path was handled differently. However, with the current implementation, the following code:  

```ruby
if bundle_name == ReactOnRails.configuration.server_bundle_js_file
  ReactOnRails::Utils.server_bundle_js_file_path
else
  ReactOnRails::Utils.bundle_js_file_path(bundle_name)
end
```  

is now effectively equivalent to:  

```ruby
ReactOnRails::Utils.bundle_js_file_path(bundle_name)
```  

This is because `server_bundle_js_file_path` internally calls `bundle_js_file_path(ReactOnRails.configuration.server_bundle_js_file)`. Therefore, the updated code you mentioned enables locating the `react-client-manifest` file path rather than simply replacing `server_bundle` with `react-client-manifest`.

PR #1681, opened 2025-01-16
Export stream functions only for node bundles

AbanoubGhadban:
### Summary
We can't remove `streamServerRenderedReactComponent` from the `ReactOnRails` object in the global scope because it is directly invoked by the JavaScript code responsible for server-side rendering of streamed components, as demonstrated [here](https://github.com/shakacode/react_on_rails_pro/blob/4b7c863/lib/react_on_rails_pro/server_rendering_js_code.rb#L22).  

It is not feasible to use:  
```js
import streamServerRenderedReactComponent from 'react-on-rails/stream';
```  
within the Node.js renderer.  

To avoid errors, we kept `streamServerRenderedReactComponent` inside the `ReactOnRails` object when the package is imported in a Node.js bundle. Importing this function directly in a non-Node.js bundle, such as a web bundle, will result in a runtime error.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file  

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1681)


spec/dummy/package.json
```diff
@@ -72,12 +72,13 @@
     "fs": false
   },
   "scripts": {
-    "yalc-postinstall": "yalc link react-on-rails",
+    "preinstall": "yarn run link-source && yalc add --link react-on-rails",
```
AbanoubGhadban, line 75: While not part of the original PR goal, this change addressed the issue of occasionally failing to utilize the latest updates in the local `react-on-rails`.

PR #1715, opened 2025-03-17
Fix TypeScript issues

alexeyr-ci2:
### Summary

Follow-up to #1713: enable type-driven rules in TypeScript-ESLint and fix corresponding issues

### Pull Request checklist

- ~[ ] Update documentation~
- ~[ ] Update CHANGELOG file~

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1715)

## Summary by CodeRabbit

- **Refactor**
  - Enhanced asynchronous processing during component rendering and store hydration, enabling smoother, non-blocking updates.
  - Optimized page lifecycle callbacks to better manage concurrent operations.
  - Improved type safety for various components and functions.

- **Bug Fixes**
  - Improved error handling and type-safety measures to reduce potential runtime issues and false-positive alerts.

- **Chores**
  - Updated linting rules and configuration settings for improved code reliability.
  - Added new configuration file for TypeScript linting.
  - Refined inline documentation to aid future maintenance.

### DIFF
--- a/node_package/src/ClientSideRenderer.ts
+++ b/node_package/src/ClientSideRenderer.ts
@@ -16,7 +16,7 @@ const REACT_ON_RAILS_STORE_ATTRIBUTE = 'data-js-react-on-rails-store';
 
 function delegateToRenderer(
   componentObj: RegisteredComponent,
-  props: Record<string, string>,
+  props: Record<string, unknown>,
   railsContext: RailsContext,
   domNodeId: string,
   trace: boolean,
@@ -81,7 +81,7 @@ class ComponentRenderer {
     // This must match lib/react_on_rails/helper.rb
     const name = el.getAttribute('data-component-name') || '';
     const { domNodeId } = this;
-    const props = el.textContent !== null ? JSON.parse(el.textContent) : {};
+    const props = el.textContent !== null ? (JSON.parse(el.textContent) as Record<string, unknown>) : {};
     const trace = el.getAttribute('data-trace') === 'true';
 
     try {
@@ -183,7 +183,10 @@ class StoreRenderer {
     }
 
     const name = storeDataElement.getAttribute(REACT_ON_RAILS_STORE_ATTRIBUTE) || '';
-    const props = storeDataElement.textContent !== null ? JSON.parse(storeDataElement.textContent) : {};
+    const props =
+      storeDataElement.textContent !== null
+        ? (JSON.parse(storeDataElement.textContent) as Record<string, unknown>)
+        : {};
     this.hydratePromise = this.hydrate(context, railsContext, name, props);
   }
 
@@ -191,7 +194,7 @@ class StoreRenderer {
     context: Context,
     railsContext: RailsContext,
     name: string,
-    props: Record<string, string>,
+    props: Record<string, unknown>,
   ) {
     const storeGenerator = await context.ReactOnRails.getOrWaitForStoreGenerator(name);
     if (this.state === 'unmounted') {
--- a/node_package/src/ClientSideRenderer.ts
+++ b/node_package/src/ClientSideRenderer.ts
@@ -14,13 +14,13 @@ import { debugTurbolinks } from './turbolinksUtils';
 
 const REACT_ON_RAILS_STORE_ATTRIBUTE = 'data-js-react-on-rails-store';
 
-function delegateToRenderer(
+async function delegateToRenderer(
   componentObj: RegisteredComponent,
   props: Record<string, unknown>,
   railsContext: RailsContext,
   domNodeId: string,
   trace: boolean,
-): boolean {
+): Promise<boolean> {
   const { name, component, isRenderer } = componentObj;
 
   if (isRenderer) {
@@ -32,7 +32,7 @@ function delegateToRenderer(
       );
     }
 
-    (component as RenderFunction)(props, railsContext, domNodeId);
+    await (component as RenderFunction)(props, railsContext, domNodeId);
     return true;
   }
 
@@ -92,7 +92,7 @@ class ComponentRenderer {
           return;
         }
 
-        if (delegateToRenderer(componentObj, props, railsContext, domNodeId, trace)) {
+        if (await delegateToRenderer(componentObj, props, railsContext, domNodeId, trace)) {
           return;
         }
 
--- a/node_package/src/ClientSideRenderer.ts
+++ b/node_package/src/ClientSideRenderer.ts
@@ -222,7 +222,7 @@ const renderedRoots = new Map<string, ComponentRenderer>();
 
 export function renderOrHydrateComponent(domIdOrElement: string | Element): ComponentRenderer | undefined {
   const domId = getDomId(domIdOrElement);
-  debugTurbolinks(`renderOrHydrateComponent ${domId}`);
+  debugTurbolinks('renderOrHydrateComponent', domId);
   let root = renderedRoots.get(domId);
   if (!root) {
     root = new ComponentRenderer(domIdOrElement);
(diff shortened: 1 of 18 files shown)


node_package/src/ClientSideRenderer.ts
```diff
@@ -32,7 +32,7 @@ function delegateToRenderer(
       );
     }
 
-    (component as RenderFunction)(props, railsContext, domNodeId);
+    await (component as RenderFunction)(props, railsContext, domNodeId);
```
alexeyr-ci2, line 35: Wasn't properly awaited before.

node_package/src/ReactOnRails.client.ts
```diff
@@ -345,7 +343,7 @@ ctx.ReactOnRails = {
 
 ctx.ReactOnRails.resetOptions();
 
-ClientStartup.clientStartup(ctx);
+void ClientStartup.clientStartup(ctx);
```
alexeyr-ci2: Probably can't use top-level await here, right?

node_package/src/clientStartup.ts
```diff
@@ -11,7 +11,9 @@ import { debugTurbolinks } from './turbolinksUtils';
 
 export async function reactOnRailsPageLoaded() {
   debugTurbolinks('reactOnRailsPageLoaded');
-  await Promise.all([hydrateAllStores(), renderOrHydrateAllComponents()]);
+  const promise = hydrateAllStores();
+  renderOrHydrateAllComponents();
```
alexeyr-ci2: `renderOrHydrateAllComponents` is not async.

node_package/src/reactHydrateOrRender.ts
```diff
@@ -7,38 +7,39 @@ type HydrateOrRenderType = (domNode: Element, reactElement: ReactElement) => Ren
...
   }
 }
 
-/* eslint-disable react/no-deprecated -- while we need to support React 16 */
+/* eslint-disable react/no-deprecated,@typescript-eslint/no-deprecated,@typescript-eslint/no-non-null-assertion --
+ * while we need to support React 16
+ */
 const reactHydrate: HydrateOrRenderType = supportsRootApi
-  ? reactDomClient.hydrateRoot
+  ? reactDomClient!.hydrateRoot
```
alexeyr-ci2, line 29: TypeScript thinks `reactDomClient` may not be defined here, even though it definitely is.

alexeyr-ci2:
@coderabbitai review

node_package/src/pageLifecycle.ts
```diff
@@ -69,17 +73,17 @@ function initializePageEventListeners(): void {
...
   }
   pageLoadedCallbacks.add(callback);
   initializePageEventListeners();
 }
 
-export function onPageUnloaded(callback: PageLifecycleCallback): void {
+export async function onPageUnloaded(callback: PageLifecycleCallback): Promise<void> {
   if (currentPageState === 'unload') {
-    callback();
+    await callback();
```
alexeyr-ci2: @AbanoubGhadban Or do you think these callbacks shouldn't be awaited?

node_package/src/transformRSCStreamAndReplayConsoleLogs.ts
```diff
@@ -25,7 +27,7 @@ export default function transformRSCStreamAndReplayConsoleLogs(stream: ReadableS
 
         for (const jsonChunk of jsonChunks) {
           const { html, consoleReplayScript = '' } = jsonChunk;
-          controller.enqueue(encoder.encode(html));
+          controller.enqueue(encoder.encode(html ?? ''));
```
alexeyr-ci2, line 30: @AbanoubGhadban The old code would encode `null` as the string `"null"`. Is that the desired behavior? Or maybe we should use
```
if (html) {
  controller.enqueue(encoder.encode(html));
}
```
?

node_package/src/ClientSideRenderer.ts
```diff
@@ -92,7 +92,7 @@ class ComponentRenderer {
           return;
         }
 
-        if (delegateToRenderer(componentObj, props, railsContext, domNodeId, trace)) {
+        if (await delegateToRenderer(componentObj, props, railsContext, domNodeId, trace)) {
```
AbanoubGhadban:
```suggestion
        // Abort if component unmounted during awaiting the async operation
        if (await delegateToRenderer(componentObj, props, railsContext, domNodeId, trace) || this.state === 'unmounted') {
```

node_package/src/ClientSideRenderer.ts
AbanoubGhadban: The renderer function is not handled properly on unmount. It has existed from the beginning in react-on-rails, and I was planning to fix it in a follow-up PR. However, awaiting the renderer function is a good first step.

node_package/src/ReactOnRails.client.ts
AbanoubGhadban: Yes

node_package/src/pageLifecycle.ts
AbanoubGhadban: This callback is conditionally awaited only if the page has been unloaded. As a result, the caller of the `onPageUnloaded` function may mistakenly assume that the callback is awaited, when in fact it is not executed.

node_package/src/transformRSCStreamAndReplayConsoleLogs.ts
AbanoubGhadban: Good catch, lets keep your implementation

node_package/src/clientStartup.ts
```diff
@@ -36,9 +38,9 @@ export async function clientStartup(context: Context): Promise<void> {
 
   // force loaded components and stores are rendered and hydrated immediately
   renderOrHydrateForceLoadedComponents();
-  hydrateForceLoadedStores();
+  await hydrateForceLoadedStores();
```
AbanoubGhadban: Do not wait here. The hydration process can handle the concurrent hydration of components and stores. Therefore, let's avoid blocking the hydration of non-force loaded components, in case the page is already loaded.

node_package/src/clientStartup.ts
AbanoubGhadban: Lets update `renderOrHydrateComponent` and make it await until the component is hydrated
replace
```js
return root;
```
with
```js
await root.waitUntilRendered();
```

Then, we can make `renderOrHydrateForceLoadedComponents`, `renderOrHydrateAllComponents` and `reactOnRailsComponentLoaded` functions async function

node_package/src/ClientSideRenderer.ts
alexeyr-ci2: Reverted.

node_package/src/clientStartup.ts
alexeyr-ci2: https://github.com/shakacode/react_on_rails/commit/e308754d5a391e572e00ab262548355b0cc9b597.

node_package/src/clientStartup.ts
alexeyr-ci2: https://github.com/shakacode/react_on_rails/commit/c00e52f09023cb9d5dda97e36a974f20caf774b8.

alexeyr-ci2: (and 48d03d11).

alexeyr-ci2:
> Lets update renderOrHydrateComponent and make it await until the component is hydrated
replace

alexeyr-ci2:
@AbanoubGhadban 

> Lets update renderOrHydrateComponent and make it await until the component is hydrated
replace

Since I can't reply anymore, fixed version of this is 48d03d11.

AbanoubGhadban (approved):
LGTM

PR #1728, opened 2025-04-10
Enable Jest ESLint plugin with recommended config

alexeyr-ci2:
### Summary

Include the Jest ESLint plugin and fix the tests accordingly.

### Pull Request checklist

- ~[ ] Update documentation~
- ~[ ] Update CHANGELOG file~

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1728)

## Summary by CodeRabbit

- **New Features**
  - Enhanced linting configuration with improved Jest integration for test files.
- **Tests**
  - Removed explicit assertion count checks across multiple test suites for clearer test code.
  - Refined test descriptions and reorganized setup hooks for better test isolation.
  - Adjusted error handling structure in asynchronous tests without changing behavior.
  - Updated some assertions to use strict equality checks for improved precision.
- **Chores**
  - Added `eslint-plugin-jest` and updated `typescript-eslint` dependency versions.
  - Updated scripts to manage package.json dependencies more effectively.


node_package/tests/ComponentRegistry.test.js
```diff
@@ -176,10 +176,12 @@ describe('ComponentRegistry', () => {
...
     expect.assertions(1);
+    let error;
     try {
       await ComponentRegistry.getOrWaitForComponent('NonExistent');
-    } catch (error) {
-      expect(error.message).toMatch(/Could not find component/);
+    } catch (e) {
+      error = e;
     }
+    expect(error.message).toMatch(/Could not find component/);
```
alexeyr-ci2, line 174: Avoids conditional `expect` (though `expect.assertions` will catch the problem if there is no error too, we could just disable the rule here instead).

node_package/tests/Authenticity.test.js
```diff
@@ -8,28 +8,22 @@ meta.content = testToken;
 document.head.appendChild(meta);
 
 describe('authenticityToken', () => {
-  expect.assertions(2);
```
alexeyr-ci2, line 11: `eslint-plugin-jest` bans these `expect.assertions(...)` calls outside actual tests. 

The ones in them are allowed but not actually useful, they are good in two cases:
1. testing async logic with callbacks instead of `async`/`await`;
2. conditional `expect` including in `try`/`catch`, also banned by the plugin.

Judahmeek (approved):

PR #1735, opened 2025-04-18
Fix `yarn`/`bundle install` in CI

alexeyr-ci2:
### Summary

* Cache actions for `node_modules` and `vendor/bundle` should use lockfile hash in their key, not the main `package.json`/`Gemfile`
* Run `yarn install` and `bundle install` in frozen lockfile mode to be sure the lockfiles are updated.

### Pull Request checklist

- ~[ ] Update documentation~
- ~[ ] Update CHANGELOG file~

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1735)

## Summary by CodeRabbit

- **Chores**
	- Improved workflow dependency caching by switching cache keys to use lockfiles for both Node and Ruby dependencies, enhancing cache accuracy.
	- Updated environment variables and install commands to enforce stricter or conditional dependency installation based on the matrix version, increasing consistency and reliability in CI runs.
	- Adjusted Yarn and bundle install steps to use the `--frozen-lockfile` flag and `BUNDLE_FROZEN` variable where appropriate.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -8,12 +8,13 @@ on:
 
 jobs:
   examples:
-    env:
-      SKIP_YARN_COREPACK_CHECK: 0
     strategy:
       fail-fast: false
       matrix:
         versions: ['oldest', 'newest']
+    env:
+      SKIP_YARN_COREPACK_CHECK: 0
+      BUNDLE_FROZEN: ${{ matrix.versions == 'oldest' && 'false' || 'true' }}
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
@@ -71,7 +72,7 @@ jobs:
         run: echo "sha=\"$(git rev-parse HEAD)\"" >> "$GITHUB_OUTPUT"
       - name: Install Node modules with Yarn for renderer package
         run: |
-          yarn install --no-progress --no-emoji
+          yarn install --no-progress --no-emoji ${{ matrix.versions == 'newest' && '--frozen-lockfile' || '' }}
           sudo yarn global add yalc
       - name: yalc publish for react-on-rails
         run: yalc publish
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -46,6 +46,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
+          cache: yarn
+          cache-dependency-path: '**/yarn.lock'
       - name: Print system information
         run: |
           echo "Linux release: "; cat /etc/issue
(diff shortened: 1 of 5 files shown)


.github/workflows/examples.yml
```diff
@@ -45,6 +46,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
+          cache: yarn
+          cache-dependency-path: '**/yarn.lock'
```
alexeyr-ci2, line 50: See https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data and https://github.com/actions/cache/blob/main/examples.md#node---npm for why this is preferred over caching `node_modules`. In our case it also avoids duplication between `node_modules` and `spec/dummy/node_modules`.

.github/workflows/examples.yml
```diff
@@ -57,21 +60,16 @@ jobs:
...
-        uses: actions/cache@v4
-        with:
-          path: node_modules
-          key: v5-package-node-modules-cache-${{ hashFiles('yarn.lock') }}
       - name: Save root ruby gems to cache
         uses: actions/cache@v4
         with:
           path: vendor/bundle
-          key: package-app-gem-cache-${{ hashFiles('react_on_rails.gemspec') }}-${{ hashFiles('Gemfile.development_dependencies') }}-${{ matrix.versions }}
+          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-${{ matrix.versions }}
```
alexeyr-ci2, line 67: `setup-ruby` documentation [says `action/cache` isn't enough](https://github.com/ruby/setup-ruby?tab=readme-ov-file#caching-bundle-install-manually) and [`action/cache` docs](https://github.com/actions/cache/blob/main/examples.md#ruby---bundler) link to it.

But I am [awaiting their reply](https://github.com/ruby/setup-ruby/discussions/748) for what to do for multiple `Gemfile.lock`s in a single repo.

Judahmeek: `ruby-setup` with `bundle-cache: true` and a `BUNDLE_GEMFILE` set to a specific Gemfile set should work, right?

Judahmeek (approved):

.github/workflows/examples.yml
alexeyr-ci2: `BUNDLE_GEMFILE` can only be set to a single Gemfile, right? Not to both root and spec/dummy Gemfiles at once. Their recommendation is actually to run `setup-ruby` twice with different Gemfiles, but I decided to avoid this change for now.

PR #1736, opened 2025-04-27
[WIP] Add ability to render server components inside client components (add support for react-router)

AbanoubGhadban:
# Rendering React Server Components Inside Client Components

This PR introduces a powerful new capability to the React on Rails framework: the ability to render React Server Components (RSC) directly inside React Client Components. This bypasses a fundamental limitation of React's component model while maintaining performance and preventing hydration issues.

## 📋 Overview

> [!IMPORTANT]
> This feature should be used judiciously. It's best suited for server components whose props change very rarely, such as router routes. **Do not** use this with components whose props change frequently as it triggers HTTP requests to the server on each re-render.

### Before this PR

Server Components could only be embedded inside Client Components if passed as a prop from a parent Server Component:

```tsx
// Parent Server Component
export default function Parent() {
  return (
    <ClientComponent>
      <ServerComponent />
    </ClientComponent>
  );
}
```

### After this PR

Server Components can be directly rendered inside Client Components using the new `RSCRoute` component:

```tsx
'use client';

import RSCRoute from 'react-on-rails/RSCRoute';

export default function ClientComponent() {
  return (
    <div>
      <RSCRoute componentName="ServerComponent" componentProps={{ user }} />
    </div>
  );
}
```

### 🔄 RSC Payload Lifecycle

The RSC payload lifecycle follows two distinct paths depending on whether the component is being server-side rendered or client-side rendered.

#### Server-Side Rendering Flow

1. **Client Component Renders RSCRoute**

   - A client component includes `<RSCRoute componentName="ServerComponent" componentProps={props} />`
   - During SSR, this triggers a request for the server component

2. **RSC Payload Generation**

   - `RSCRoute` communicates with the RSC bundle via `getRSCPayloadStream`
   - The RSC bundle generates the RSC payload and returns it as a stream

3. **Server-Side HTML Generation**

   - The server uses the RSC payload to render HTML for the server component
   - `injectRSCPayload` function collects and tracks all RSC payloads

4. **Payload Embedding**
   - RSC payloads are embedded directly into the HTML as script tags
   - Each payload is stored in a global array using the pattern:
     ```js
     <script>(self.REACT_ON_RAILS_RSC_PAYLOADS||={})["ComponentName-props-renderRequestId"]||=[]</script>
     <script>(self.REACT_ON_RAILS_RSC_PAYLOADS["ComponentName-props-renderRequestId"]).push(chunk)</script>
     ```

#### Client-Side Rendering Flow

1. **Initial Hydration**

   - When `RSCRoute` renders during hydration, it checks for embedded payloads
   - If found, it uses the embedded payload without making a network request

2. **Post-Navigation Rendering**

   - For components rendered after client-side navigation or state changes:
     - No embedded payload exists for the new component/props combination
     - `RSCRoute` makes an HTTP request to fetch the RSC payload
     - Request goes to `/${rscPayloadGenerationUrl}/${componentName}?props=${propsString}`

3. **Component Rendering**
   - The fetched or embedded payload is used to render the server component
   - Client components referenced in the payload are hydrated with their props

```mermaid
flowchart TD
    A[Client Component<br>renders RSCRoute] --> B{Is SSR?}

    %% Server-side path
    B -->|Yes| C[RSCRoute communicates<br>with RSC bundle]
    C --> D[RSC payload generated]
    D --> E[Server renders HTML<br>using RSC payload]
    E --> F[RSC payload embedded<br>in HTML response]
    F --> G[HTML sent to browser]

    %% Client-side paths
    G --> H[Browser renders HTML]
    H --> I{Component<br>previously SSRed?}

    %% Hydration path
    I -->|Yes| J[Use embedded<br>RSC payload]
    J --> K[Hydrate components<br>without network request]

    %% Client navigation path
    B -->|No| L[Make HTTP request<br>for RSC payload]
    I -->|No| L
    L --> M[Process RSC payload]
    M --> N[Render server component<br>with client components]

    %% Styling
    classDef serverProcess fill:#f9f,stroke:#333,stroke-width:2px
    classDef clientProcess fill:#bbf,stroke:#333,stroke-width:2px
    classDef dataFlow fill:#bfb,stroke:#333,stroke-width:2px

    class C,D,E,F serverProcess
    class A,H,I,J,K,L,M,N clientProcess
    class G dataFlow
```

The key innovation in this architecture is how it handles the two rendering scenarios differently:

1. **For SSR + Hydration**: Embeds payloads directly in the HTML to eliminate network requests
2. **For Client Navigation**: Makes on-demand HTTP requests for new server components

This dual approach provides the best of both worlds: fast initial page loads with embedded data and dynamic component loading for subsequent interactions.

#### RSCProvider

At the foundation of this architecture is the `RSCProvider`, which provides the core functionality for fetching and rendering server components.

```tsx
// Interface
type RSCContextType = {
  getCachedComponent: (componentName: string, componentProps: unknown) => React.ReactNode;
  getComponent: (componentName: string, componentProps: unknown) => Promise<React.ReactNode>;
};

// Usage via hook
const { getCachedComponent, getComponent } = useRSC();
```

The `RSCProvider` handles the complex logic of:

- Determining whether to use embedded payloads or fetch from the server
- Caching components to prevent duplicate requests
- Managing request promises to prevent race conditions
- Converting RSC payloads into renderable React elements

This component is environment-aware and behaves differently when:

- On the server: It communicates directly with the RSC bundle
- On the client: It either uses embedded payloads or makes HTTP requests

While you could use the `useRSC` hook directly in client components, it's generally better to use the `RSCRoute` component which provides a simpler interface.

#### wrapServerComponentRenderer

Because the `RSCProvider` has different implementations for client and server environments, you shouldn't create it manually. Instead, the `wrapServerComponentRenderer` function handles this complexity:

```tsx
// Client usage
import wrapServerComponentRenderer from 'react-on-rails/wrapServerComponentRenderer/client';

// Server usage
import wrapServerComponentRenderer from 'react-on-rails/wrapServerComponentRenderer/server';

// Basic usage pattern
const WrappedComponent = wrapServerComponentRenderer(YourClientComponent);
```

This higher-order function:

- Takes a client component or render function
- Wraps it with the appropriate `RSCProvider` implementation
- Properly handles hydration on the client or rendering on the server
- Ensures Suspense boundaries are set up correctly

The wrapped component is ready to be registered with React on Rails and will properly handle server component rendering in any environment.

#### RSCRoute

Building on the foundation provided by `RSCProvider` and `wrapServerComponentRenderer`, the `RSCRoute` component provides the simplest way to render server components:

```tsx
// Interface
type RSCRouteProps = {
  componentName: string;
  componentProps: unknown;
};

// Usage
<RSCRoute componentName="MyServerComponent" componentProps={{ user }} />;
```

`RSCRoute` is the primary component you'll use in your application to:

- Render server components inside client components
- Automatically handle server/client transitions
- Manage caching and fetching of server component data
- Integrate with Suspense for loading states

It uses React's `use` hook to integrate with Suspense boundaries and provide a consistent loading experience across your application.

These three components form a cohesive system that makes server components usable in client components while maintaining performance and preventing hydration mismatches.

### ❌ Bad Example - Frequently Changing Props

```tsx
'use client';
import { useState } from 'react';
import RSCRoute from 'react-on-rails/RSCRoute';

export default function ClientComponent() {
  const [count, setCount] = useState(0);

  return (
    <div>
      <button onClick={() => setCount(count + 1)}>Increment</button>
      <label>Count: {count}</label>
      {/* BAD EXAMPLE: Server Component props change with each button click */}
      <RSCRoute componentName="ServerComponent" componentProps={{ count }} />
    </div>
  );
}
```

> [!WARNING]
> This implementation will make a server request on every state change, significantly impacting performance.

### ✅ Good Example - Router Integration

```tsx
'use client';
import { Routes, Route, Link } from 'react-router-dom';
import RSCRoute from 'react-on-rails/RSCRoute';
import AnotherClientComponent from './AnotherClientComponent';

export default function AppRouter({ user }: { user: any }) {
  return (
    <>
      <nav>
        <Link to="/">Home</Link>
        <Link to="/about">About</Link>
        <Link to="/client-component">Client Component</Link>
      </nav>
      <Routes>
        {/* Mix client and server components in your router */}
        <Route path="/client-component" element={<AnotherClientComponent />} />
        {/* GOOD EXAMPLE: Server Component props rarely change */}
        <Route path="/about" element={<RSCRoute componentName="About" componentProps={{ user }} />} />
        <Route path="/" element={<RSCRoute componentName="Home" componentProps={{ user }} />} />
      </Routes>
    </>
  );
}
```

### 🧩 Nested Route Example

The framework supports nesting client and server components to arbitrary depth:

```tsx
'use client';
import { Routes, Route } from 'react-router-dom';
import RSCRoute from 'react-on-rails/RSCRoute';
import ServerRouteLayout from './ServerRouteLayout';
import ClientRouteLayout from './ClientRouteLayout';
import AnotherClientComponent from './AnotherClientComponent';

export default function AppRouter() {
  return (
    <Routes>
      <Route path="/main-server-route" element={<ServerRouteLayout />}>
        <Route path="/server-subroute" element={<RSCRoute componentName="MyServerComponent" />} />
        <Route path="/client-subroute" element={<AnotherClientComponent />} />
      </Route>
      <Route path="/main-client-route" element={<ClientRouteLayout />}>
        <Route path="/client-subroute" element={<AnotherClientComponent />} />
        <Route path="/server-subroute" element={<RSCRoute componentName="MyServerComponent" />} />
      </Route>
    </Routes>
  );
}
```

To use `Outlet` in server components, create a client version:

```tsx
// ./components/Outlet.tsx
'use client';
export { Outlet as default } from 'react-router-dom';
```

Then use in server components:

```tsx
// ./components/ServerRouteLayout.tsx
import Outlet from './Outlet';

export default function ServerRouteLayout() {
  return (
    <div>
      <h1>Server Route Layout</h1>
      <Outlet />
    </div>
  );
}
```

## 🔧 Internal Implementation

This section details the inner workings of the React Server Components integration in React on Rails, explaining how server components are generated, transported, and rendered.

### getReactServerComponent

At the core of the RSC implementation is the `getReactServerComponent` function, which has environment-specific implementations.

#### Server Implementation

```tsx
// Simplified interface
type RSCServerRootProps = {
  componentName: string;
  componentProps: unknown;
  railsContext: RailsContext;
};

const getReactServerComponent = async ({
  componentName,
  componentProps,
  railsContext,
}: RSCServerRootProps) => {
  // Implementation details...
};
```

On the server, this function:

1. Validates that required context properties exist
2. Creates an SSR manifest by loading client and server manifests
   ```tsx
   const ssrManifest = await createSSRManifest(
     railsContext.reactServerClientManifestFileName,
     railsContext.reactClientManifestFileName,
   );
   ```
3. Gets the RSC payload stream via `ReactOnRails.getRSCPayloadStream` (Explained in [RSCPayloadGenerator](#rscpayloadgenerator))
4. Processes the stream with React's runtime to create React elements
   ```tsx
   return createFromReactOnRailsNodeStream(rscPayloadStream, ssrManifest);
   ```

The manifest creation process maps module IDs between server and client bundles, allowing React to correctly associate server components with their client counterparts.

#### Client Implementation

```tsx
// Simplified interface
type ClientGetReactServerComponentProps = {
  componentName: string;
  componentProps: unknown;
  railsContext: RailsContext;
};

const getReactServerComponent = ({
  componentName,
  componentProps,
  railsContext,
}: ClientGetReactServerComponentProps) => {
  // Implementation details...
};
```

On the client, this function:

1. Generates a unique cache key based on component name, props, and request ID
2. Checks if an embedded payload exists in `window.REACT_ON_RAILS_RSC_PAYLOADS[key]`
3. If found, creates a stream from the array and processes it
   ```tsx
   if (payloads) {
     return createFromPreloadedPayloads(payloads);
   }
   ```
4. If not found, makes an HTTP request to fetch the payload
   ```tsx
   return fetchRSC({ componentName, componentProps, railsContext });
   ```

Both paths ultimately use React's `createFromReadableStream` to convert the RSC payload into React elements.

```mermaid
flowchart TD
    A[getReactServerComponent] --> B{Environment?}
    
    B -->|Server| C[Validate railsContext]
    C --> D[Load client & server manifests]
    D --> E[Get RSC payload stream]
    E --> F[Process with createFromNodeStream]
    F --> G[Return React Elements]
    
    B -->|Client| H[Generate cache key]
    H --> I{Embedded payload<br>exists?}
    I -->|Yes| J[Create stream from<br>embedded payload]
    I -->|No| K[Fetch payload via HTTP]
    J --> L[Process with createFromReadableStream]
    K --> L
    L --> G
    
    classDef serverProcess fill:#f9f,stroke:#333,stroke-width:2px
    classDef clientProcess fill:#bbf,stroke:#333,stroke-width:2px
    classDef common fill:#fff,stroke:#333,stroke-width:2px
    
    class C,D,E,F serverProcess
    class H,I,J,K,L clientProcess
    class A,B,G common
```

### RSCProvider

The `RSCProvider` creates a React context for server component operations. The provider accepts a `getServerComponent` function, making it environment-agnostic. It can work with either the server or client implementation of `getReactServerComponent`:

```tsx
// Simplified interface
type RSCContextType = {
  getCachedComponent: (componentName: string, componentProps: unknown) => React.ReactNode;
  getComponent: (componentName: string, componentProps: unknown) => Promise<React.ReactNode>;
};

// Factory function
export const createRSCProvider = ({
  railsContext,
  getServerComponent,
}: {
  railsContext: RailsContext;
  getServerComponent: typeof getReactServerComponent;
}) => {
  // Implementation details...
  return ({ children }: { children: React.ReactNode }) => (
    <RSCContext.Provider value={{ getCachedComponent, getComponent }}>
      {children}
    </RSCContext.Provider>
  );
};
```

Key aspects of the implementation:

1. **Component Caching**: Uses in-memory cache to store already rendered components
   ```tsx
   const cachedComponents: Record<string, React.ReactNode> = {};
   ```

2. **Promise Tracking**: Prevents duplicate requests for the same component
   ```tsx
   const fetchRSCPromises: Record<string, Promise<React.ReactNode>> = {};
   ```

3. **Cache Key Generation**: Creates unique keys based on component name, props, and request ID
   ```tsx
   const generateCacheKey = (componentName: string, componentProps: unknown) => {
     return `${componentName}-${JSON.stringify(componentProps)}-${railsContext.componentSpecificMetadata?.renderRequestId}`;
   };
   ```

### wrapServerComponentRenderer

This higher-order function wraps components with RSC capability. It creates the proper RSCProvider for each environment, passing the environment-specific implementation of `getServerComponent` to the provider. On the server, it passes the server implementation of `getReactServerComponent`, while on the client, it passes the client implementation.

#### Server Implementation

```tsx
// node_package/src/wrapServerComponentRenderer/server.tsx
const WrapServerComponentRenderer = (componentOrRenderFunction: ReactComponentOrRenderFunction) => {
  // Validation...
  
  const wrapper: RenderFunction = async (props, railsContext) => {
    // Get component...
    
    const RSCProvider = createRSCProvider({
      railsContext,
      getServerComponent: getReactServerComponent,
    });

    return () => (
      <RSCProvider>
        <React.Suspense fallback={null}>
          <Component {...props} />
        </React.Suspense>
      </RSCProvider>
    );
  };

  return wrapper;
};
```

#### Client Implementation

```tsx
// node_package/src/wrapServerComponentRenderer/client.tsx
const WrapServerComponentRenderer = (componentOrRenderFunction: ReactComponentOrRenderFunction) => {
  // Validation...
  
  const wrapper: RenderFunction = async (props, railsContext, domNodeId) => {
    // Get component...
    
    const RSCProvider = createRSCProvider({
      railsContext,
      getServerComponent: getReactServerComponent,
    });

    const root = (
      <RSCProvider>
        <React.Suspense fallback={null}>
          <Component {...props} />
        </React.Suspense>
      </RSCProvider>
    );

    // DOM operations
    if (domNode.innerHTML) {
      ReactDOMClient.hydrateRoot(domNode, root, { identifierPrefix: domNodeId });
    } else {
      ReactDOMClient.createRoot(domNode, { identifierPrefix: domNodeId }).render(root);
    }
    
    return '';
  };

  return wrapper;
};
```

Key differences:

1. **Server**: Returns a render function that produces React elements
2. **Client**: Returns a renderer function that performs DOM operations (hydration/rendering)
3. **Both**: Wrap the component in RSCProvider and Suspense

This function replaces the previous `RSCServerRoot` and `RSCClientRoot` components, providing a more flexible approach that works with any client component or render function containing server components. The wrapped component can be either a React component or a render function that returns a React component on the server, or a renderer function on the client.

```mermaid
sequenceDiagram
    participant App as Application Code
    participant WSCR as WrapServerComponentRenderer
    participant RSCP as RSCProvider
    participant RSC as RSCRoute
    participant React as React Runtime
    
    App->>WSCR: Pass client component
    WSCR->>WSCR: Create render function on server and renderer function on client
    App->>WSCR: Call render or renderer with props and railsContext
    WSCR->>RSCP: Create RSCProvider for server or client
    WSCR->>React: Wrap component in Suspense
    
    alt Server Environment
        WSCR->>App: Return react component that will be rendered and streamed to the client
    else Client Environment
        WSCR->>React: Hydrate or render the component after wrapping in RSCProvider
    end
    
    RSC->>RSCP: useRSC()
    RSCP->>RSC: Return context methods
```

### RSCPayloadGenerator

The `RSCPayloadGenerator` coordinates the generation and tracking of RSC payloads:

```tsx
// Simplified interface
export const getRSCPayloadStream = async (
  componentName: string,
  props: unknown,
  railsContext: RailsContext,
): Promise<NodeJS.ReadableStream> => {
  // Implementation details...
};

export const onRSCPayloadGenerated = (
  railsContext: RailsContext, 
  callback: RSCPayloadCallback
) => {
  // Implementation details...
};

export const getRSCPayloadStreams = (
  railsContext: RailsContext
): RSCPayloadStreamInfo[] => {
  // Implementation details...
};

export const clearRSCPayloadStreams = (
  railsContext: RailsContext
) => {
  // Implementation details...
};
```

The implementation:

1. **Payload Generation**: Calls the global `generateRSCPayload` function to get the RSC stream
   ```tsx
   const stream = await generateRSCPayload(componentName, props, railsContext);
   ```

2. **Stream Duplication**: Creates duplicate streams to serve different consumers: one will be used to SSR the server component and one will be used to inject the payload into the HTML
   ```tsx
   const stream1 = new PassThrough();
   stream.pipe(stream1);
   const stream2 = new PassThrough();
   stream1.pipe(stream2);
   ```

3. **Payload Tracking**: Stores streams by railsContext for later injection
   ```tsx
   const streamInfo: RSCPayloadStreamInfo = {
     componentName,
     props,
     stream: stream2,
   };
   streams.push(streamInfo);
   mapRailsContextToRSCPayloadStreams.set(railsContext, streams);
   ```

4. **Callback Notification**: Notifies registered callbacks about new streams
   ```tsx
   const callbacks = rscPayloadCallbacks.get(railsContext) || [];
   callbacks.forEach((callback) => callback(streamInfo));
   ```

These functions are exposed globally via the ReactOnRails object, making them accessible anywhere in the application.

### injectRSCPayload

The `injectRSCPayload` function embeds RSC payloads into the HTML stream:

```tsx
// Simplified interface
export default function injectRSCPayload(
  pipeableHtmlStream: NodeJS.ReadableStream | PipeableStream,
  railsContext: RailsContext,
) {
  // Implementation details...
  return resultStream;
}
```

Key aspects:

1. **Stream Initialization**: Creates a result stream to hold HTML + RSC payloads
   ```tsx
   const resultStream = new PassThrough();
   ```

2. **RSC Payload Monitoring**: Listens for newly generated payloads
   ```tsx
   ReactOnRails.onRSCPayloadGenerated?.(railsContext, (streamInfo) => {
     // Process each RSC payload stream
   });
   ```

3. **Array Initialization**: Creates a global array to store payload chunks
   ```tsx
   // Generates JavaScript like:
   // (self.REACT_ON_RAILS_RSC_PAYLOADS||={})["ComponentName-props-id"]||=[]
   initializeCacheKeyJSArray(cacheKey, resultStream);
   ```

4. **Chunk Embedding**: Writes each chunk to the result stream as a script tag
   ```tsx
   // Generates JavaScript like:
   // (self.REACT_ON_RAILS_RSC_PAYLOADS["ComponentName-props-id"]).push(chunk)
   writeChunk(JSON.stringify(decodedChunk), resultStream, cacheKey);
   ```

5. **HTML Processing**: Passes through HTML while ensuring payloads are injected
   ```tsx
   htmlStream.on('data', (chunk: Buffer) => {
     htmlBuffer.push(chunk);
     // Process HTML chunks
   });
   ```

The timing of these operations is critical for proper hydration:

```mermaid
sequenceDiagram
    participant App as Server Rendering
    participant HTML as HTML Stream
    participant RSCGen as RSCPayloadGenerator
    participant Inject as injectRSCPayload
    participant Result as Combined HTML+RSC Stream
    
    App->>HTML: Begin HTML rendering
    App->>RSCGen: Generate RSC payload
    RSCGen-->>Inject: Notify via callback
    Inject->>Result: Write array initialization
    
    loop For each RSC chunk
        RSCGen-->>Inject: Stream chunk
        Inject->>Result: Write push statement
    end
    
    HTML-->>Inject: Stream HTML chunks
    Inject->>Result: Pass through HTML
    
    Result-->>App: Complete response
```

The entire system ensures that:
- RSC payloads are generated on demand
- Payloads are efficiently tracked and embedded in responses
- Client-side code can access payloads without additional network requests
- Hydration happens correctly and in the right order

This complex orchestration allows server components to be rendered inside client components while maintaining performance and preventing hydration mismatches.

## 🔄 Preventing Race Conditions During Hydration

A critical aspect of the RSC implementation is preventing race conditions during component hydration. Specifically, we need to ensure that:

**The Race Condition**: Without careful implementation, client components might be hydrated before their corresponding RSC payload arrays are initialized in the page, causing unnecessary network requests.

### How We Prevent This Race Condition

Our approach relies on a fundamental behavior of React's hydration process (You can check it out in `SuspenseHydration.test.tsx` and at the `/async_on_server_sync_on_client` page in the RORP dummy app):

1. **React's Hydration Sequence**: Client components only hydrate after their HTML content has been streamed to the client and is present in the DOM.

2. **Early Payload Initialization**: When a server component is rendered during SSR, `getReactServerComponent` calls `ReactOnRails.getRSCPayloadStream` which:
   - Immediately initializes the global payload array in the HTML stream
   - Does this synchronously, before the component finishes rendering
   ```js
   // This script is injected into the HTML stream first
   <script>(self.REACT_ON_RAILS_RSC_PAYLOADS||={})["ComponentName-props-id"]||=[]</script>
   ```

3. **Parallel Streaming**: The HTML stream processes payload chunks and component HTML in parallel after initialization of the payload array:
   1. Payload array initialization script
   2. Payload chunks and component HTML streamed simultaneously

```mermaid
sequenceDiagram
    participant Server as Server Rendering
    participant HTML as HTML Stream
    participant Client as Client Hydration
    
    Server->>HTML: Initialize RSC payload array
    Note right of HTML: <script>(self.REACT_ON_RAILS_RSC_PAYLOADS||={})[key]||=[]</script>
    
    par Parallel Streaming
        Server->>HTML: Stream RSC payload chunks
        Note right of HTML: <script>(self.REACT_ON_RAILS_RSC_PAYLOADS[key]).push(chunk)</script>
        Server->>HTML: Stream component HTML
        Note right of HTML: <div id="component">Component HTML</div>
    end
    
    HTML->>Client: Browser processes HTML & scripts as they arrive
    Client->>Client: Hydration begins only after component HTML is in DOM
    Client->>Client: getReactServerComponent checks for embedded payload
    Note right of Client: RSC payload array exists before hydration starts
```

### Why This Works

This implementation guarantees that:

1. The global `REACT_ON_RAILS_RSC_PAYLOADS` object and the component-specific array are initialized before the component's HTML appears in the document

2. When React begins hydration of the component, it triggers `getReactServerComponent`, which finds the pre-initialized array and uses it instead of making an HTTP request

3. Even in cases where the full payload hasn't finished streaming, the client code can read from the array as new chunks arrive, ensuring progressive rendering

4. If during component hydration, no RSC payload array is found in the global `REACT_ON_RAILS_RSC_PAYLOADS` object, it indicates that the component was not server-side rendered. In this case, `getReactServerComponent` falls back to making an HTTP request to fetch the RSC payload. This scenario occurs in two common cases:

   - **Client Navigation**: When a user navigates to a new route client-side, the component wasn't part of the initial SSR, so no payload array exists
   - **Dynamic Rendering**: When a server component is conditionally rendered based on client-side state changes. For example:
     ```tsx
     function ParentComponent() {
       const [showChild, setShowChild] = useState(false);
       
       return (
         <div>
           <button onClick={() => setShowChild(true)}>Show Child</button>
           {showChild && <RSCRoute componentName="MyServerComponent" />}
         </div>
       );
     }
     ```
     In this case, when `showChild` becomes true, `MyServerComponent` needs to be fetched via HTTP since it wasn't part of the initial SSR.

This approach has been verified through tests like `SuspenseHydration.test.tsx` which confirm that React's hydration behavior consistently follows this pattern, effectively preventing the race condition.

## 📝 Registration Process

Here's a streamlined guide to integrating server components within client components:

### 1. Register the server component in your RSC bundle

First, register any server components you'll need to reference:

```tsx
// packs/server_bundle.ts
import { registerServerComponent } from 'react-on-rails/registerServerComponent/server';
import MyServerComponent from './components/MyServerComponent';

registerServerComponent({ MyServerComponent });
```

### 2. Create your client component with RSCRoute

Build a client component that references server components using RSCRoute:

```tsx
// components/AppRouter.tsx
export default function AppRouter() {
  return (
    <Routes>
      <Route path="/my-server-component" element={<RSCRoute componentName="MyServerComponent" />} />
      <Route path="/" element={<Home />} />
    </Routes>
  )
}
```

#### Client entry point:

```tsx
// components/AppRouter.client.tsx
"use client";

import ReactOnRails from 'react-on-rails';
import { wrapServerComponentRenderer } from 'react-on-rails/wrapServerComponentRenderer/client';
import AppRouter from './components/AppRouter';

const WrappedAppRouter = wrapServerComponentRenderer(AppRouter);

ReactOnRails.register({ AppRouter: WrappedAppRouter });
```

#### Server entry point:

```tsx
// components/AppRouter.server.tsx
import ReactOnRails from 'react-on-rails';
import { wrapServerComponentRenderer } from 'react-on-rails/wrapServerComponentRenderer/server';
import AppRouter from './components/AppRouter';

const WrappedAppRouter = wrapServerComponentRenderer(AppRouter);

ReactOnRails.register({ AppRouter: WrappedAppRouter });
```

### 4. Use in your Rails view

```erb
<%= stream_react_component('AppRouter', props: { some: 'props' }, prerender: true) %>
```

> [!TIP]
> With `auto_load_bundle: true` in your configuration, you can skip the manual `ReactOnRails.register` and `registerServerComponent` calls, but you still need to create both client and server wrappers with the appropriate `wrapServerComponentRenderer` imports.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1736)

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive support for React Server Components (RSC), including new configuration options, context management, and streaming capabilities.
  - Added new components and providers for RSC server-side and client-side rendering.
  - Enabled injection of RSC payloads into HTML streams for more efficient hydration.
  - Exposed new API methods for managing RSC payload streams.

- **Bug Fixes**
  - Improved rendering logic to prevent double rendering and reduce unnecessary HTTP requests.
  - Enhanced error handling and validation for configuration and rendering processes.

- **Documentation**
  - Expanded configuration and changelog documentation to cover new RSC features and integration steps.

- **Tests**
  - Added extensive test coverage for new RSC features, configuration options, and payload injection logic.

- **Chores**
  - Updated dependencies to support new RSC features and maintain compatibility.
  - Improved ESLint configuration and code style enforcement.


eslint.config.ts
```diff
@@ -164,9 +164,7 @@ const config = tsEslint.config([
     languageOptions: {
       parserOptions: {
         projectService: {
-          allowDefaultProject: ['eslint.config.ts', 'knip.ts', 'node_package/tests/*.test.ts'],
-          // Needed because `import * as ... from` instead of `import ... from` doesn't work in this file
-          // for some imports.
```
alexeyr-ci2: Why remove this comment?

knip.ts
```diff
@@ -7,10 +7,12 @@ const config: KnipConfig = {
...
         'node_package/src/ReactOnRailsRSC.ts!',
-        'node_package/src/registerServerComponent/client.ts!',
-        'node_package/src/registerServerComponent/server.ts!',
+        'node_package/src/registerServerComponent/client.tsx!',
+        'node_package/src/registerServerComponent/server.tsx!',
         'node_package/src/registerServerComponent/server.rsc.ts!',
-        'node_package/src/RSCClientRoot.ts!',
+        'node_package/src/wrapServerComponentRenderer/server.tsx!',
+        'node_package/src/wrapServerComponentRenderer/server.rsc.tsx!',
+        'node_package/src/RSCRoute.ts!',
```
alexeyr-ci2, line 15: It's imported in https://github.com/shakacode/react_on_rails/blob/8172e076039c405e654be44d5926e1cb7ae4e6fd/node_package/src/registerServerComponent/client.tsx#L3 so adding it as an `entry` shouldn't be necessary (but doesn't hurt either).

lib/react_on_rails/configuration.rb
```diff
@@ -312,8 +312,6 @@ def ensure_webpack_generated_files_exists
         react_client_manifest_file,
         react_server_client_manifest_file
       ].compact_blank
-
-      self.webpack_generated_files = files
```
alexeyr-ci2, line 316: This should be fixed in the base PR instead.

lib/react_on_rails/react_component/render_options.rb
```diff
@@ -15,9 +15,16 @@ class RenderOptions
       def initialize(react_component_name: required("react_component_name"), options: required("options"))
         @react_component_name = react_component_name.camelize
         @options = options
+        # The render_request_id serves as a unique identifier for each render request.
+        # We cannot rely solely on dom_id, as it should be unique for each component on the page,
+        # but the server can render the same page multiple times concurrently for different users.
+        # Therefore, we need an additional unique identifier that can be used both on the client and server.
+        # This ID can also be used to associate specific data with a particular rendered component
+        # on either the server or client.
+        @render_request_id = self.class.generate_request_id
```
alexeyr-ci2: I don't understand why this is necessary now and wasn't before. 

Or is it needed for RSCs in general and just missed in the previous PRs? If yes, can we easily avoid extra work for non-RSCs, even if there's little of it?

lib/react_on_rails/utils.rb
```diff
@@ -128,6 +128,11 @@ def self.react_server_client_manifest_file_path
       return @react_server_manifest_path if @react_server_manifest_path && !Rails.env.development?
 
       asset_name = ReactOnRails.configuration.react_server_client_manifest_file
+      if asset_name.nil?
+        raise ReactOnRails::Error,
+              "react_server_client_manifest_file is nil, ensure to set it in your configuration"
```
alexeyr-ci2:
```suggestion
              "react_server_client_manifest_file is nil, ensure it is set in your configuration"
```
(or `make sure`).

node_package/src/ClientSideRenderer.ts
```diff
@@ -83,6 +83,18 @@ class ComponentRenderer {
     const { domNodeId } = this;
     const props = el.textContent !== null ? (JSON.parse(el.textContent) as Record<string, unknown>) : {};
     const trace = el.getAttribute('data-trace') === 'true';
+    const renderRequestId = el.getAttribute('data-render-request-id');
+
+    if (!renderRequestId) {
+      console.error(`renderRequestId is missing for ${name} in dom node with id: ${domNodeId}`);
```
alexeyr-ci2: Are the users supposed to do anything in this case?
```suggestion
      console.error(`renderRequestId is missing for component ${name} in the DOM node with id ${domNodeId}`);
```

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,103 @@
...
+  function generateRSCPayload(
+    componentName: string,
+    props: unknown,
+    railsContext: RailsContext,
+  ): Promise<NodeJS.ReadableStream>;
+}
+
+const mapRailsContextToRSCPayloadStreams = new Map<RailsContext, RSCPayloadStreamInfo[]>();
+
+const rscPayloadCallbacks = new Map<RailsContext, Array<RSCPayloadCallback>>();
```
alexeyr-ci2:
```suggestion
const rscPayloadCallbacks = new Map<RailsContext, RSCPayloadCallback[]>();
```
for consistency with the above.

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,103 @@
...
+ * It ensures payload array initialization happens before component HTML appears
+ * in the response stream.
+ *
+ * @param railsContext - Context for the current request
+ * @param callback - Function to call when an RSC payload is generated
+ */
+export const onRSCPayloadGenerated = (railsContext: RailsContext, callback: RSCPayloadCallback) => {
+  const callbacks = rscPayloadCallbacks.get(railsContext) || [];
+  callbacks.push(callback);
+  rscPayloadCallbacks.set(railsContext, callbacks);
```
alexeyr-ci2: `set` shouldn't be necessary if it's already present and does extra work, so maybe
```suggestion
  const callbacks = rscPayloadCallbacks.get(railsContext);
  if (callbacks) {
    callbacks.push(callback);
  } else {
    rscPayloadCallbacks.set(railsContext, [callback]);
  }
```
but it shouldn't matter much.

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,103 @@
...
+  stream.pipe(stream1);
+  const stream2 = new PassThrough();
+  stream1.pipe(stream2);
+
+  const streamInfo: RSCPayloadStreamInfo = {
+    componentName,
+    props,
+    stream: stream2,
+  };
+  streams.push(streamInfo);
```
alexeyr-ci2: I'd move `const streams` definition to just above this line (plus see the previous comment).

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (typeof generateRSCPayload !== 'function') {
+    throw new Error(
+      'generateRSCPayload is not defined. Please ensure that you are using at least version 4.0.0 of ' +
+        'React on Rails Pro and the Node renderer, and that ReactOnRailsPro.configuration.enable_rsc_support ' +
+        'is set to true.',
+    );
+  }
+
+  const stream = await generateRSCPayload(componentName, props, railsContext);
+  const streams = mapRailsContextToRSCPayloadStreams.get(railsContext) ?? [];
```
alexeyr-ci2: You have `||` in most cases. I suggest either `??` or `||` everywhere for consistency.

node_package/src/RSCProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const getComponent = async (componentName: string, componentProps: unknown) => {
+    const cachedComponent = getCachedComponent(componentName, componentProps);
+    if (cachedComponent) {
+      return cachedComponent;
+    }
+    const key = generateCacheKey(componentName, componentProps);
+    if (key in fetchRSCPromises) {
+      return fetchRSCPromises[key];
+    }
```
alexeyr-ci2, line 52:
```suggestion
    const existingPromise = fetchRSCPromises[key];
    if (existingPromise) {
      return existingPromise;
    }
```
to avoid doing the work twice.

node_package/src/RSCProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+      cachedComponents[key] = rsc;
+      return rsc;
+    });
+    fetchRSCPromises[key] = promise;
+    return promise;
+  };
+
+  return ({ children }: { children: React.ReactNode }) => {
+    // eslint-disable-next-line react/jsx-no-constructed-context-values
+    return <RSCContext.Provider value={{ getCachedComponent, getComponent }}>{children}</RSCContext.Provider>;
```
alexeyr-ci2: Should this have `useMemo` as https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md suggests?

If not, please add an explanation.

node_package/src/RSCRoute.ts
```diff
@@ -0,0 +1,40 @@
...
+  componentName: string;
+  componentProps: unknown;
+};
+
+const RSCRoute = ({ componentName, componentProps }: RSCRouteProps) => {
+  const { getComponent, getCachedComponent } = useRSC();
+  let component = getCachedComponent(componentName, componentProps);
+  if (!component) {
+    component = React.use(getComponent(componentName, componentProps));
+  }
```
alexeyr-ci2:
```suggestion
  const component = getCachedComponent(componentName, componentProps) ?? React.use(getComponent(componentName, componentProps));
```

node_package/src/getReactServerComponent.client.ts
```diff
@@ -0,0 +1,137 @@
...
+ * This is used for client-side navigation or when rendering components
+ * that weren't part of the initial server render.
+ *
+ * @param props - Object containing component name, props, and railsContext
+ * @returns A Promise resolving to the rendered React element
+ */
+const fetchRSC = ({ componentName, componentProps, railsContext }: ClientGetReactServerComponentProps) => {
+  const propsString = JSON.stringify(componentProps);
+  const { rscPayloadGenerationUrl } = railsContext;
+  const strippedUrlPath = rscPayloadGenerationUrl?.replace(/^\/|\/$/g, '');
```
alexeyr-ci2: Does this strip `/` at the beginning and at the end? Wouldn't `rscPayloadGenerationUrl` normally start with a scheme, not with `/`, if it's an URL?

node_package/src/getReactServerComponent.client.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Instead, use the useRSC hook which provides getComponent and getCachedComponent functions
+ * for fetching or retrieving cached server components. For rendering server components,
+ * consider using RSCRoute component which handles the rendering logic automatically.
+ */
+const getReactServerComponent = ({
+  componentName,
+  componentProps,
+  railsContext,
+}: ClientGetReactServerComponentProps) => {
+  const componentKey = `${componentName}-${JSON.stringify(componentProps)}-${railsContext.componentSpecificMetadata?.renderRequestId}`;
```
alexeyr-ci2: Please reuse https://github.com/shakacode/react_on_rails/blob/8172e076039c405e654be44d5926e1cb7ae4e6fd/node_package/src/RSCProvider.tsx#L42-L44

node_package/src/injectRSCPayload.ts
```diff
@@ -0,0 +1,144 @@
...
+  let timeout: NodeJS.Timeout | null = null;
+  const resultStream = new PassThrough();
+
+  const startRSC = async () => {
+    try {
+      const rscPromises: Promise<void>[] = [];
+
+      ReactOnRails.onRSCPayloadGenerated?.(railsContext, (streamInfo) => {
+        const { stream, props, componentName } = streamInfo;
+        const cacheKey = `${componentName}-${JSON.stringify(props)}-${railsContext.componentSpecificMetadata?.renderRequestId}`;
```
alexeyr-ci2: Please reuse https://github.com/shakacode/react_on_rails/blob/8172e076039c405e654be44d5926e1cb7ae4e6fd/node_package/src/RSCProvider.tsx#L42-L44

node_package/src/injectRSCPayload.ts
```diff
@@ -0,0 +1,144 @@
...
+        );
+      });
+
+      await new Promise((resolve) => {
+        if (htmlStream.readableEnded) {
+          resolve(Promise.all(rscPromises));
+        } else {
+          htmlStream.on('end', () => {
+            resolve(Promise.all(rscPromises));
+          });
```
alexeyr-ci2: Can we use Promise-returning `finished` to simplify this https://nodejs.org/api/stream.html#streamfinishedstream-options? Something like
```
if (htmlStream.readableEnded) {
  await Promise.all(rscPromises);
} else {
  await finished(htmlStream); // { readable: true } if we want to wait for stream to be unreadable only
}
```
Or will this perform worse?

node_package/src/registerServerComponent/server.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * 1. Takes server component implementations
+ * 2. Wraps each component with RSCRoute using WrapServerComponentRenderer
+ * 3. Registers the wrapped components with ReactOnRails
+ *
+ * @param components - Object mapping component names to their implementations
+ *
+ * @example
+ * ```js
+ * registerServerComponent({
+ *   ServerComponent1: ServerComponent1Component,
```
alexeyr-ci2, line 20: Any reason the name would have `Component` twice? Does it come from some naming convention I'm missing?

node_package/src/streamServerRenderedReactComponent.ts
```diff
@@ -111,29 +102,12 @@ export const transformRenderStreamChunksToResultObject = (renderState: StreamRen
...
-      if (consoleReplayChunk) {
-        this.push(`${consoleReplayChunk}\n`);
-      }
       callback();
     },
   });
 
-  let pipedStream: PipeableStream | null = null;
-  const pipeToTransform = (pipeableStream: PipeableStream) => {
+  let pipedStream: PipeableStream | NodeJS.ReadableStream | null = null;
```
alexeyr-ci2: Is it worth naming `PipeableStream | NodeJS.ReadableStream` as a separate type? It ends up being used 3 times.

node_package/src/streamServerRenderedReactComponent.ts
```diff
@@ -147,7 +121,9 @@ export const transformRenderStreamChunksToResultObject = (renderState: StreamRen
   const writeChunk = (chunk: string) => transformStream.write(chunk);
   const endStream = () => {
     transformStream.end();
-    pipedStream?.abort();
+    if (pipedStream && 'abort' in pipedStream) {
+      pipedStream.abort();
+    }
```
alexeyr-ci2, line 133: Could be `pipedStream?.abort?.()`, not sure which is clearer. I guess keep yours?

node_package/src/types/index.ts
```diff
@@ -396,6 +409,45 @@ export interface ReactOnRailsInternal extends ReactOnRails {
...
+
+  // These functions are intended for use in Node.js environments only. They should be used on the server side and excluded from client-side bundles to reduce bundle size.
+  /**
+   * Generates a ReadableStream for a given component's RSC payload.
+   * @param componentName - The name of the component.
+   * @param props - The properties to pass to the component.
+   * @param railsContext - The Rails context of the current rendering request.
+   * @returns A promise that resolves to a NodeJS.ReadableStream.
+   */
+  getRSCPayloadStream?: (
```
alexeyr-ci2, line 477: I guess in the contexts it's useful we can assume it's present and not need `?`.

node_package/src/wrapServerComponentRenderer/server.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const wrapper: RenderFunction = async (props, railsContext) => {
+    if (!railsContext) {
+      throw new Error('RSCClientRoot: No railsContext provided');
+    }
+
+    const Component = isRenderFunction(componentOrRenderFunction)
+      ? await componentOrRenderFunction(props, railsContext)
+      : componentOrRenderFunction;
```
alexeyr-ci2, line 37: Duplicates https://github.com/shakacode/react_on_rails/blob/8172e076039c405e654be44d5926e1cb7ae4e6fd/node_package/src/wrapServerComponentRenderer/client.tsx#L36-L38

Not sure it's worth extracting, though.

node_package/src/wrapServerComponentRenderer/client.tsx
```diff
@@ -0,0 +1,82 @@
...
+      getServerComponent: getReactServerComponent,
+    });
+
+    const SuspensableRSCRoute = (
+      <React.Suspense fallback={null}>
+        <Component {...props} />
+      </React.Suspense>
+    );
+
+    const root = <RSCProvider>{SuspensableRSCRoute}</RSCProvider>;
```
alexeyr-ci2: Why not inline it? 
```suggestion
    const root = (
      <RSCProvider>
        <React.Suspense fallback={null}>
          <Component {...props} />
        </React.Suspense>
      </RSCProvider>
    );
```

But if you don't want to: 
1. it should be `Suspendable` (there is also [archaic/rare](https://en.wiktionary.org/wiki/suspensible) `Suspensible`).
2. I am not sure how `SuspensableRSCRoute` is related to `RSCRoute` itself.

node_package/src/wrapServerComponentRenderer/server.tsx
```diff
@@ -0,0 +1,63 @@
...
+      getServerComponent: getReactServerComponent,
+    });
+
+    const suspensableServerComponent = (
+      <React.Suspense fallback={null}>
+        <Component {...props} />
+      </React.Suspense>
+    );
+
+    const root = <RSCProvider>{suspensableServerComponent}</RSCProvider>;
```
alexeyr-ci2: Same as the above `SuspensableRSCRoute`. Except it may make sense to inline directly into the returned function:
```
return () => (
  <RSCProvider>
      <React.Suspense fallback={null}>
        <Component {...props} />
      </React.Suspense>
  </RSCProvider>
);
```

node_package/src/wrapServerComponentRenderer/client.tsx
alexeyr-ci2: Also maybe move checks which don't need `root` 
```
    if (!domNodeId) {
      throw new Error('RSCClientRoot: No domNodeId provided');
    }
    const domNode = document.getElementById(domNodeId);
    if (!domNode) {
      throw new Error(`RSCClientRoot: No DOM node found for id: ${domNodeId}`);
    }
```
above it?

node_package/tests/SuspenseHydration.test.tsx
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * This test verifies that async components are only hydrated on the client side after:
+ * 1. The component is rendered on the server
+ * 2. The rendered HTML is streamed to the client
+ *
+ * This behavior is critical for RSCRoute components, as they require their server-side
+ * RSC payload to be present in the page before client-side hydration can occur.
+ * And because the RSCRoute is rendered on the server because it's being hydrated on the client,
+ * we can sure that the RSC payload is present in the page before the RSCRoute is hydrated on the client.
+ * That's because `getRSCPayloadStream` function embeds the RSC payload immediately to the html stream even before the RSCRoute is rendered on the server.
```
alexeyr-ci2:
```suggestion
 * That's because `getRSCPayloadStream` function embeds the RSC payload immediately to the HTML stream even before the RSCRoute is rendered on the server.
```

node_package/tests/SuspenseHydration.test.tsx
```diff
@@ -0,0 +1,210 @@
...
+    onContainerHydrated,
+    onAsyncComponentHydrated,
+    writeFirstChunk,
+    writeSecondChunk,
+    hydrate,
+  };
+}
+
+// React Server Components tests are compatible with React 19
+// That only run with node version 18 and above
```
alexeyr-ci2:
```suggestion
// React Server Components tests require React 19 and only run with Node version 18 (`newest` in our CI matrix)
```

node_package/tests/injectRSCPayload.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  entries.forEach(([delay, value], index) => {
+    setTimeout(() => {
+      const chunksArray = Array.isArray(value) ? value : [value];
+      chunksArray.forEach((chunk) => {
+        passThrough.push(new TextEncoder().encode(chunk));
+      });
+      if (index === keysLength - 1) {
+        passThrough.push(null);
+      }
+    }, +delay);
```
alexeyr-ci2, line 24: Why is `+` needed here? Can `delay` be something other than a number?

node_package/tests/registerServerComponent.client.test.jsx
```diff
@@ -0,0 +1,253 @@
...
+      await act(async () => {
+        window.REACT_ON_RAILS_RSC_PAYLOADS['TestComponent-{}-test-render-request-id'].push(`${chunk2}\n`);
+
+        // Set document.readyState to 'complete' and dispatch readystatechange event
+        Object.defineProperty(document, 'readyState', { value: 'complete', writable: true });
+        document.dispatchEvent(new Event('readystatechange'));
+        document.dispatchEvent(new Event('DOMContentLoaded'));
+      });
+
+      // After second chunk, AsyncComponent should now be visible and loading indicator gone
```
alexeyr-ci2:
```suggestion
      // After the second chunk, AsyncComponent should now be visible and loading indicator gone
```

lib/react_on_rails/react_component/render_options.rb
Judahmeek: Since ReactOnRails Pro contains all the related caching functionality, it should also handle any necessary cache busting key generation.

That said, ReactOnRails Pro already has a request digest: https://github.com/shakacode/react_on_rails_pro/pull/352

Judahmeek (approved):
LGTM once Alexey's review comments are addressed.

eslint.config.ts
AbanoubGhadban: reverted

knip.ts
AbanoubGhadban: It is already intended as an entry point and can be imported directly from the package, so I believe it's acceptable to keep it as an entry here as well.

lib/react_on_rails/react_component/render_options.rb
AbanoubGhadban: @alexeyr-ci2 Could you take a look at the [RSCPayloadGenerator.ts](https://github.com/shakacode/react_on_rails/blob/e1949c4e26d897e5f6868a6c82a837f7c63623ec/node_package/src/RSCPayloadGenerator.ts#L91)(https://github.com/shakacode/react_on_rails/blob/e1949c4e26d897e5f6868a6c82a837f7c63623ec/node_package/src/RSCPayloadGenerator.ts#L91) file? It uses a global variable—a `Map` that associates each render request ID with an array of RSC payload streams. Since this global state is shared across all rendering requests within the same VM (we don’t create a separate VM per request), it's crucial that each render request has a unique ID to prevent data leakage. This approach avoids the risks of simply storing streams in a flat array, which would lead to cross-request contamination when ReactOnRails embeds the stream content into the HTML page.

AbanoubGhadban: @Judahmeek The `request digest` is currently an MD5 hash of the rendering request code. However, this can result in identical digests for different requests if they render the same component with the same props. This violates the uniqueness requirement I mentioned earlier, as we need each rendering request to have a truly unique identifier—even if the input data is the same.

node_package/src/ClientSideRenderer.ts
AbanoubGhadban: No, it means there is a bug in ReactOnRails

lib/react_on_rails/react_component/render_options.rb
alexeyr-ci2: @AbanoubGhadban This seems to fall under 

> Or is it needed for RSCs in general and just missed in the previous PRs?

right? So the question about avoiding it for non-RSC requests remains.

node_package/src/RSCProvider.tsx
AbanoubGhadban: Doing so, causes
```
Expected non-Promise value in a boolean conditional.
```
error

node_package/src/RSCProvider.tsx
AbanoubGhadban: I fixed the ESLint error without using `useMemo`. The context value is created at the parent function, which is not a React component. So, it's created once anyway.

node_package/src/RSCProvider.tsx
alexeyr-ci2: Then `if (!!existingPromise)` or `if (Boolean(existingPromise))` should work.

node_package/src/getReactServerComponent.client.ts
AbanoubGhadban: The `rscPayloadGenerationUrl` name is not accurate. I updated it to be `rscPayloadGenerationUrlPath`. It should contain the URL path. It can be considered as a relative URL.

node_package/src/injectRSCPayload.ts
AbanoubGhadban: Good catch. Updated it to `await finished(htmlStream).then(() => Promise.all(rscPromises));`.

node_package/src/registerServerComponent/server.tsx
AbanoubGhadban: No, it's by mistake. Updated it

node_package/src/streamServerRenderedReactComponent.ts
AbanoubGhadban: Honestly, I can't find a name for that type; even ChatGPT didn't provide good suggestions. 😊

node_package/src/streamServerRenderedReactComponent.ts
AbanoubGhadban: TS doesn't allow it

node_package/src/types/index.ts
AbanoubGhadban: It's not needed on client side. That's why I don't set these functions client version of ReactOnRails to reduce the bundle size.

node_package/src/wrapServerComponentRenderer/server.tsx
AbanoubGhadban: I don't think it worth creating a new utils function

node_package/src/wrapServerComponentRenderer/client.tsx
AbanoubGhadban: Good suggestions. I applied them all.

node_package/tests/injectRSCPayload.test.ts
AbanoubGhadban: `delay` is an object key, and object keys can either be a string or a symbol. When used as object keys, numbers are converted to strings.

lib/react_on_rails/configuration.rb
AbanoubGhadban: It's OK, both PRs will be merged together before being merged into master

node_package/src/streamServerRenderedReactComponent.ts
alexeyr-ci2: I mean, the trivial solution is `PipeableOrReadableStream` :)

node_package/src/streamServerRenderedReactComponent.ts
alexeyr-ci2: OK.

lib/react_on_rails/react_component/render_options.rb
AbanoubGhadban: It's not possible to determine in advance whether a request involves rendering a server component. Therefore, I updated the code to generate a rendering request ID only when RSC support is enabled (`ReactOnRailsPro.configuration.enable_rsc_support == true`).

lib/react_on_rails/utils.rb
```diff
@@ -208,6 +213,13 @@ def self.react_on_rails_pro_version
                                     end
     end
 
+    def self.rsc_support_enabled?
+      return false unless react_on_rails_pro?
+
+      rorp_config = ReactOnRailsPro.configuration
+      rorp_config.respond_to?(:enable_rsc_support) && rorp_config.enable_rsc_support
```
alexeyr-ci2: Maybe worth caching (especially with `respond_to?` involved)? Something like
```suggestion
      return false unless react_on_rails_pro?
      return @rsc_support_enabled if defined?(@rsc_support_enabled)

      rorp_config = ReactOnRailsPro.configuration
      @rsc_support_enabled = rorp_config.respond_to?(:enable_rsc_support) && rorp_config.enable_rsc_support
```

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,130 @@
...
+
+declare global {
+  function generateRSCPayload(
+    componentName: string,
+    props: unknown,
+    railsContext: RailsContextWithServerComponentCapabilities,
+  ): Promise<NodeJS.ReadableStream>;
+}
+
+const mapRailsContextToRSCPayloadStreams = new Map<string, RSCPayloadStreamInfo[]>();
```
alexeyr-ci2: Is `rscPayloadStreams` a better name (for consistency with `rscPayloadCallbacks` below)? At least it maps `renderRequestId` to streams, not `railsContext`, and `map` should be at the end of the name.

Would it make sense for it (and these functions) to take `renderRequestId` directly instead of `railsContext`?

node_package/src/RSCPayloadGenerator.ts
```diff
@@ -0,0 +1,130 @@
+import { PassThrough } from 'stream';
+import {
+  RailsContextWithServerComponentCapabilities,
+  RSCPayloadStreamInfo,
+  RSCPayloadCallback,
+} from './types/index.ts';
+
+declare global {
+  function generateRSCPayload(
```
alexeyr-ci2, line 9: Where does this function come from? I've checked the corresponding RORP PR https://github.com/shakacode/react_on_rails_pro/pull/533 but don't see it there.

node_package/src/injectRSCPayload.ts
```diff
@@ -0,0 +1,136 @@
...
+        rscPromises.push(
+          new Promise((resolve, reject) => {
+            (async () => {
+              for await (const chunk of stream ?? []) {
+                const decodedChunk = typeof chunk === 'string' ? chunk : decoder.decode(chunk);
+                writeChunk(JSON.stringify(decodedChunk), resultStream, cacheKey);
+              }
+              resolve();
+            })().catch(reject);
+          }),
```
alexeyr-ci2:
```suggestion
          (async () => {
            try {
              for await (const chunk of stream ?? []) {
                const decodedChunk = typeof chunk === 'string' ? chunk : decoder.decode(chunk);
                writeChunk(JSON.stringify(decodedChunk), resultStream, cacheKey);
              }
            } catch (err) {
              return Promise.reject(err);
            }
          })();
```
may be a simpler alternative. Or even remove `try`/`catch`, I _think_ that's still the same.

node_package/src/postSSRHooks.ts
```diff
@@ -0,0 +1,20 @@
...
+type PostSSRHook = () => void;
+const postSSRHooks = new Map<string, PostSSRHook[]>();
+
+export const addPostSSRHook = (
+  railsContext: RailsContextWithServerComponentCapabilities,
+  hook: PostSSRHook,
+) => {
+  const hooks = postSSRHooks.get(railsContext.componentSpecificMetadata.renderRequestId) || [];
+  hooks.push(hook);
+  postSSRHooks.set(railsContext.componentSpecificMetadata.renderRequestId, hooks);
```
alexeyr-ci2: Should be similar to https://github.com/shakacode/react_on_rails/blob/8520d4efb46d78b226a0e44f0518c75fc7a2a2c8/node_package/src/RSCPayloadGenerator.ts#L98-L103

script/convert
```diff
@@ -26,12 +26,15 @@ gsub_file_content("../package.json", /"eslint-plugin-jest": "[^"]*",/, "")
 gsub_file_content("../package.json", /"globals": "[^"]*",/, "")
 gsub_file_content("../package.json", /"knip": "[^"]*",/, "")
 gsub_file_content("../package.json", /"publint": "[^"]*",/, "")
-gsub_file_content("../package.json", /"typescript-eslint": "[^"]*",/, "")
+gsub_file_content("../package.json", /"typescript-eslint": "[^"]*",?/, "")
```
alexeyr-ci2, line 29: Does this actually work? It should end up with a trailing comma in the previous dependency, that's why I didn't do it this way in the first place.

It may be better to load the file as JSON, manipulate it in Ruby and save the JSON.

node_package/src/RSCPayloadGenerator.ts
AbanoubGhadban: Done. However, I prefer to keep `railsContext` as a parameter for better encapsulation. These functions might require additional information from the Rails side in the future, which can be conveniently passed through the context.

node_package/src/RSCPayloadGenerator.ts
AbanoubGhadban: It's at https://github.com/shakacode/react_on_rails_pro/pull/515 PR. Specifically at https://github.com/shakacode/react_on_rails_pro/pull/515/files#diff-99d134e7721ea40ab59932a16dd1dde5188ec6f4e5e582d32fb5c3d0bbb7140dR39

script/convert
AbanoubGhadban: I already added a code that cleans the trailing commas:
https://github.com/shakacode/react_on_rails/pull/1736/files#diff-59fdcbe769964bbf8a369d03094edaaf4dfd832e68a2bebab164e788aff94ed9R36

node_package/src/ReactOnRailsRSC.ts
```diff
@@ -38,9 +46,15 @@ const streamRenderRSCComponent = (
...
-      const rscStream = renderToPipeableStream(reactElement, reactClientManifest, {
+  Promise.resolve(reactRenderingResult)
+    .then(async (reactElement) => {
+      if (!serverRenderer) {
+        const reactClientManifest = await loadJsonFile<BundleManifest>(reactClientManifestFileName);
+        serverRenderer = buildServerRenderer(reactClientManifest);
+      }
+
+      const { renderToPipeableStream } = serverRenderer;
+      const rscStream = renderToPipeableStream(reactElement, {
```
alexeyr-ci2: I think `Promise.resolve(reactRenderingResult).then(async (reactElement) => ...)` can be simplified; just use
```
async () => { /* replace reactElement with reactRenderingResult here */ }
```
and `.catch` with `try`/`catch`.

node_package/src/ReactOnRailsRSC.ts
```diff
@@ -61,6 +75,10 @@ const streamRenderRSCComponent = (
       const jsonResult = JSON.stringify(createResultObject(htmlResult, buildConsoleReplay(), renderState));
       return stringToStream(jsonResult);
```
alexeyr-ci2: I don't see how the last 3 lines in `catch` https://github.com/shakacode/react_on_rails/blob/11e4a54cbe581aa6d8d5291b6bddd54683437b95/node_package/src/ReactOnRailsRSC.ts#L74-L76 are ever used. They affect the `Promise.resolve(...).catch(...)` value, sure, but that value itself is ignored now, right?

node_package/src/RSCPayloadGenerator.ts
alexeyr-ci2: Maybe add a comment explaining this (with a link once that PR is merged)?

node_package/src/getReactServerComponent.client.ts
```diff
@@ -0,0 +1,140 @@
...
+      // eslint-disable-next-line no-param-reassign
+      payloads.push = (...chunks) => {
+        chunks.forEach(handleChunk);
+        return chunks.length;
+      };
+      streamController = controller;
+    },
+  });
+
+  if (typeof document !== 'undefined' && document.readyState === 'loading') {
```
alexeyr-ci2, line 72: Should the listener be added if the state is `interactive` too, since `DOMContentLoaded` hasn't fired in that case yet? I expect not, but asking just in case.

node_package/src/types/index.ts
```diff
@@ -53,6 +63,52 @@ export type RailsContext = {
...
+    );
+  }
+};
+
+export const assertRailsContextWithServerComponentCapabilities: (
+  context: RailsContext | undefined,
+) => asserts context is RailsContextWithServerComponentCapabilities = (
+  context: RailsContext | undefined,
+): asserts context is RailsContextWithServerComponentCapabilities => {
+  if (
```
alexeyr-ci2, line 96: Should we do this check in production, or only in development?

node_package/src/injectRSCPayload.ts
alexeyr-ci2:
```suggestion
        const cacheKey = createRSCPayloadKey(componentName, props, railsContext)`;
```
(except import needs to be added as well).

node_package/src/utils.ts
```diff
@@ -8,5 +10,28 @@ const customFetch = (...args: Parameters<typeof fetch>) => {
...
+  return `${componentName}-${JSON.stringify(componentProps)}-${railsContext.componentSpecificMetadata.renderRequestId}`;
+};
+
+/**
+ * Wraps a promise from react-server-dom-webpack in a standard JavaScript Promise.
+ *
+ * This is necessary because promises returned by react-server-dom-webpack's methods
+ * (like `createFromReadableStream` and `createFromNodeStream`) have non-standard behavior:
+ * their `then()` method returns `null` instead of the promise itself, which breaks
+ * promise chaining. This wrapper creates a new standard Promise that properly
```
alexeyr-ci2, line 29: Is that behavior documented somewhere (including issues)? If so, please link it.

node_package/tests/SuspenseHydration.test.tsx
```diff
@@ -0,0 +1,209 @@
+import * as React from 'react';
+import { Suspense } from 'react';
+import { renderToReadableStream } from 'react-dom/server';
+import { hydrateRoot } from 'react-dom/client';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore - TypeScript error can be ignored because:
```
alexeyr-ci2, line 6: It doesn't really matter, but why is it getting type-checked under `oldest` in the first place?

node_package/src/ReactOnRailsRSC.ts
AbanoubGhadban: I defined an inner function within the `streamRenderRSCComponent` function. I don't prefer Immediately Invoked Function Expressions (IIFE).

node_package/src/ReactOnRailsRSC.ts
AbanoubGhadban: I don't understand what you mean. The `catch` block should catch errors that occur inside the returned promise from the rendering request, as well as any errors that may happen during the execution of the function passed to the `then` method.

node_package/src/getReactServerComponent.client.ts
AbanoubGhadban: There is no need to wait. If the state is `interactive`, it is safe to close the stream immediately. In the `interactive` state, all embedded scripts in the HTML page should have already been parsed, and no additional chunks will be added to the stream.

node_package/src/types/index.ts
AbanoubGhadban: If the assertion fails, it indicates that the user is either using an incompatible version of RORP or there is a bug in RORP. In both situations, failing to throw an error here results in an unknown or ambiguous error occurring in the calling function. Therefore, it is advisable to explicitly fail at this point and provide a descriptive error message.

node_package/src/utils.ts
AbanoubGhadban: No, the package is already under development, and it's okay to have some bugs. You can test these functions yourself to verify that the bug exists. I will submit an issue about it later.

node_package/tests/SuspenseHydration.test.tsx
AbanoubGhadban: Jest fails when it faces TS error

node_package/src/ReactOnRailsRSC.ts
alexeyr-ci2: Then what about just `Promise.resolve().then(async () => { ... })`?

node_package/src/ReactOnRailsRSC.ts
alexeyr-ci2: Yes, of course. The first lines of the `catch` block have an effect by modifying `renderState`:
```
      const error = convertToError(e);
      renderState.hasErrors = true;
      renderState.error = error;
```
and actually `handleError` itself has some side effects from logging to the console. But I can't see what changes if you remove the last two lines (`const jsonResult = ...; return stringToStream(jsonResult);`) or replace them with just `return ''`.

AbanoubGhadban: Great observation! I’m not sure how I overlooked it. I have updated it to send the error HTML to the stream.

alexeyr-ci2 (approved):

PR #1742, opened 2025-07-02
Fix invalid version checker warnings and add pre-release comparison

alexeyr-ci2:
### Summary

There were incorrect warnings when using RC versions with Popmenu:
```
**WARNING** ReactOnRails: Your node package version for react-on-rails contains a ^ or ~
Detected: 15.0.0-rc.1
     gem: 15.0.0.rc.1
Ensure the installed version of the gem is the same as the version of
your installed node package. Do not use >= or ~> in your Gemfile for react_on_rails.
Do not use ^ or ~ in your package.json for react-on-rails.
```
This PR fixes them.

Also adds warnings when e.g. using `15.0.0-rc.0` Node package with `15.0.0.rc.1` gem, since those may have breaking changes between them.

### Pull Request checklist

- ~[ ] Update documentation~
- [x] Update CHANGELOG file

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1742)

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved incorrect warnings about version mismatches when using pre-release versions of React on Rails.
  * Improved detection and messaging for non-exact version specifications in configuration files.
  * Enhanced version parsing and comparison accuracy for better compatibility checks.
  * Updated compatibility checks and error messages related to Shakapacker version requirements.

* **Documentation**
  * Updated changelog to document the fix for version mismatch warnings and improved guidance on version specification requirements.

### DIFF
--- a/lib/react_on_rails/packer_utils.rb
+++ b/lib/react_on_rails/packer_utils.rb
@@ -57,6 +57,8 @@ module ReactOnRails
     end
 
     def self.shakapacker_version_as_array
+      return @shakapacker_version_as_array if defined?(@shakapacker_version_as_array)
+
       match = shakapacker_version.match(ReactOnRails::VersionChecker::VERSION_PARTS_REGEX)
 
       # match[4] is the pre-release version, not normally a number but something like "beta.1" or `nil`
--- a/lib/react_on_rails/packer_utils.rb
+++ b/lib/react_on_rails/packer_utils.rb
@@ -57,9 +57,7 @@ module ReactOnRails
     end
 
     def self.shakapacker_version_as_array
-      match = shakapacker_version.match(ReactOnRails::VersionChecker::MAJOR_MINOR_PATCH_VERSION_REGEX)
-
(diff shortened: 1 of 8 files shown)


lib/react_on_rails/packer_utils.rb
```diff
@@ -57,9 +57,7 @@ def self.shakapacker_version
     end
 
     def self.shakapacker_version_as_array
-      match = shakapacker_version.match(ReactOnRails::VersionChecker::MAJOR_MINOR_PATCH_VERSION_REGEX)
-
-      @shakapacker_version_as_array = [match[1].to_i, match[2].to_i, match[3].to_i]
+      Gem::Version.new(shakapacker_version).segments
```
AbanoubGhadban: Why don’t we use the `VERSION_PARTS_REGEX` to extract the parts instead? `Gem::Version#segments` can return different results from the regex parts in certain cases.

AbanoubGhadban: Also, please check the `semver_to_string` function at `lib/react_on_rails/packer_utils.rb` and check if it needs updating.

PR #1745, opened 2025-07-07
Get rid of `renderRequestId` and use local trackers instead of global tracking

AbanoubGhadban:
### Problem

When React Server Components (RSC) support is enabled, components are not being cached effectively due to a fundamental issue with how RSC payloads are tracked. The root cause is that `renderRequestId` changes with every new rendering request, which causes cache keys to change on each request, preventing proper component caching.

The initial approach of excluding `renderRequestId` from cache key computation revealed a deeper architectural issue: cached content still contained old `renderRequestId` values, breaking the expectation that this ID should be unique per request.

### Solution

This PR implements a comprehensive refactoring that eliminates the dependency on `renderRequestId` for RSC payload tracking by introducing **local request-scoped tracking**:

#### Key Changes

1. **Eliminated Global State Management**
   - Removed global Maps that tracked RSC payloads and post-SSR hooks by `renderRequestId`
   - Eliminated TTL-based cleanup mechanisms that were prone to memory leaks

2. **Introduced Local Request Trackers**
   - `PostSSRHookTracker`: Manages post-SSR hooks for a single request
   - `RSCRequestTracker`: Handles RSC payload generation and stream tracking per request
   - Both trackers are automatically garbage collected when requests complete

3. **Updated Cache Key Strategy**
   - Cache keys now use `domNodeId` instead of `renderRequestId`
   - Provides better collision prevention for multiple component instances
   - Aligns client-side and server-side caching mechanisms

4. **Removed `renderRequestId` Dependencies**
   - Removed `renderRequestId` generation from Ruby `RenderOptions` class
   - Eliminated `data-render-request-id` attribute from component specification tags
   - Updated TypeScript types to remove `componentSpecificMetadata` requirements

5. **Enhanced Error Handling**
   - Added comprehensive error handling in RSC payload generation
   - Improved stream cleanup with graceful error management
   - Better validation for missing configuration parameters

### Benefits

- **Enables Proper Caching**: Components can now be cached effectively since cache keys remain stable
- **Eliminates Memory Leaks**: No more stale global state that could accumulate over time
- **Prevents Race Conditions**: Each request operates in complete isolation
- **Improves Debugging**: Request-scoped state makes issues easier to trace and reproduce
- **Better Testability**: No shared global state to reset between tests
- **Predictable Cleanup**: Automatic resource management through garbage collection

### Architecture

The new architecture implements request-scoped tracking where:
- Each request gets its own `PostSSRHookTracker` and `RSCRequestTracker` instances
- Trackers inject their capabilities into the Rails context for component access
- State is automatically cleaned up when requests complete
- No shared state exists between concurrent requests

### Testing

Tests have been updated to reflect the new architecture, including:
- Updated mock implementations for the new tracker-based approach
- Increased timeout values where needed for improved reliability
- Modified cache key expectations to use `domNodeId` instead of `renderRequestId`

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file

### Other Information

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1745)

## Summary by CodeRabbit

* **Refactor**
  * Streamlined server-side rendering and React Server Component (RSC) payload management by introducing per-request trackers for hooks and payloads, replacing global state and deprecated context metadata.
  * Unified and simplified Rails context types for server streaming and component metadata, removing obsolete properties and methods.
  * Updated APIs and internal logic to use new per-request trackers, improving isolation and reliability of SSR and RSC workflows.
  * Enhanced error handling and payload key generation for React Server Components to avoid collisions and ensure proper caching.
  * Simplified RSC provider setup by requiring explicit payload key generators and removing redundant context assertions.

* **Breaking Changes**
  * Removed several deprecated properties and methods from public interfaces and context types.
  * Updated function signatures and required parameters for RSC and SSR-related utilities.
  * Eliminated usage of render request IDs in payload keys and context metadata, affecting cache key formats and component payload tracking.

* **Bug Fixes**
  * Improved cleanup and execution of post-SSR hooks and RSC payload streams to prevent memory leaks and race conditions.


node_package/src/PostSSRHookTracker.ts
```diff
@@ -0,0 +1,39 @@
...
+  notifySSREnd(): void {
+    this.hooks.forEach((hook) => hook());
+    this.hooks = [];
+  }
+
+  /**
+   * Clears all hooks without executing them.
+   * Should be called if the request is aborted or cleanup is needed.
+   */
+  clear(): void {
```
alexeyr-ci2: Does this ever get called? Searching for `clear()`, I don't find such a case.

node_package/src/RSCRequestTracker.ts
```diff
@@ -0,0 +1,124 @@
+import { PassThrough, Readable } from 'stream';
+import {
+  RSCPayloadStreamInfo,
+  RSCPayloadCallback,
+  RailsContextWithServerComponentMetadata,
+} from './types/index.ts';
+
+declare global {
+  function generateRSCPayload(
```
alexeyr-ci2, line 19: Please include a comment saying where it's actually defined (I think I asked for this last time as well).

node_package/src/ReactOnRails.node.ts
```diff
@@ -1,19 +1,7 @@
...
-  onRSCPayloadGenerated,
-} from './RSCPayloadGenerator.ts';
-import { addPostSSRHook } from './postSSRHooks.ts';
 
 ReactOnRails.streamServerRenderedReactComponent = streamServerRenderedReactComponent;
-ReactOnRails.getRSCPayloadStream = getRSCPayloadStream;
-ReactOnRails.getRSCPayloadStreams = getRSCPayloadStreams;
-ReactOnRails.clearRSCPayloadStreams = clearRSCPayloadStreams;
-ReactOnRails.onRSCPayloadGenerated = onRSCPayloadGenerated;
-ReactOnRails.addPostSSRHook = addPostSSRHook;
```
alexeyr-ci2, line 16: This API was also added after 14.2.1 and so removing it isn't a breaking change, right?

AbanoubGhadban:
@coderabbitai review

node_package/src/utils.ts
```diff
@@ -15,9 +13,15 @@ export { customFetch as fetch };
 export const createRSCPayloadKey = (
   componentName: string,
   componentProps: unknown,
-  railsContext: RailsContextWithComponentSpecificMetadata,
+  domNodeId: string | undefined,
 ) => {
-  return `${componentName}-${JSON.stringify(componentProps)}-${railsContext.componentSpecificMetadata.renderRequestId}`;
+  if (!domNodeId) {
```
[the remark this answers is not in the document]
AbanoubGhadban: I updated the server implementation of the `wrapServerComponentRenderer` function to use the string 'server' as the DOM ID instead of an empty string.

node_package/src/RSCRequestTracker.ts
```diff
@@ -0,0 +1,124 @@
...
+    componentName: string,
+    props: unknown,
+    railsContext: RailsContextWithServerComponentMetadata,
+  ): Promise<NodeJS.ReadableStream>;
+}
```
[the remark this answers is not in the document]
AbanoubGhadban: The `generateRSCPayload` function is globally defined by RORP rendering request.

node_package/src/getReactServerComponent.client.ts
```diff
@@ -41,13 +42,39 @@ const createFromFetch = async (fetchPromise: Promise<Response>) => {
...
+  try {
+    const propsString = JSON.stringify(componentProps);
+    const strippedUrlPath = rscPayloadGenerationUrlPath.replace(/^\/|\/$/g, '');
+    const encodedParams = new URLSearchParams({ props: propsString }).toString();
+    const fetchUrl = `/${strippedUrlPath}/${componentName}?${encodedParams}`;
+
+    return createFromFetch(fetch(fetchUrl)).catch((error: unknown) => {
+      throw new Error(
+        `Failed to fetch RSC payload for component "${componentName}" from "${fetchUrl}": ${
+          error instanceof Error ? error.message : String(error)
```
alexeyr-ci2: Do we already have a function for this?

node_package/src/getReactServerComponent.client.ts
```diff
@@ -41,13 +42,39 @@ const createFromFetch = async (fetchPromise: Promise<Response>) => {
  *
  * @param props - Object containing component name, props, and railsContext
```
alexeyr-ci2: `props - Object containing ... props` is potentially confusing. Maybe "component name and props" is better?

node_package/src/getReactServerComponent.client.ts
```diff
@@ -131,14 +158,14 @@ const getReactServerComponent = ({
...
 }: ClientGetReactServerComponentProps) => {
-  assertRailsContextWithComponentSpecificMetadata(railsContext);
-  const componentKey = createRSCPayloadKey(componentName, componentProps, railsContext);
+  const componentKey = createRSCPayloadKey(componentName, componentProps);
   const payloads = window.REACT_ON_RAILS_RSC_PAYLOADS?.[componentKey];
   if (!enforceRefetch && payloads) {
     return createFromPreloadedPayloads(payloads);
   }
-  return fetchRSC({ componentName, componentProps, railsContext });
+  return fetchRSC({ componentName, componentProps, railsContext, createRSCPayloadKey });
```
alexeyr-ci2: `fetchRSC` doesn't currently use `createRSCPayloadKey` https://github.com/shakacode/react_on_rails/blob/521c5b3d3b1e99d483eb5cb0bbf34c50828743d1/node_package/src/getReactServerComponent.client.ts#L47 

If it needs to be changed, can we just pass `componentKey` instead of `createRSCPayloadKey`?

node_package/src/RSCProvider.tsx
```diff
@@ -33,30 +32,33 @@ const RSCContext = React.createContext<RSCContextType | undefined>(undefined);
 export const createRSCProvider = ({
   railsContext,
   getServerComponent,
+  createRSCPayloadKey,
```
alexeyr-ci2: It looks like the third parameter can be `domNodeId` instead and lead to simpler code.

node_package/src/utils.ts
```diff
@@ -12,12 +10,39 @@ const customFetch = (...args: Parameters<typeof fetch>) => {
...
+ * @param componentProps - Props passed to the component (serialized to JSON)
+ * @param domNodeId - DOM node ID of the parent react component rendered at the rails view (prevents instance collisions)
+ * @returns A unique cache key string
+ * @throws Error if domNodeId is missing, which could lead to cache collisions
+ */
 export const createRSCPayloadKey = (
   componentName: string,
   componentProps: unknown,
-  railsContext: RailsContextWithComponentSpecificMetadata,
+  domNodeId: string | undefined,
```
alexeyr-ci2: Why allow `undefined` if we fail on it immediately?

node_package/tests/streamServerRenderedReactComponent.test.jsx
```diff
@@ -10,7 +10,7 @@ import ReactOnRails from '../src/ReactOnRails.node.ts';
 
 const AsyncContent = async ({ throwAsyncError }) => {
   await new Promise((resolve) => {
-    setTimeout(resolve, 10);
+    setTimeout(resolve, 50);
```
alexeyr-ci2, line 13: Is this an intentional change?

node_package/src/PostSSRHookTracker.ts
AbanoubGhadban: removed

node_package/src/ReactOnRails.node.ts
AbanoubGhadban: Yes

node_package/tests/streamServerRenderedReactComponent.test.jsx
AbanoubGhadban: Yes, but it's not necessary for this PR. It's a flaky test, and I increased the delay to make it reliable.

node_package/src/RSCRequestTracker.ts
```diff
@@ -0,0 +1,161 @@
...
+      this.streams.push(streamInfo);
+
+      // Notify callbacks about the new stream in a sync manner to maintain proper hydration timing
+      this.callbacks.forEach((callback) => callback(streamInfo));
+
+      return stream1;
+    } catch (error) {
+      // Provide a more helpful error message that includes context
+      throw new Error(
+        `Failed to generate RSC payload for component "${componentName}": ${extractErrorMessage(error)}`,
```
alexeyr-ci2, line 145: Actually, better question than previous: why not
```suggestion
        `Failed to generate RSC payload for component "${componentName}"`,
        { cause: error },
```
? And if so, same for other `extractErrorMessage` uses.

I guess it requires Node 16.9, but if that's newer than our current minimum version, I'm OK with increasing it.

node_package/src/getReactServerComponent.client.ts
```diff
@@ -104,41 +131,48 @@ const createFromPreloadedPayloads = (payloads: string[]) => {
...
-  const payloads = window.REACT_ON_RAILS_RSC_PAYLOADS?.[componentKey];
-  if (!enforceRefetch && payloads) {
-    return createFromPreloadedPayloads(payloads);
-  }
-  return fetchRSC({ componentName, componentProps, railsContext });
-};
+const getReactServerComponent =
+  (domNodeId: string, railsContext: RailsContext) =>
+  ({ componentName, componentProps, enforceRefetch = false }: ClientGetReactServerComponentProps) => {
+    const componentCacheKey = createServerComponentCacheKey(componentName, componentProps);
```
alexeyr-ci2: Maybe `createServerComponentCacheKey` should still have an optional `domNodeId` parameter to avoid doing it manually here and below, but I'm not sure about the name then. Or make `createRSCPayloadKey` a separate function.

node_package/src/getReactServerComponent.client.ts
```diff
@@ -104,41 +131,48 @@ const createFromPreloadedPayloads = (payloads: string[]) => {
...
+
+    const rscPayloadKey = `${componentCacheKey}-${domNodeId}`;
+    const payloads = window.REACT_ON_RAILS_RSC_PAYLOADS?.[rscPayloadKey];
+    if (!enforceRefetch && payloads) {
+      return createFromPreloadedPayloads(payloads);
+    }
```
alexeyr-ci2:
```suggestion
    if (!enforceRefetch) {
      const componentCacheKey = createServerComponentCacheKey(componentName, componentProps);
      const rscPayloadKey = `${componentCacheKey}-${domNodeId}`;
      const payloads = window.REACT_ON_RAILS_RSC_PAYLOADS?.[rscPayloadKey];
      if (payloads) {
        return createFromPreloadedPayloads(payloads);
      }
    }
```
since `payloads` are only needed in that case. Or even `if (!enforceRefetch && window.REACT_ON_RAILS_RSC_PAYLOADS)` (and remove `?.`).

node_package/src/getReactServerComponent.server.ts
```diff
@@ -49,41 +54,28 @@ const createFromReactOnRailsNodeStream = async (
...
-  componentProps,
-  railsContext,
-}: GetReactServerComponentOnServerProps) => {
-  assertRailsContextWithServerComponentCapabilities(railsContext);
+const getReactServerComponent =
+  (railsContext: RailsContextWithServerStreamingCapabilities) =>
+  async ({ componentName, componentProps }: GetReactServerComponentOnServerProps) => {
+    const rscPayloadStream = await railsContext.getRSCPayloadStream(componentName, componentProps);
 
-  if (typeof ReactOnRails.getRSCPayloadStream !== 'function') {
```
alexeyr-ci2, line 70: Is this checked elsewhere now or just not needed?

node_package/src/RSCRequestTracker.ts
AbanoubGhadban: It requires updating `@tsconfig/node` to a higher version. Please do it in another PR.

node_package/src/getReactServerComponent.server.ts
AbanoubGhadban: This function is removed from the global `ReactOnRails` object and added to the `RailsContext` object. It's checked at the `assertRailsContextWithServerComponentMetadata` function.

AbanoubGhadban:
@Judahmeek caching is provided by RORP, so tests are added to
https://github.com/shakacode/react_on_rails_pro/pull/539

alexeyr-ci2 (approved):

PR #1749, opened 2025-07-12
Have the lint job its own cache keys

Judahmeek:
Currently, the lint job is trying to use the oldest spec/dummy's gem cache without running the conversion script. This causes every gem that would be removed/modified by the conversion script to instead be installed every time the job runs.

However, running the conversion script would cause eslint to be removed.

Therefore, we should switch to using the newest spec/dummy gem cache.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1749)

## Summary by CodeRabbit

* **Tests**
  * Improved test assertions to explicitly verify that registering with an empty object does not throw errors.
  * Enhanced test setup to conditionally mock console output based on an environment variable, allowing more flexible control over console logging during tests.

* **Chores**
  * Updated caching keys for Ruby gem dependencies in the continuous integration workflow.

### DIFF
--- a/.github/workflows/lint-js-and-ruby.yml
+++ b/.github/workflows/lint-js-and-ruby.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: vendor/bundle
-          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-newest
+          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-lint
       - name: Install Node modules with Yarn for renderer package
         run: |
           yarn install --no-progress --no-emoji --frozen-lockfile
@@ -60,7 +60,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: spec/dummy/vendor/bundle
-          key: dummy-app-gem-cache-${{ hashFiles('spec/dummy/Gemfile.lock') }}-newest
+          key: dummy-app-gem-cache-${{ hashFiles('spec/dummy/Gemfile.lock') }}-lint
       - name: Install Ruby Gems for dummy app
         run: |
           cd spec/dummy
--- a/.github/workflows/lint-js-and-ruby.yml
+++ b/.github/workflows/lint-js-and-ruby.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: vendor/bundle
-          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-oldest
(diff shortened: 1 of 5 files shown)


.github/workflows/lint-js-and-ruby.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: vendor/bundle
-          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-oldest
+          key: package-app-gem-cache-${{ hashFiles('Gemfile.lock') }}-lint
```
alexeyr-ci2, line 42: > Therefore, we should switch to using the newest spec/dummy gem cache.

Newest makes sense, but a separate `-lint` one doesn't to me.

node_package/tests/jest.setup.js
```diff
@@ -62,3 +62,13 @@ if (typeof window !== 'undefined') {
   global.ReadableStream = ReadableStream;
   global.ReadableStreamDefaultReader = ReadableStreamDefaultReader;
 }
+
+global.console.log('All calls to console have been disabled in jest.setup.js');
```
alexeyr-ci2: Why? It's useful to see logs when running tests. Error/warning messages from React and other libraries in particular.

At least make an environment variable to disable this.

alexeyr-ci2 (changes requested):

.github/workflows/lint-js-and-ruby.yml
Judahmeek: newest didn't work. That's why one of the commits is named ["newest cache is still borked"](https://github.com/shakacode/react_on_rails/pull/1749/commits/40e21f0e28ebda69f5c571bf68cbb094e82da1af). It results in the gems being reinstalled just like the oldest cache key did.

The newest cache key works for other jobs, like Rspec, so I have no idea what is wrong.

All I know is that the lint key fixes the problem.

node_package/tests/jest.setup.js
Judahmeek: The logs make it difficult to distinguish which test is failing and which console error is connected to the failure.

This change clears out all console output from the passing tests. Which makes seeing the output that actually matters a lot easier.

Happy to add an environment variable if you think it helps.

alexeyr-ci2: I don't see why it would only clear console output from passing tests. If it does leave it in place for failing tests, great.

.github/workflows/lint-js-and-ruby.yml
alexeyr-ci2: OK.

alexeyr-ci2 (approved):

PR #1838, opened 2025-09-29
Fix RSpec helper optimization gap with private SSR directories

### LINKED ISSUE #1822: Need to test out rspec helper with new private ssr files directory
Links:
https://github.com/search?q=repo%3Ashakacode%2Freact_on_rails%20manifest.json&type=code
https://github.com/shakacode/react_on_rails/blob/120bd432d4f88161f3228d7121612fab18cf195d/docs/guides/rspec-configuration.md?plain=1#L62

The place that we used "manifest_json" as a "bundle" was to save time checking for changes that trigger webpack builds for rspec.

ihabadham:
## Summary

Fixes an optimization gap in the RSpec helper where server bundle staleness was not detected when using private SSR directories (introduced in PR #1798).

- **Issue**: Default `webpack_generated_files: ['manifest.json']` configuration only monitored client assets, missing server bundle changes in private `ssr-generated/` directory
- **Impact**: Tests could run with stale server-side code, leading to incorrect test results
- **Solution**: Enhanced `ensure_webpack_generated_files_exists` to automatically include server bundles and other critical files in monitoring list

## Root Cause

PR #1798 moved server bundles to private directories for security, but the RSpec helper optimization continued monitoring only `manifest.json` by default. This created a gap where:

1. Server bundle changes in `ssr-generated/server-bundle.js` 
2. Client assets remain fresh in `public/packs-test/manifest.json`
3. RSpec helper reports "no rebuild needed" ❌
4. Tests run with stale server-side code

**Why this worked before:** Prior to PR #1798, both client and server bundles were co-located in the same `public/packs/` directory, so the RSpec helper's fallback logic would detect server bundle staleness even when only monitoring `manifest.json`.

## Fix Details

**Before:**
```ruby
def ensure_webpack_generated_files_exists
  return unless webpack_generated_files.empty?  # ❌ Never runs with default config
  # ...
end
```

**After:**
```ruby
def ensure_webpack_generated_files_exists
  all_required_files = [
    "manifest.json",
    server_bundle_js_file,
    rsc_bundle_js_file,
    react_client_manifest_file,
    react_server_client_manifest_file
  ].compact_blank

  if webpack_generated_files.empty?
    self.webpack_generated_files = all_required_files
  else
    missing_files = all_required_files.reject { |file| webpack_generated_files.include?(file) }
    self.webpack_generated_files += missing_files if missing_files.any?
  end
end
```

## Testing

- ✅ Comprehensive test coverage added for all scenarios
- ✅ Verified fix resolves gap with test apps
- ✅ Backward compatibility maintained
- ✅ Works across different directory configurations

## Test plan

- [x] RSpec helper correctly detects server bundle staleness with default config
- [x] Cross-directory monitoring works (client in `public/`, server in `ssr-generated/`)
- [x] Existing configurations continue working unchanged
- [x] Edge cases handled (nil values, custom configurations)

Fixes #1822

### DIFF
--- a/lib/react_on_rails/configuration.rb
+++ b/lib/react_on_rails/configuration.rb
@@ -320,15 +320,20 @@ module ReactOnRails
     end
 
     def ensure_webpack_generated_files_exists
-      return unless webpack_generated_files.empty?
-
-      self.webpack_generated_files = [
+      all_required_files = [
         "manifest.json",
         server_bundle_js_file,
         rsc_bundle_js_file,
         react_client_manifest_file,
         react_server_client_manifest_file
       ].compact_blank
+
+      if webpack_generated_files.empty?
+        self.webpack_generated_files = all_required_files
+      else
+        missing_files = all_required_files.reject { |file| webpack_generated_files.include?(file) }
+        self.webpack_generated_files += missing_files if missing_files.any?
+      end
     end
 
     def configure_skip_display_none_deprecation
--- a/spec/react_on_rails/configuration_spec.rb
+++ b/spec/react_on_rails/configuration_spec.rb
@@ -525,6 +525,152 @@ module ReactOnRails
         end
       end
     end
+
+    describe "#ensure_webpack_generated_files_exists" do
+      let(:config) { described_class.new }
+
+      before do
+        # Reset to test defaults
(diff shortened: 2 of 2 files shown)


justin808:
@ihabadham I'm finishing this one up.

PR #1848, opened 2025-10-06
fix: Auto-remove default gem files to prevent installer conflicts

justin808:
## Summary

Automatically detect and remove default `bin/dev` (Rails) and `config/shakapacker.yml` (Shakapacker) files before running the React on Rails installer to prevent interactive conflict prompts.

## Changes

- Added `remove_default_gem_files` method that runs before generator invocation
- Added `remove_default_bin_dev` to check and delete Rails default `bin/dev`
- Added `remove_default_shakapacker_yml` to check and delete Shakapacker default config
- Added `shakapacker_yml_matches_default?` helper to safely compare files
- Only removes files that exactly match gem defaults (prevents accidental deletion)
- Provides user feedback when removing default files

## Problem Solved

Previously, when running `rails generate react_on_rails:install` on a Rails app with default Shakapacker installation, the generator would pause with interactive prompts asking whether to overwrite conflicting files. This interrupted automated installations and required manual intervention.

## Solution

The installer now:
1. Detects if `bin/dev` exists and matches the Rails default
2. Detects if `config/shakapacker.yml` exists and matches Shakapacker default  
3. Automatically removes these files if they match defaults
4. Proceeds with installation without interruption
5. Preserves any customized versions of these files

## Test Plan

- [ ] Run installer on fresh Rails app with default `bin/dev` - should auto-remove
- [ ] Run installer on app with Shakapacker installed - should auto-remove default config
- [ ] Run installer on app with customized files - should preserve them
- [ ] Verify RuboCop passes (already verified locally)
- [ ] Check CI passes


justin808:
@claude review this PR

justin808:
Closing this PR as there are no net changes after addressing code review feedback.

## Summary

This PR went through several iterations:

1. **Initial Implementation**: Added auto-remove feature for default `bin/dev` and `config/shakapacker.yml` files
2. **Iterative Refinements**: Multiple commits improving safety, robustness, and edge case handling
3. **CI Test Failures**: Encountered persistent issues with the feature in CI environment
4. **Feature Revert** (commit 24c4c9cd): Removed the entire auto-remove feature as it proved too complex and caused more problems than it solved
5. **Final Revert** (commit 4fd39fc0): Reverted the Gemfile path change based on code review feedback

## Final State

After all changes, this PR has **zero net changes** compared to master. The simpler `"Gemfile"` path is more reliable in generator context than `File.join(destination_root, "Gemfile")`.

## Lessons Learned

- Auto-removing files during installation is more complex than anticipated
- Race conditions between gem installers and file checks are difficult to handle robustly
- Sometimes the simpler approach (no auto-removal) is the correct solution
- Code review feedback correctly identified that the Gemfile path "fix" was actually making things more complex

Thank you to the reviewers for the detailed feedback!

PR #1853, opened 2025-10-07
feat: Implement async props with incremental rendering for React Server Components

AbanoubGhadban:
### TL;DR

This PR adds a new helper `stream_react_component_with_async_props` that lets you start rendering a React component **immediately** while database queries run in the background. As each query completes, the result streams to the component, which progressively updates. This dramatically improves Time to First Byte (TTFB) for pages with slow data fetches.

```ruby
# Before: Waits for ALL queries before rendering starts (3+ seconds)
<%= stream_react_component("Dashboard", props: {
  users: User.all.to_a,        # 1 second
  posts: Post.recent.to_a,     # 1 second
  metrics: Metric.summary      # 1 second
}) %>

# After: Renders immediately, queries run in parallel (50ms TTFB)
<%= stream_react_component_with_async_props("Dashboard", props: { title: "My Dashboard" }) do |emit|
  emit.call("users", User.all.to_a)      # Streams when ready
  emit.call("posts", Post.recent.to_a)   # Streams when ready
  emit.call("metrics", Metric.summary)   # Streams when ready
end %>
```

---

## Table of Contents

1. [The Problem](#the-problem)
2. [The Solution](#the-solution)
3. [Key Concepts & Terminology](#key-concepts--terminology)
4. [Dependencies & Infrastructure](#dependencies--infrastructure)
5. [How It Works](#how-it-works)
6. [Architecture Overview](#architecture-overview)
7. [Usage Guide](#usage-guide)
8. [Design Decisions](#design-decisions)
9. [Implementation Details](#implementation-details)
10. [Testing](#testing)
11. [Files Changed](#files-changed)

---

## The Problem

React on Rails Pro supports **streaming SSR** using React 18's `renderToPipeableStream`. This allows the server to start sending HTML before the entire page is rendered, showing loading states via `<Suspense>` boundaries.

However, there's a fundamental bottleneck: **all props must be fetched before rendering begins**.

```
Current Flow:
┌────────────────────────────────────────────────────────────────────┐
│ Time: 0ms                                              3000ms      │
│                                                                    │
│ Rails:  [──── Fetch users ────][── Fetch posts ──][─ metrics ─]   │
│                                                          │         │
│ Node:                                                    [render]  │
│                                                              │     │
│ Browser:                                                     [HTML]│
│                                                                    │
│ User sees: Nothing for 3 seconds, then entire page appears        │
└────────────────────────────────────────────────────────────────────┘
```

Even though React *can* stream HTML incrementally, we're not leveraging it because Rails blocks on data fetching first.

**What we want:**

```
Desired Flow:
┌────────────────────────────────────────────────────────────────────┐
│ Time: 0ms        500ms      1500ms     2500ms          3000ms      │
│                                                                    │
│ Rails:  [── users ──]                                              │
│         [────── posts ──────]                                      │
│         [─────────── metrics ───────────]                          │
│              ↓           ↓           ↓                             │
│ Node:   [shell]────[+users]────[+posts]────[+metrics]              │
│            ↓           ↓           ↓           ↓                   │
│ Browser: [loading] [users!]   [posts!]    [complete]               │
│                                                                    │
│ User sees: Loading state at 50ms, progressive content after       │
└────────────────────────────────────────────────────────────────────┘
```

---

## The Solution

We introduce **async props** - props that are declared upfront but whose values arrive *after* rendering has already started.

The React component renders immediately with loading states. As each async prop value arrives from Rails, the component re-renders with the new data. This is powered by:

1. **Bidirectional HTTP/2 streaming** between Rails and Node
2. **Promise-based suspension** in React Server Components
3. **Request-scoped state sharing** to coordinate between initial render and updates

### What This PR Adds

| Component | Purpose |
|-----------|---------|
| `stream_react_component_with_async_props` | New Rails helper that accepts a block for async props |
| `AsyncPropsEmitter` | Ruby class that sends prop values to Node as they become available |
| `/incremental-render/` endpoint | New Node endpoint that accepts streaming updates |
| `AsyncPropsManager` | TypeScript class that manages promises for async props |
| `sharedExecutionContext` | Request-scoped Map for sharing state between render and updates |

---

## Key Concepts & Terminology

Before diving into the architecture, let's define the key concepts:

### NDJSON (Newline-Delimited JSON)

A simple streaming format where each line is a valid JSON object:

```
{"type": "init", "data": "..."}
{"type": "update", "propName": "users", "value": [...]}
{"type": "update", "propName": "posts", "value": [...]}
```

**Why NDJSON?** It allows streaming JSON objects one at a time over a single HTTP connection. Each line can be parsed independently as it arrives.

### Bidirectional Streaming (HTTP/2)

HTTP/2 allows sending data in both directions simultaneously on a single connection:

```
┌─────────────────────────────────────────────────────────────┐
│                    HTTP/2 Stream                             │
├─────────────────────────────────────────────────────────────┤
│  Rails → Node:  Request data, then updates as props ready   │
│  Node → Rails:  HTML chunks as React renders them           │
│                                                              │
│  Both happen at the same time on ONE connection!            │
└─────────────────────────────────────────────────────────────┘
```

### React Suspense & Streaming SSR

React 18's streaming SSR can pause rendering when it hits a Promise, continue with other parts of the tree, and resume when the Promise resolves:

```jsx
function Dashboard() {
  return (
    <div>
      <Header />  {/* Renders immediately */}
      <Suspense fallback={<Spinner />}>
        <UserList users={asyncPromise} />  {/* Waits for promise */}
      </Suspense>
      <Footer />  {/* Renders immediately */}
    </div>
  );
}
```

### ExecutionContext & sharedExecutionContext

When Node renders a React component, it creates an **ExecutionContext** - an isolated environment for that specific HTTP request.

The `sharedExecutionContext` is a `Map` within this environment that allows the initial render code and subsequent update code to share data (like the AsyncPropsManager instance) **without using global variables**.

```javascript
// During initial render:
sharedExecutionContext.set("asyncPropsManager", manager);

// During update (same request, moments later):
const manager = sharedExecutionContext.get("asyncPropsManager");
manager.setProp("users", data);
```

**Why not globals?** The Node renderer handles multiple concurrent requests. Global variables would leak data between users.

---

#### `async` gem (already in use)

Provides cooperative concurrency via Fibers. Key constructs we use:

```ruby
# Barrier: Allows spawning concurrent tasks and waiting for all
Async do |task|
  barrier = Async::Barrier.new

  barrier.async { fetch_users }    # Runs concurrently
  barrier.async { fetch_posts }    # Runs concurrently

  barrier.wait  # Waits for all to complete
end
```

We use `barrier.async` to run the async props block concurrently with streaming the response.

#### `httpx` gem with `stream_bidi` plugin

HTTPX is our HTTP client for communicating with the Node renderer. The `stream_bidi` plugin enables bidirectional streaming:

```ruby
# Normal request: Send body, then receive response
response = connection.post(url, body: data)

# Bidirectional: Send data WHILE receiving response
request = connection.build_request("POST", url, stream: true)
response = connection.request(request, stream: true)
request << "more data"  # Send while response streams back
request.close           # Signal end of request body
```

#### Fastify HTTP Server

The Node renderer uses Fastify to handle HTTP requests. We register a new endpoint:

```typescript
app.post('/bundles/:bundleTimestamp/incremental-render/:digest', handler);
```

#### Node.js VM Module

Each JavaScript bundle runs in an isolated VM context. The `ExecutionContext` wraps VM operations and manages the `sharedExecutionContext` Map.

---

## How It Works

Let's trace through a complete request step by step:

### Step 1: Rails View Calls Helper

```erb
<%= stream_react_component_with_async_props("Dashboard", props: { title: "Hi" }) do |emit|
  emit.call("users", User.all.to_a)
end %>
```

The helper:
1. Stores the block in `options[:async_props_block]`
2. Calls the standard `stream_react_component` pipeline

### Step 2: Ruby Detects Async Props Mode

In `NodeRenderingPool#eval_streaming_js`:

```ruby
async_props_block = render_options.internal_option(:async_props_block)

if async_props_block
  # Use incremental rendering endpoint
  Request.render_code_with_incremental_updates(path, js_code, async_props_block: async_props_block)
else
  # Use standard streaming endpoint
  Request.render_code_as_stream(path, js_code)
end
```

### Step 3: Rails Opens Bidirectional HTTP/2 Stream

```ruby
# Build request that supports bidirectional streaming
request = connection.build_request("POST", path,
  headers: { "content-type" => "application/x-ndjson" },
  body: [],
  stream: true
)

# Create emitter that will write to this request
emitter = AsyncPropsEmitter.new(bundle_timestamp, request)

# Start the request - response begins streaming immediately
response = connection.request(request, stream: true)

# Send initial render request as first NDJSON line
request << "#{initial_data.to_json}\n"
```

### Step 4: Rails Runs Async Props Block Concurrently

```ruby
barrier.async do
  async_props_block.call(emitter)  # User's block runs here
ensure
  request.close  # Signals end of request body
end
```

The block runs in a separate Fiber while the response streams back to the client.

### Step 5: Node Receives Initial Request

The `/incremental-render/` endpoint parses the first NDJSON line:

```typescript
// handleIncrementalRenderStream.ts
for await (const chunk of request.raw) {
  // Parse NDJSON lines
  const parsed = JSON.parse(line);

  if (!hasReceivedFirstObject) {
    // First line: Start rendering
    const result = await handleIncrementalRenderRequest(parsed);
    // Returns response stream + sink for updates
  }
}
```

### Step 6: Node Executes Initial Render

The rendering code includes setup for async props:

```javascript
// Generated by ServerRenderingJsCode#async_props_setup_js
if (ReactOnRails.isRSCBundle) {
  var { props, asyncPropManager } =
    ReactOnRails.addAsyncPropsCapabilityToComponentProps(usedProps);
  usedProps = props;
  sharedExecutionContext.set("asyncPropsManager", asyncPropManager);
}

// Then render the component with enhanced props
return ReactOnRails.serverRenderRSCReactComponent({ name, props: usedProps, ... });
```

### Step 7: React Component Requests Async Props

```tsx
async function Dashboard({ title, getReactOnRailsAsyncProp }) {
  // This returns a Promise that will resolve when the prop arrives
  const usersPromise = getReactOnRailsAsyncProp('users');

  return (
    <Suspense fallback={<Spinner />}>
      <UserList users={usersPromise} />  {/* Suspends on promise */}
    </Suspense>
  );
}
```

React suspends on the unresolved promise and streams the initial shell with loading state.

### Step 8: Rails Sends Async Prop Value

When `emit.call("users", data)` is called:

```ruby
# AsyncPropsEmitter#call
update_chunk = {
  bundleTimestamp: @bundle_timestamp,
  updateChunk: <<~JS
    (function(){
      var manager = sharedExecutionContext.get("asyncPropsManager");
      manager.setProp("users", #{data.to_json});
    })()
  JS
}
@request_stream << "#{update_chunk.to_json}\n"
```

### Step 9: Node Executes Update Chunk

```typescript
// handleIncrementalRenderStream.ts - for subsequent NDJSON lines
await incrementalSink.add(parsed);

// handleIncrementalRenderRequest.ts - sink.add implementation
add: async (chunk) => {
  await executionContext.runInVM(chunk.updateChunk, bundlePath);
}
```

The update chunk JavaScript runs in the same ExecutionContext, accessing the same `sharedExecutionContext` Map.

### Step 10: Promise Resolves, React Continues

```typescript
// AsyncPropsManager#setProp
setProp(propName, propValue) {
  const controller = this.propNameToPromiseController.get(propName);
  controller.resolve(propValue);  // Promise resolves!
}
```

React detects the resolved promise, renders the component with data, and streams more HTML.

### Step 11: Stream Ends

When the Rails block completes and `request.close` is called:

1. HTTP/2 END_STREAM flag is sent
2. Node calls `sink.handleRequestClosed()`
3. This executes `onRequestClosedUpdateChunk` which calls `asyncPropsManager.endStream()`
4. Any props that were never sent get their promises rejected

---

## Architecture Overview

Now that you understand the concepts, here's how they fit together:

### Sequence Diagram

```mermaid
sequenceDiagram
    participant View as Rails View
    participant Emitter as AsyncPropsEmitter
    participant HTTPX as HTTPX (HTTP/2)
    participant Node as Node Renderer
    participant VM as ExecutionContext
    participant React as React Component
    participant Manager as AsyncPropsManager

    View->>HTTPX: Open bidirectional stream
    HTTPX->>Node: NDJSON Line 1: {renderingRequest}

    Node->>VM: Execute initial render code
    VM->>Manager: Create AsyncPropsManager
    VM->>VM: Store manager in sharedExecutionContext
    VM->>React: Render component

    React->>Manager: getReactOnRailsAsyncProp("users")
    Manager-->>React: Returns Promise (unresolved)
    React-->>React: Suspends on Promise

    Node-->>HTTPX: Stream initial HTML (loading states)
    HTTPX-->>View: Forward HTML to client

    Note over View,Emitter: Meanwhile, in barrier.async fiber...

    View->>Emitter: emit.call("users", data)
    Emitter->>HTTPX: NDJSON Line 2: {updateChunk}
    HTTPX->>Node: Forward update

    Node->>VM: Execute updateChunk
    VM->>Manager: setProp("users", data)
    Manager->>React: Resolve Promise

    React-->>React: Resume rendering with data
    Node-->>HTTPX: Stream more HTML
    HTTPX-->>View: Forward HTML

    View->>HTTPX: Close request (block complete)
    HTTPX->>Node: END_STREAM
    Node->>VM: Execute onRequestClosedUpdateChunk
    VM->>Manager: endStream()
```

### Component Diagram

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                              RAILS APPLICATION                               │
│                                                                              │
│  ┌────────────────────────────────────────────────────────────────────────┐ │
│  │  View: dashboard/index.html.erb                                        │ │
│  │                                                                         │ │
│  │  <%= stream_react_component_with_async_props("Dashboard") do |emit| %> │ │
│  │    <% emit.call("users", User.all) %>                                  │ │
│  │  <% end %>                                                             │ │
│  └──────────────────────────────┬─────────────────────────────────────────┘ │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  ReactOnRailsProHelper                                                │   │
│  │  ├── stream_react_component_with_async_props                         │   │
│  │  │   └── Stores block, calls stream_react_component                  │   │
│  │  └── consumer_stream_async                                            │   │
│  │      └── Sets up barrier for concurrent execution                     │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  NodeRenderingPool#eval_streaming_js                                  │   │
│  │  └── Detects async_props_block, routes to incremental endpoint       │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  Request#render_code_with_incremental_updates                         │   │
│  │  ├── Creates HTTPX bidirectional stream                               │   │
│  │  ├── Sends initial NDJSON request                                    │   │
│  │  └── barrier.async { block.call(emitter); request.close }            │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  AsyncPropsEmitter                                                    │   │
│  │  ├── call(prop_name, value) → Sends NDJSON update line              │   │
│  │  └── end_stream_chunk → Code to run when request closes              │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
└─────────────────────────────────┼────────────────────────────────────────────┘
                                  │
                                  │  HTTP/2 + NDJSON
                                  ▼
┌─────────────────────────────────────────────────────────────────────────────┐
│                              NODE RENDERER                                   │
│                                                                              │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  worker.ts - Fastify Server                                           │   │
│  │  └── POST /bundles/:timestamp/incremental-render/:digest             │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  handleIncrementalRenderStream.ts                                     │   │
│  │  ├── Parses incoming NDJSON stream                                   │   │
│  │  ├── Line 1 → onRenderRequestReceived (initial render)              │   │
│  │  ├── Line N → onUpdateReceived (prop updates)                       │   │
│  │  └── END_STREAM → onRequestEnded (cleanup)                          │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  handleIncrementalRenderRequest.ts                                    │   │
│  │  ├── Executes initial render via handleRenderRequest                 │   │
│  │  └── Returns "sink" object with:                                     │   │
│  │      ├── add(chunk) → Executes update in VM                         │   │
│  │      └── handleRequestClosed() → Executes cleanup code              │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  vm.ts - ExecutionContext                                             │   │
│  │  ├── Created per HTTP request (not shared between requests!)        │   │
│  │  ├── sharedExecutionContext = new Map()                              │   │
│  │  │   └── Allows initial render & updates to share state             │   │
│  │  └── runInVM(code) → Executes JS in isolated VM context             │   │
│  └──────────────────────────────────────────────────────────────────────┘   │
│                                                                              │
└─────────────────────────────────────────────────────────────────────────────┘
                                  │
                                  │  VM Execution
                                  ▼
┌─────────────────────────────────────────────────────────────────────────────┐
│                           REACT-ON-RAILS-PRO PACKAGE                         │
│                                                                              │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  ReactOnRailsRSC.ts                                                   │   │
│  │  └── addAsyncPropsCapabilityToComponentProps(props)                  │   │
│  │      ├── Creates AsyncPropsManager instance                          │   │
│  │      ├── Adds getReactOnRailsAsyncProp function to props            │   │
│  │      └── Returns { asyncPropManager, props }                         │   │
│  └──────────────────────────────┬───────────────────────────────────────┘   │
│                                 │                                            │
│                                 ▼                                            │
│  ┌──────────────────────────────────────────────────────────────────────┐   │
│  │  AsyncPropsManager.ts                                                 │   │
│  │  ├── getProp(name) → Returns Promise (cached per prop name)         │   │
│  │  ├── setProp(name, value) → Resolves the Promise                    │   │
│  │  └── endStream() → Rejects all unresolved Promises                  │   │
│  └──────────────────────────────────────────────────────────────────────┘   │
│                                                                              │
└─────────────────────────────────────────────────────────────────────────────┘
```

---

### Basic Example

**Rails View:**
```erb
<%# app/views/dashboard/index.html.erb %>
<%= stream_react_component_with_async_props("Dashboard",
      props: { title: "My Dashboard" }) do |emit| %>

  <%# These queries run concurrently with rendering %>
  <% emit.call("users", User.active.includes(:profile).to_a) %>
  <% emit.call("recent_posts", Post.recent.limit(10).to_a) %>
  <% emit.call("notifications", current_user.notifications.unread.to_a) %>

<% end %>
```

**React Component:**
```tsx
// app/javascript/components/Dashboard.tsx
import { Suspense } from 'react';
import type { WithAsyncProps } from 'react-on-rails-pro';

// Define your prop types
type SyncProps = {
  title: string;
};

type AsyncProps = {
  users: User[];
  recent_posts: Post[];
  notifications: Notification[];
};

// WithAsyncProps adds getReactOnRailsAsyncProp to your props
type Props = WithAsyncProps<AsyncProps, SyncProps>;

export default function Dashboard({ title, getReactOnRailsAsyncProp }: Props) {
  return (
    <div>
      <h1>{title}</h1>  {/* Renders immediately */}

      <Suspense fallback={<UsersSkeleton />}>
        <UsersList users={getReactOnRailsAsyncProp('users')} />
      </Suspense>

      <Suspense fallback={<PostsSkeleton />}>
        <RecentPosts posts={getReactOnRailsAsyncProp('recent_posts')} />
      </Suspense>

      <Suspense fallback={<NotificationsSkeleton />}>
        <NotificationsBell items={getReactOnRailsAsyncProp('notifications')} />
      </Suspense>
    </div>
  );
}

// Async component that awaits the promise
async function UsersList({ users }: { users: Promise<User[]> }) {
  const resolvedUsers = await users;
  return (
    <ul>
      {resolvedUsers.map(user => (
        <li key={user.id}>{user.name}</li>
      ))}
    </ul>
  );
}
```

### Error Handling

If a prop is never sent (e.g., the block raises an error), the promise will reject when `endStream()` is called:

```tsx
import { ErrorBoundary } from 'react-error-boundary';

function Dashboard({ getReactOnRailsAsyncProp }) {
  return (
    <ErrorBoundary fallback={<p>Failed to load users</p>}>
      <Suspense fallback={<Spinner />}>
        <UsersList users={getReactOnRailsAsyncProp('users')} />
      </Suspense>
    </ErrorBoundary>
  );
}
```

### When to Use Async Props

**Good candidates:**
- Slow database queries (N+1, complex joins)
- External API calls
- Heavy computations
- Data that varies in load time

**Not needed for:**
- Fast queries (< 50ms)
- Static data
- Data needed for the initial shell/layout

---

### Why `sharedExecutionContext` Instead of Global Variables?

**Problem:** The initial render and update chunks need to share the `AsyncPropsManager` instance. Where should it live?

**Option 1: Global Variable (Rejected)**
```javascript
// ❌ DANGEROUS - Node handles multiple concurrent requests
global.asyncPropsManager = new AsyncPropsManager();

// Request A stores Alice's manager
global.asyncPropsManager = aliceManager;

// Request B (different user!) might read it
global.asyncPropsManager.setProp("data", bobsData); // Corrupts Alice's render!
```

**Option 2: Request-Scoped Map (Chosen)**
```javascript
// ✅ SAFE - Each request has its own Map
// Created in buildExecutionContext():
const sharedExecutionContext = new Map();

// Initial render stores manager
sharedExecutionContext.set("asyncPropsManager", manager);

// Update chunk (same ExecutionContext) retrieves it
const manager = sharedExecutionContext.get("asyncPropsManager");
```

The `sharedExecutionContext` is:
- Created once per HTTP request
- Passed to VM only during code execution
- Never stored globally
- Automatically garbage collected when request ends

### Why Function in Props Instead of a Hook?

**Problem:** How should the component access async props?

**Option 1: Custom Hook (Rejected)**
```tsx
// ❌ DOESN'T WORK - React Server Components can't use hooks
function Dashboard() {
  const users = useAsyncProps('users');  // Error: Hooks not available in RSC
  return <div>{users}</div>;
}
```

**Option 2: Function in Props (Chosen)**
```tsx
// ✅ WORKS - Just a closure, no hook runtime needed
function Dashboard({ getReactOnRailsAsyncProp }) {
  const users = await getReactOnRailsAsyncProp('users');
  return <div>{users}</div>;
}
```

React Server Components are async functions, not components with a lifecycle. They can't use hooks because the hook runtime isn't available server-side.

### Why Cache Promises in AsyncPropsManager?

**Problem:** React may call render functions multiple times. What happens if `getProp()` returns a new Promise each time?

```typescript
// ❌ BAD - New promise each call
getProp(name) {
  return new Promise(resolve => { ... });  // Different object!
}

// React renders, gets Promise A
// React re-renders, gets Promise B (different object)
// React thinks state changed, re-renders again
// Gets Promise C... infinite loop! 💥
```

**Solution:** Return the same Promise object for repeated calls:

```typescript
// ✅ GOOD - Cached promise
private cache = new Map<string, PromiseController>();

getProp(name) {
  if (this.cache.has(name)) {
    return this.cache.get(name).promise;  // Same object
  }
  // ... create and cache new promise
}
```

---

### NDJSON Protocol

**Request (Rails → Node):**

Line 1 - Initial render request:
```json
{
  "gemVersion": "16.2.0",
  "protocolVersion": "2.0.0",
  "password": "...",
  "renderingRequest": "(function(componentName = 'Dashboard') { ... })()",
  "onRequestClosedUpdateChunk": {
    "bundleTimestamp": "abc123",
    "updateChunk": "(function(){ sharedExecutionContext.get('asyncPropsManager').endStream(); })()"
  }
}
```

Line 2+ - Update chunks:
```json
{
  "bundleTimestamp": "abc123",
  "updateChunk": "(function(){ sharedExecutionContext.get('asyncPropsManager').setProp('users', [...]); })()"
}
```

**Response (Node → Rails):**

Streaming HTML chunks:
```json
{"html": "<div id=\"Dashboard\"><!--$-->Loading...", "consoleReplayScript": "", "isShellReady": true}
{"html": "<ul><li>User 1</li></ul><!--/$-->", "consoleReplayScript": "", "isShellReady": true}
```

### HTTPX stream_bidi Patch

We discovered a bug in HTTPX's `stream_bidi` plugin where the `@headers_sent` flag isn't reset on retry, causing crashes. We include a patch:

```ruby
# lib/react_on_rails_pro/httpx_stream_bidi_patch.rb
module HTTPX::Plugins::StreamBidi::RequestMethodsRetryFix
  def transition(nextstate)
    @headers_sent = false if nextstate == :idle
    super
  end
end
```

This will be removed once fixed upstream.

---

### Running Tests

```bash
# Node Renderer unit tests
cd packages/react-on-rails-pro-node-renderer
pnpm test

# AsyncPropsManager unit tests
cd packages/react-on-rails-pro
pnpm test

# Ruby unit tests
cd react_on_rails_pro
bundle exec rspec

# Integration tests (requires running node renderer)
cd react_on_rails_pro/spec/dummy
bundle exec rspec spec/requests/incremental_rendering_integration_spec.rb
```

### Test Coverage

| Component | Test File | Lines |
|-----------|-----------|-------|
| NDJSON Stream Parser | `handleIncrementalRenderStream.test.ts` | 205 |
| Incremental Render Handler | `incrementalRender.test.ts` | 875 |
| HTML Streaming | `incrementalHtmlStreaming.test.ts` | 233 |
| AsyncPropsManager | `AsyncPropManager.test.ts` | 142 |
| Ruby AsyncPropsEmitter | `async_props_emitter_spec.rb` | 43 |
| Ruby Request | `request_spec.rb` | 184 |

---

### New Files

| File | Purpose |
|------|---------|
| `handleIncrementalRenderRequest.ts` | Handles initial incremental render, returns sink |
| `handleIncrementalRenderStream.ts` | Parses NDJSON stream from Rails |
| `AsyncPropsManager.ts` | Manages promises for async props |
| `async_props_emitter.rb` | Sends prop values to Node |
| `httpx_stream_bidi_patch.rb` | Fixes HTTPX retry bug |

### Modified Files

| File | Changes |
|------|---------|
| `worker.ts` | Added `/incremental-render/` endpoint |
| `vm.ts` | Added `sharedExecutionContext` to ExecutionContext |
| `request.rb` | Added `render_code_with_incremental_updates` |
| `server_rendering_js_code.rb` | Added `async_props_setup_js` |
| `node_rendering_pool.rb` | Routes async props to incremental endpoint |
| `react_on_rails_pro_helper.rb` | Added `stream_react_component_with_async_props` |
| `ReactOnRailsRSC.ts` | Added `addAsyncPropsCapabilityToComponentProps` |
| `types/index.ts` | Added `WithAsyncProps` type |

---

## Breaking Changes

**None.** This is an additive feature. Existing code continues to work unchanged.

---

## Future Improvements

- [ ] Add timeout support for individual async props
- [ ] Performance benchmarks comparing before/after
- [ ] DevTools integration to visualize async prop timing
- [ ] Support for async props in non-RSC streaming (client components)

## Summary by CodeRabbit

* **New Features**
  * Incremental NDJSON HTML streaming over HTTP/2 with bidirectional support and per-request execution contexts.
  * Async props system for progressive prop delivery to components during streaming; test and demo components added.
  * Bundle existence validation to prevent renders when required bundles are missing.

* **Bug Fixes**
  * Fixed HTTP/2 streaming reliability and END_STREAM signaling; improved streaming error handling and size limits.

* **Security**
  * Dependency updates to mitigate CVE-2025-55182.

### DIFF
--- a/react_on_rails_pro/app/helpers/react_on_rails_pro_helper.rb
+++ b/react_on_rails_pro/app/helpers/react_on_rails_pro_helper.rb
@@ -142,6 +142,11 @@ module ReactOnRailsProHelper
     stream_react_component(component_name, options)
   end
 
+  def rsc_payload_react_component_with_async_props(component_name, options = {}, &props_block)
+    options[:async_props_block] = props_block
+    rsc_payload_react_component(component_name, options)
+  end
+
   # Renders the React Server Component (RSC) payload for a given component. This helper generates
   # a special format designed by React for serializing server components and transmitting them
   # to the client.
--- a/react_on_rails_pro/app/views/react_on_rails_pro/rsc_payload.html.erb
+++ b/react_on_rails_pro/app/views/react_on_rails_pro/rsc_payload.html.erb
@@ -1,1 +1,19 @@
+<%# TODO Remove after adding support for requesting async props from node renderer side %>
+
+<% if @rsc_payload_component_name.start_with? "AsyncPropsComponent" %>
+<%= rsc_payload_react_component_with_async_props("AsyncPropsComponent", props: { name: "John Doe", age: 30, description: "Software Engineer" }) do |emit|
+  if @rsc_payload_component_name == "AsyncPropsComponentForTesting"
+    read_async_props_from_redis(emit)
+    next
+  end
+
+  # Simulate fetching async props
+  sleep 1
+  emit.call("books", ["The Pragmatic Programmer", "Clean Code", "Design Patterns"])
+
+  sleep 1
+  emit.call("researches", ["Machine Learning Study", "React Performance Optimization", "Database Indexing Strategies"])
+end %>
+<% else %>
 <%= rsc_payload_react_component(@rsc_payload_component_name, props: @rsc_payload_component_props) %>
+<% end %>
--- a/react_on_rails_pro/spec/dummy/app/controllers/pages_controller.rb
+++ b/react_on_rails_pro/spec/dummy/app/controllers/pages_controller.rb
@@ -173,8 +173,40 @@ class PagesController < ApplicationController # rubocop:disable Metrics/ClassLen
   # See files in spec/dummy/app/views/pages
 
   helper_method :calc_slow_app_props_server_render
+  helper_method :read_async_props_from_redis
 
   private
+
+  def read_async_props_from_redis(emitter)
+    redis = ::Redis.new
+    request_id = params[:request_id]
+
+    unless request_id
+      sleep 1
+      raise "You must pass the request_id param to the page, this page is inteded to be used for testing only"
+    end
+
+    ended = false
+    last_received_id = "0-0"
+    stream_id = "stream:#{request_id}"
+    until ended
+      received_messages = redis.xread(stream_id, last_received_id, block: 0)[stream_id]
+      # receive_messages are like [[msg1_id, [**msg_entries]], [msg2_id, [**msg_entries]]]
+      last_received_id = received_messages.last.first
+      received_messages.each_value do |message_entries|
+        message_entries.each do |message_key, message_value|
+          if message_key == "end"
+            ended = true
+            next
+          end
+
+          sleep 0.1
+          # Key starts with :
+          emitter.call(message_key[1..], JSON.parse(message_value))
+        end
+      end
+    end
+  end
 
   def calc_slow_app_props_server_render
     msg = <<-MSG.strip_heredoc
--- a/react_on_rails_pro/spec/dummy/app/views/pages/server_router.html.erb
+++ b/react_on_rails_pro/spec/dummy/app/views/pages/server_router.html.erb
@@ -1,11 +1,35 @@
-<%= stream_react_component("ServerComponentRouter", 
+<%
+  options = {
     props: @app_props_server_render.merge(
       artificialDelay: params[:artificial_delay] || 0,
       postsCount: params[:posts_count] || 2,
       requestId: params[:request_id],
+      name: "John Doe",
+      age: 30,
+      description: "Software Engineer",
     ),
     trace: true, 
-    id: "ServerComponentRouter-react-component-0") %>
+    id: "ServerComponentRouter-react-component-0"
+  }
+%>
+
+<% if request.path.include?("/async-props-component") %>
+  <%= stream_react_component_with_async_props("ServerComponentRouter", **options) do |emitter|
+        if request.path.end_with?("/async-props-component-for-testing")
+          read_async_props_from_redis(emitter)
+          next
+        end
+
+        # Simulate fetching async props
+        sleep 1
+        emitter.call("books", ["The Pragmatic Programmer", "Clean Code", "Design Patterns"])
+
+        sleep 1
+        emitter.call("researches", ["Machine Learning Study", "React Performance Optimization", "Database Indexing Strategies"])
+      end %>
+<% else %>
+  <%= stream_react_component("ServerComponentRouter", **options) %>
+<% end %>
 <hr/>
 
 <h1>React Rails Server Streaming Server Rendered RSC Components</h1>
--- a/react_on_rails_pro/spec/dummy/client/app/components/ServerComponentRouter.tsx
+++ b/react_on_rails_pro/spec/dummy/client/app/components/ServerComponentRouter.tsx
@@ -47,6 +47,9 @@ export default function App({ basePath = '/server_router', ...props }: { basePat
           <li>
             <Link to={`${basePath}/server-component-with-retry`}>Server Component with Retry</Link>
           </li>
+          <li>
+            <Link to={`${basePath}/async-props-component`}>Async Props Component</Link>
+          </li>
         </ul>
       </nav>
       <Suspense fallback={<div>Loading Page...</div>}>
@@ -83,6 +86,14 @@ export default function App({ basePath = '/server_router', ...props }: { basePat
             element={<RSCRoute componentName="AsyncComponentsTreeForTesting" componentProps={props} />}
           />
           <Route path={`${basePath}/server-component-with-retry`} element={<ServerComponentWithRetry />} />
+          <Route
+            path={`${basePath}/async-props-component`}
+            element={<RSCRoute componentName="AsyncPropsComponent" componentProps={props} />}
+          />
+          <Route
+            path={`${basePath}/async-props-component-for-testing`}
+            element={<RSCRoute componentName="AsyncPropsComponentForTesting" componentProps={props} />}
+          />
         </Routes>
       </Suspense>
     </ErrorBoundary>
--- a/react_on_rails_pro/spec/dummy/client/app/ror-auto-load-components/AsyncPropsComponentForTesting.tsx
+++ b/react_on_rails_pro/spec/dummy/client/app/ror-auto-load-components/AsyncPropsComponentForTesting.tsx
@@ -1,0 +1,2 @@
+// eslint-disable-next-line no-restricted-exports
+export { default } from './AsyncPropsComponent';
--- a/react_on_rails_pro/spec/dummy/client/node-renderer.js
+++ b/react_on_rails_pro/spec/dummy/client/node-renderer.js
@@ -42,10 +42,10 @@ const config = {
   // be set if you wish to have automatic worker restarting, say to clear memory leaks.
 
   // time in minutes between restarting all workers
-  allWorkersRestartInterval: (env.CI ? 2 : env.RENDERER_ALL_WORKERS_RESTART_INTERVAL) || 10,
+  allWorkersRestartInterval: (env.CI ? 2 : env.RENDERER_ALL_WORKERS_RESTART_INTERVAL) || 0.2,
 
   // time in minutes between each worker restarting when restarting all workers
-  delayBetweenIndividualWorkerRestarts: env.CI ? 0.01 : 1,
+  delayBetweenIndividualWorkerRestarts: env.CI ? 0.01 : 0.01,
 
   // If set to true, `supportModules` enables the server-bundle code to call a default set of NodeJS modules
   // that get added to the VM context: { Buffer, process, setTimeout, setInterval, clearTimeout, clearInterval }.
--- a/react_on_rails_pro/spec/dummy/config/shakapacker.yml
+++ b/react_on_rails_pro/spec/dummy/config/shakapacker.yml
@@ -38,7 +38,6 @@ development:
 
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
-    https: false
     host: localhost
     port: 3035
     # Hot Module Replacement updates modules while the application is running without a full reload
--- a/react_on_rails_pro/spec/dummy/e2e-tests/asyncProps.spec.ts
+++ b/react_on_rails_pro/spec/dummy/e2e-tests/asyncProps.spec.ts
@@ -1,0 +1,29 @@
+import { expect } from '@playwright/test';
+import { asyncPropsAtRouterPageTest } from './fixture';
+
+asyncPropsAtRouterPageTest(
+  'async props are streamed on request to the page',
+  async ({ page, sendRedisValue, endRedisStream }) => {
+    await expect(page.getByRole('heading', { name: 'Async Props Component' })).toBeVisible();
+    await expect(page.getByText('Name: John Doe')).toBeVisible();
+    await expect(page.getByText('Age: 30')).toBeVisible();
+    await expect(page.getByText('Description: Software Engineer')).toBeVisible();
+    await expect(page.getByText('Clean Code')).not.toBeVisible();
+    await expect(page.getByText('Design Patterns')).not.toBeVisible();
+    await expect(page.getByText('Engineering in software')).not.toBeVisible();
+    await expect(page.getByText('The Need for Software Engineering')).not.toBeVisible();
+
+    await sendRedisValue('books', ['Clean Code', 'Design Patterns']);
+    await expect(page.getByText('Clean Code')).toBeVisible();
+    await expect(page.getByText('Design Patterns')).toBeVisible();
+    await expect(page.getByText('Engineering in software')).not.toBeVisible();
+    await expect(page.getByText('The Need for Software Engineering')).not.toBeVisible();
+
+    await sendRedisValue('researches', ['Engineering in software', 'The Need for Software Engineering']);
+    await expect(page.getByText('Clean Code')).toBeVisible();
+    await expect(page.getByText('Design Patterns')).toBeVisible();
+    await expect(page.getByText('Engineering in software')).toBeVisible();
+    await expect(page.getByText('The Need for Software Engineering')).toBeVisible();
+    await endRedisStream();
+  },
+);
--- a/react_on_rails_pro/spec/dummy/e2e-tests/fixture.ts
+++ b/react_on_rails_pro/spec/dummy/e2e-tests/fixture.ts
@@ -11,13 +11,14 @@ type RedisRequestIdFixture = {
   nonBlockingNavigateWithRequestId: (path: string) => Promise<Response | null>;
 };
 
-type RedisReceiverPageFixture = {
+type PageFixture = {
   pagePath: string;
 };
 
 export type RedisReceiverControllerFixture = {
   sendRedisValue: (key: string, value: unknown) => Promise<void>;
   sendRedisItemValue: (itemIndex: number, value: unknown) => Promise<void>;
+  endRedisStream: () => Promise<void>;
   matchPageSnapshot: (snapshotPath: string) => Promise<void>;
   waitForConsoleMessage: (msg: string) => Promise<void>;
   getNetworkRequests: (requestUrlPattern: RegExp) => Promise<Request[]>;
@@ -66,6 +67,11 @@ const redisReceiverPageController = redisControlledTest.extend<RedisReceiverCont
       await sendRedisValue(`Item${itemIndex}`, value);
     });
   },
+  endRedisStream: async ({ redisClient, redisRequestId }, use) => {
+    await use(async () => {
+      await redisClient.xAdd(`stream:${redisRequestId}`, '*', { end: 'true' });
+    });
+  },
   matchPageSnapshot: async ({ page }, use) => {
     await use(async (snapshotPath) => {
       await expect(page.locator('.redis-receiver-container:visible')).toBeVisible();
@@ -92,7 +98,7 @@ const redisReceiverPageController = redisControlledTest.extend<RedisReceiverCont
   },
 });
 
-const redisReceiverPageTest = redisReceiverPageController.extend<RedisReceiverPageFixture>({
+const redisReceiverPageTest = redisReceiverPageController.extend<PageFixture>({
   pagePath: [
     async ({ nonBlockingNavigateWithRequestId }, use) => {
       const pagePath = '/redis_receiver_for_testing';
@@ -103,25 +109,35 @@ const redisReceiverPageTest = redisReceiverPageController.extend<RedisReceiverPa
   ],
 });
 
-const redisReceiverPageWithAsyncClientComponentTest =
-  redisReceiverPageController.extend<RedisReceiverPageFixture>({
-    pagePath: [
-      async ({ page, nonBlockingNavigateWithRequestId, sendRedisValue }, use) => {
-        const pagePath = '/redis_receiver_for_testing?async_toggle_container=true';
-        await nonBlockingNavigateWithRequestId(pagePath);
+const asyncPropsAtRouterPageTest = redisReceiverPageController.extend<PageFixture>({
+  pagePath: [
+    async ({ nonBlockingNavigateWithRequestId }, use) => {
+      const pagePath = '/server_router/async-props-component-for-testing';
+      await nonBlockingNavigateWithRequestId(pagePath);
+      await use(pagePath);
+    },
+    { auto: true },
+  ],
+});
 
-        await expect(page.getByText('Loading ToggleContainer')).toBeVisible();
-        await expect(page.locator('.toggle-button')).not.toBeVisible();
+const redisReceiverPageWithAsyncClientComponentTest = redisReceiverPageController.extend<PageFixture>({
+  pagePath: [
+    async ({ page, nonBlockingNavigateWithRequestId, sendRedisValue }, use) => {
+      const pagePath = '/redis_receiver_for_testing?async_toggle_container=true';
+      await nonBlockingNavigateWithRequestId(pagePath);
 
-        await sendRedisValue('ToggleContainer', 'anything');
-        await expect(page.locator('.toggle-button')).toBeVisible();
-        await use(pagePath);
-      },
-      { auto: true },
-    ],
-  });
+      await expect(page.getByText('Loading ToggleContainer')).toBeVisible();
+      await expect(page.locator('.toggle-button')).not.toBeVisible();
 
-const redisReceiverInsideRouterPageTest = redisReceiverPageController.extend<RedisReceiverPageFixture>({
+      await sendRedisValue('ToggleContainer', 'anything');
+      await expect(page.locator('.toggle-button')).toBeVisible();
+      await use(pagePath);
+    },
+    { auto: true },
+  ],
+});
+
+const redisReceiverInsideRouterPageTest = redisReceiverPageController.extend<PageFixture>({
   pagePath: [
     async ({ nonBlockingNavigateWithRequestId }, use) => {
       const pagePath = '/server_router/redis-receiver-for-testing';
@@ -132,7 +148,7 @@ const redisReceiverInsideRouterPageTest = redisReceiverPageController.extend<Red
   ],
 });
 
-const redisReceiverPageAfterNavigationTest = redisReceiverPageController.extend<RedisReceiverPageFixture>({
+const redisReceiverPageAfterNavigationTest = redisReceiverPageController.extend<PageFixture>({
   pagePath: [
     async ({ nonBlockingNavigateWithRequestId, page }, use) => {
       await nonBlockingNavigateWithRequestId('/server_router/simple-server-component');
@@ -145,6 +161,7 @@ const redisReceiverPageAfterNavigationTest = redisReceiverPageController.extend<
 });
 
 export {
+  asyncPropsAtRouterPageTest,
   redisReceiverPageController,
   redisReceiverPageTest,
   redisReceiverInsideRouterPageTest,
--- a/react_on_rails_pro/spec/dummy/package.json
+++ b/react_on_rails_pro/spec/dummy/package.json
@@ -11,6 +11,7 @@
     "@babel/preset-env": "7",
     "@babel/preset-react": "^7.10.1",
     "@babel/runtime": "7",
+    "@dr.pogodin/react-helmet": "^3.0.4",
     "@honeybadger-io/js": "^4.3.1",
     "@loadable/babel-plugin": "^5.16.1",
     "@loadable/component": "^5.16.3",
@@ -19,7 +20,6 @@
     "@sentry/node": "^7.120.0",
     "@shakacode/use-ssr-computation.macro": "^1.2.4",
     "@shakacode/use-ssr-computation.runtime": "^2.0.0",
-    "@webpack-cli/serve": "^1.6.0",
     "autoprefixer": "^10.4.14",
     "babel-loader": "8",
     "babel-plugin-macros": "^3.1.0",
@@ -49,7 +49,6 @@
     "react": "^19.0.3",
     "react-dom": "^19.0.3",
     "react-error-boundary": "^4.1.2",
-    "@dr.pogodin/react-helmet": "^3.0.4",
     "react-on-rails-pro": "link:.yalc/react-on-rails-pro",
     "react-on-rails-pro-node-renderer": "link:.yalc/react-on-rails-pro-node-renderer",
     "react-on-rails-rsc": "^19.0.4",
@@ -88,7 +87,7 @@
     "pino-pretty": "^13.0.0",
     "preload-webpack-plugin": "^3.0.0-alpha.1",
     "typescript": "^5.2.2",
-    "webpack-cli": "^4.10.0",
+    "webpack-cli": "^6.0.1",
     "webpack-dev-server": "^5.2.1"
   },
   "browser": {
--- a/react_on_rails_pro/spec/dummy/pnpm-lock.yaml
+++ b/react_on_rails_pro/spec/dummy/pnpm-lock.yaml
@@ -5,6 +5,7 @@ settings:
   excludeLinksFromLockfile: false
 
 importers:
+
   .:
     dependencies:
       '@apollo/client':
@@ -55,9 +56,6 @@ importers:
       '@shakacode/use-ssr-computation.runtime':
         specifier: ^2.0.0
         version: 2.0.0(react@19.2.3)
-      '@webpack-cli/serve':
-        specifier: ^1.6.0
-        version: 1.7.0(webpack-cli@4.10.0)(webpack-dev-server@5.2.2)
       autoprefixer:
         specifier: ^10.4.14
         version: 10.4.22(postcss@8.5.6)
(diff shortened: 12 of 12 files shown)


AbanoubGhadban:
## Response to Claude AI Review Comments

Thank you for the thorough reviews! After careful investigation, I want to address the issues raised and clarify which ones were already handled in the implementation.

---

### ✅ Issues Already Fixed / False Positives

#### 1. **sharedExecutionContext Memory Leak & Data Leakage** (Marked as CRITICAL)

**Status: Already Fixed**

The cleanup IS implemented in `vm.ts:374-381`:

```typescript
try {
  return vm.runInContext(renderingRequest, context) as RenderCodeResult;
} finally {
  // Clean up references immediately after execution.
  context.renderingRequest = undefined;
  context.sharedExecutionContext = undefined;  // ← CLEANUP HERE
  context.runOnOtherBundle = undefined;
}
```

**Security Design:**
- Each HTTP request creates a NEW `ExecutionContext` with its own `sharedExecutionContext` Map
- The VM context's reference is cleaned up immediately after each `runInVM()` call
- User B cannot access User A's data - they get completely different `ExecutionContext` instances

---

#### 2. **Bundle Timestamp Validation** (Marked as HIGH - Security)

**Status: Already Handled**

The validation IS implemented through `ExecutionContext` architecture in `vm.ts:339-342`:

```typescript
const vmContext = mapBundleFilePathToVMContext.get(bundleFilePath);
if (!vmContext) {
  throw new VMContextNotFoundError(bundleFilePath);  // ← VALIDATION HERE
}
```

- `ExecutionContext` is built with explicit `bundlePaths` array
- Only authorized bundles are in `mapBundleFilePathToVMContext`
- Unauthorized bundle timestamps throw `VMContextNotFoundError`

---

#### 3. **NDJSON Buffer DoS Protection** (Marked as HIGH - Security)

**Status: Already Implemented**

Size limits ARE enforced in `handleIncrementalRenderStream.ts:51-68`:

```typescript
// Check total request size limit (100MB)
if (totalBytesReceived > MAX_NDJSON_REQUEST_SIZE) {
  throw new Error(`NDJSON request exceeds maximum size...`);
}

// Check single line size limit (10MB - protects against missing newlines)
if (buffer.length > MAX_NDJSON_LINE_SIZE) {
  throw new Error(`NDJSON line exceeds maximum size...`);
}
```

---

#### 4. **assertIsUpdateChunk Null Check** (Marked as MEDIUM)

**Status: Already Implemented**

The null check IS there at `handleIncrementalRenderRequest.ts:17-20`:

```typescript
function assertIsUpdateChunk(value: unknown): asserts value is UpdateChunk {
  if (
    typeof value !== 'object' ||
    value === null ||  // ← NULL CHECK IS HERE
    ...
  )
```

(Claude acknowledged this in the same review: "Wait, I see this IS checked on lines 18-19! Good catch by the author.")

---

#### 5. **Uncaught Promise Rejection in Sink** (Marked as CRITICAL)

**Status: Already Handled**

The `.catch()` handler properly catches all errors in `handleIncrementalRenderRequest.ts:118-127`:

```typescript
await executionContext.runInVM(chunk.updateChunk, bundlePath).catch((err: unknown) => {
  log.error({ msg: 'Error running incremental render chunk', err, chunk });
});
```

The `await` waits for either success OR the catch handler to complete - no promise rejection escapes.

---

#### 6. **VM Context Eviction Race Condition** (Marked as HIGH)

**Status: Not a Problem**

The architecture prevents this:
- `ExecutionContext` stores references to VM contexts in its **own local Map**
- Even if global `vmContexts` evicts a VM, the `ExecutionContext` still holds a reference
- The VM object won't be garbage collected while `ExecutionContext` holds it

---

#### 7. **Stream Timeout** (Marked as MEDIUM)

**Status: Handled at HTTP Layer**

- Fastify has default connection/keep-alive timeouts
- HTTP/2 has built-in timeout mechanisms
- HTTPX (Ruby client) has explicit `connect_timeout` and `read_timeout` in `request.rb:404-406`

---

### ✅ One Valid Issue - Now Fixed

#### **console.error vs log utility**

This was a valid issue. Fixed by replacing `console.error` with `log.error` in `handleIncrementalRenderStream.ts`:

```typescript
// Before:
console.error(reportedMessage);

// After:
log.error({ msg: reportedMessage });
```

---

### Summary

| Issue | Claude's Severity | Actual Status |
|-------|------------------|---------------|
| sharedExecutionContext cleanup | CRITICAL | ✅ Already fixed |
| Bundle timestamp validation | HIGH | ✅ Already handled |
| NDJSON buffer DoS | HIGH | ✅ Already implemented |
| VM eviction race condition | HIGH | ✅ Not a problem |
| Stream timeout | MEDIUM | ✅ Handled at HTTP layer |
| Uncaught promise rejection | HIGH | ✅ Already handled |
| assertIsUpdateChunk null check | MEDIUM | ✅ Already implemented |
| **console.error vs log** | LOW | ✅ **Fixed now** |

**7 of 8 issues were false positives.** The `console.error` issue was valid and has been fixed.

packages/react-on-rails-pro-node-renderer/src/worker/handleIncrementalRenderStream.ts
```diff
@@ -0,0 +1,141 @@
+import { StringDecoder } from 'string_decoder';
+import type { ResponseResult } from '../shared/utils';
+import * as errorReporter from '../shared/errorReporter';
+import log from '../shared/log';
+
+// Maximum size for a single NDJSON line (10MB - matches Fastify fieldSizeLimit)
+export const MAX_NDJSON_LINE_SIZE = 10 * 1024 * 1024;
+
+// Maximum total request size (100MB - matches Fastify bodyLimit)
+export const MAX_NDJSON_REQUEST_SIZE = 100 * 1024 * 1024;
```
alexeyr-ci2: Should we use these constants there then? Or vice versa?

packages/react-on-rails-pro-node-renderer/src/worker/handleIncrementalRenderStream.ts
```diff
@@ -0,0 +1,141 @@
...
+  response: ResponseResult;
+  shouldContinue: boolean;
+}
+
+/**
+ * Options interface for incremental render stream handler
+ */
+export interface IncrementalRenderStreamHandlerOptions {
+  request: {
+    raw: NodeJS.ReadableStream | { [Symbol.asyncIterator](): AsyncIterator<Buffer> };
```
alexeyr-ci2, line 19:
```suggestion
    raw: NodeJS.ReadableStream | AsyncIterable<Buffer>;
```

packages/react-on-rails-pro-node-renderer/src/worker/handleIncrementalRenderStream.ts
```diff
@@ -0,0 +1,141 @@
...
+  };
+  onRenderRequestReceived: (renderRequest: unknown) => Promise<RenderRequestResult> | RenderRequestResult;
+  onResponseStart: (response: ResponseResult) => Promise<void> | void;
+  onUpdateReceived: (updateData: unknown) => Promise<void> | void;
```
alexeyr-ci2, line 23: Are the argument types really unknown here?

packages/react-on-rails-pro-node-renderer/src/worker/handleIncrementalRenderStream.ts
```diff
@@ -0,0 +1,141 @@
...
+            const errorMessage = `Invalid JSON chunk: ${err instanceof Error ? err.message : String(err)}`;
+
+            if (!hasReceivedFirstObject) {
+              // Error in first chunk - throw error to stop processing
+              throw new Error(errorMessage);
+            } else {
+              // Error in subsequent chunks - log and report but continue processing
+              const reportedMessage = `JSON parsing error in update chunk: ${err instanceof Error ? err.message : String(err)}`;
+              log.error({ msg: reportedMessage });
+              errorReporter.message(reportedMessage);
```
alexeyr-ci2:
```suggestion
              errorReporter.message(reportedMessage);
```
`errorReporter.message` already includes `log.error`. Same for other places.

packages/react-on-rails-pro-node-renderer/src/worker/vm.ts
```diff
@@ -101,95 +102,25 @@ function manageVMPoolSize() {
...
   }
 }
 
-export async function buildVM(filePath: string) {
+async function buildVM(filePath: string): Promise<VMContext> {
   // Return existing promise if VM is already being created
   if (vmCreationPromises.has(filePath)) {
-    return vmCreationPromises.get(filePath);
+    return vmCreationPromises.get(filePath) as Promise<VMContext>;
   }
```
alexeyr-ci2: This cast can be avoided:
```
const vmCreationPromise = vmCreationPromises.get(filePath);
if (vmCreationPromise) {
  return vmCreationPromise;
}
```
(or `return await` is slightly better as mentioned in https://github.com/tc39/proposal-faster-promise-adoption).

packages/react-on-rails-pro-node-renderer/src/worker/vm.ts
```diff
@@ -101,95 +102,25 @@ function manageVMPoolSize() {
...
 
   // Check if VM for this bundle already exists
   const vmContext = vmContexts.get(filePath);
   if (vmContext) {
     // Update last used time when accessing existing VM
     vmContext.lastUsed = Date.now();
-    return Promise.resolve(true);
+    return Promise.resolve(vmContext);
```
alexeyr-ci2:
```suggestion
    return vmContext;
```

packages/react-on-rails-pro-node-renderer/src/worker/vm.ts
```diff
@@ -347,6 +274,147 @@ export async function buildVM(filePath: string) {
...
+    return vmContext;
+  }
+
+  const vmCreationPromise = vmCreationPromises.get(bundleFilePath);
+  if (vmCreationPromise) {
+    return vmCreationPromise;
+  }
+
+  if (buildVmsIfNeeded) {
+    return buildVM(bundleFilePath);
```
alexeyr-ci2, line 290: `buildVM` will do the same checks as above, can we avoid it?

packages/react-on-rails-pro-node-renderer/src/worker.ts
```diff
@@ -272,17 +254,129 @@ export default function run(config: Partial<Config>) {
     }
   });
 
+  // Streaming NDJSON incremental render endpoint
+  app.post<{
+    Params: { bundleTimestamp: string; renderRequestDigest: string };
+  }>('/bundles/:bundleTimestamp/incremental-render/:renderRequestDigest', async (req, res) => {
```
alexeyr-ci2, line 258: This generic parameter should now be extracted.

packages/react-on-rails-pro-node-renderer/src/worker.ts
```diff
@@ -376,7 +489,9 @@ export default function run(config: Partial<Config>) {
     Querystring: { filename: string };
     Body: WithBodyArrayField<Record<string, unknown>, 'targetBundles'>;
   }>('/asset-exists', async (req, res) => {
-    if (!(await isAuthenticated(req, res))) {
+    const authResult = authenticate(req.body as AuthBody);
```
alexeyr-ci2, line 490: Needing to cast here is strange, should we update `Body` type above instead?

packages/react-on-rails-pro-node-renderer/tests/fixtures/bundle-incremental.js
```diff
@@ -0,0 +1,41 @@
+const { PassThrough } = require('stream');
+
+global.ReactOnRails = {
```
alexeyr-ci2, line 3: Why do we have different `ReactOnRails` values in these fixtures? This seems likely to drift away from the real value in the future so they won't test what they should.

AbanoubGhadban:
## Response to Review Comments (Alexey & CodeRabbit)

Thank you for the thorough reviews! I've addressed all the valid issues. Here's a summary:

---

### ✅ Issues Fixed

#### 1. **CRITICAL: AsyncPropsManager missing `resolved = true`** (CodeRabbit)
**Fixed in commit:** `1a3fce61f`
- Added `promiseController.resolved = true;` after `promiseController.resolve(propValue);`
- Now `endStream()` correctly skips already-resolved promises

#### 2. **Type mismatch: `onRequestClosedUpdateChunk`** (CodeRabbit)
**Fixed in commit:** `1a3fce61f`
- Changed type from `string` to `UpdateChunk`
- Updated `assertFirstIncrementalRenderRequestChunk` to reuse `assertIsUpdateChunk()` (DRY)
- Removed redundant assertion in `handleRequestClosed`

#### 3. **Redundant logging** (Alexey)
**Fixed in commit:** `002297ba0`
- `errorReporter.message()` already calls `log.error()` internally
- Removed duplicate `console.error`/`log.error` calls to avoid double logging

#### 4. **vm.ts code quality** (Alexey)
**Fixed in commit:** `bba60060d`
- Avoided type cast by using local variable for `vmCreationPromises.get()`
- Simplified `return Promise.resolve(vmContext)` to `return vmContext`

#### 5. **Duplicate constants** (Alexey)
**Fixed in commit:** `efbef6290`
- Created `src/shared/constants.ts` with `BODY_SIZE_LIMIT` and `FIELD_SIZE_LIMIT`
- Updated `worker.ts` and `handleIncrementalRenderStream.ts` to use shared constants

---

### ⏭️ Intentionally Not Changed

#### **Missing `onRequestEnded()` on early return** (CodeRabbit)
This is **intentional behavior**. When `shouldContinue` is false, we explicitly don't want to call `onRequestEnded()` because the request handling is being short-circuited.

---

### ❌ False Positives / Already Fixed

| Issue | Status |
|-------|--------|
| NDJSON buffer flush | Not a problem - Ruby always sends trailing newlines |
| httpx version >= 1.7.0 | Gemspec shows `~> 1.5`, CodeRabbit reviewed old version |
| VM lazy build suggestion | Not needed - `dependencyBundleTimestamps` explicitly lists all bundles |
| Test fixture defensive checks | Minor - test fixtures can assume environment is set up |

---

### Tests
All relevant tests pass:
- `AsyncPropManager.test.ts` ✅
- `vm.test.ts` ✅
- `incrementalRender.test.ts` ✅
- `handleIncrementalRenderStream.test.ts` ✅

(2 flaky timing tests in `htmlStreaming.test.js` failed due to system load, unrelated to these changes)

PR #1857, opened 2025-10-08
Add offline JWT-based license validation system for React on Rails Pro

AbanoubGhadban:
## Summary
- Implements a pure offline license validation system using JWT tokens signed with RSA-256
- No internet connectivity required for validation
- Validates licenses in Ruby gem and Node renderer, browser trusts server validation

## Motivation
React on Rails Pro packages need a license validation system to ensure only licensed users can access Pro features. This implementation provides:
- **Security**: JWT tokens with RSA-256 signatures that cannot be forged
- **Offline**: No internet connection required, works in air-gapped environments
- **Simple**: Just valid/invalid status, no complex feature flags
- **Developer-friendly**: Warnings in development, errors in production

### Core Implementation
- **Ruby License Validation**
  - `license_validator.rb` - Main validation logic with JWT verification
  - `license_public_key.rb` - RSA public key for signature verification
  - Added `jwt` gem dependency to gemspec

- **Node Renderer Validation**
  - `licenseValidator.ts` - TypeScript validation module
  - `licensePublicKey.ts` - Public key for Node
  - Added `jsonwebtoken` npm dependency
  - Integrated validation at Node renderer startup

### Integration Points
- Updated `react_on_rails_pro_licence_valid?` to use new validator
- Modified Rails context to use license validation for `rorPro` field  
- Node renderer validates on startup (exits if invalid in production)

### Testing & Documentation
- Comprehensive Ruby specs covering all validation scenarios
- Jest tests for Node/TypeScript validation
- `LICENSE_SETUP.md` with complete setup instructions
- Added license file to `.gitignore`

## How It Works

```
1. License JWT → ENV['REACT_ON_RAILS_PRO_LICENSE'] or config/react_on_rails_pro_license.key
2. Ruby gem validates → Sets railsContext.rorPro
3. Node renderer validates → Continues or exits based on environment
4. Browser trusts server → Uses railsContext.rorPro
```

## License Configuration

Users can provide their license via:
- **Environment variable**: `REACT_ON_RAILS_PRO_LICENSE="jwt_token"`
- **Config file**: `config/react_on_rails_pro_license.key`

## Next Steps

1. Replace placeholder public keys with ShakaCode's actual public key
2. Set up license generation infrastructure (private key, JWT issuer)
3. Test with real licenses
4. Deploy to users

## Test Plan

- [x] Ruby specs pass for all validation scenarios
- [x] Node tests cover validation logic
- [x] Files properly ignored in git
- [ ] Manual testing with real JWT tokens
- [ ] Integration testing with Pro features


lib/react_on_rails/pro_utils.rb
```diff
@@ -5,9 +5,9 @@ module ProUtils
     PRO_ONLY_OPTIONS = %i[immediate_hydration].freeze
 
     # Checks if React on Rails Pro features are available
-    # @return [Boolean] true if Pro license is valid, false otherwise
+    # @return [Boolean] true if Pro is installed and licensed, false otherwise
     def self.support_pro_features?
-      ReactOnRails::Utils.react_on_rails_pro_licence_valid?
+      ReactOnRails::Utils.react_on_rails_pro?
```
AbanoubGhadban, line 10: `react_on_rails_pro_licence_valid?` has been removed, as the app should not start unless the license is valid. Therefore, `react_on_rails_pro?` now checks for license validation.

react_on_rails_pro/lib/react_on_rails_pro/license_validator.rb
```diff
@@ -0,0 +1,118 @@
...
+      def license_data
+        @license_data ||= load_and_decode_license
+      end
+
+      attr_reader :validation_error
+
+      private
+
+      def validate_license
+        license = load_and_decode_license
```
alexeyr-ci2: Should this use `license_data` instead, to avoid potentially decoding the data twice?

react_on_rails_pro/lib/react_on_rails_pro/license_validator.rb
```diff
@@ -0,0 +1,118 @@
...
+        handle_invalid_license(@validation_error)
+      end
+
+      def load_and_decode_license
+        license_string = load_license_string
+
+        JWT.decode(
+          license_string,
+          public_key,
+          true,
```
alexeyr-ci2, line 143: Add a comment with the meaning of the argument?

react_on_rails_pro/lib/react_on_rails_pro/license_validator.rb
```diff
@@ -0,0 +1,118 @@
...
+        # First try environment variable
+        license = ENV.fetch("REACT_ON_RAILS_PRO_LICENSE", nil)
+        return license if license.present?
+
+        # Then try config file
+        config_path = Rails.root.join("config", "react_on_rails_pro_license.key")
+        return File.read(config_path).strip if config_path.exist?
+
+        @validation_error = "No license found. Please set REACT_ON_RAILS_PRO_LICENSE environment variable " \
+                            "or create config/react_on_rails_pro_license.key file. " \
```
alexeyr-ci2:
```suggestion
                            "or create #{config_path} file. " \
```
(just in case we change the expected path later, so it's only in a single place)

react_on_rails_pro/packages/node-renderer/src/shared/licenseValidator.ts
```diff
@@ -0,0 +1,192 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import { PUBLIC_KEY } from './licensePublicKey';
+
+interface LicenseData {
+  sub?: string; // Subject (email for whom the license is issued)
+  iat?: number; // Issued at timestamp
+  exp: number; // Required: expiration timestamp
+  plan?: string; // Optional: license plan (e.g., "free", "paid")
+  issued_by?: string; // Optional: who issued the license
```
alexeyr-ci2: 1. For the current version, I expect `iat` to end up with the `Subject...` comment in the IDEs, which could be fixed by moving the comments
    ```
    // Subject (email for whom the license is issued)
    sub?: string;
    ```
    etc. Maybe I'm wrong?
2. In the standard JWT payload we already have `iss` for `issuer`, either replacing `issued_by` by `iss`, or making clearer names for `sub`/`iat`/`exp` would make it more consistent.

react_on_rails_pro/packages/node-renderer/src/shared/licenseValidator.ts
```diff
@@ -0,0 +1,192 @@
...
+  // Allow additional fields
+  [key: string]: unknown;
```
alexeyr-ci2, line 18: Is this needed?

react_on_rails_pro/packages/node-renderer/src/shared/licenseValidator.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/**
+ * Performs the actual license validation logic.
+ * @private
+ */
+// eslint-disable-next-line consistent-return
+function performValidation(): boolean | never {
+  try {
+    const license = loadAndDecodeLicense();
```
alexeyr-ci2: Similar to the Ruby side, we may want to call `getLicenseData` to make sure it's only decoded once.

react_on_rails_pro/packages/node-renderer/src/shared/licenseValidator.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (!license.exp) {
+      cachedValidationError =
+        'License is missing required expiration field. ' +
+        'Your license may be from an older version. ' +
+        'Get a FREE evaluation license at https://shakacode.com/react-on-rails-pro';
+      handleInvalidLicense(cachedValidationError);
+    }
+
+    // Check expiry
+    if (Date.now() / 1000 > license.exp) {
```
alexeyr-ci2: Add a comment explaining why `/ 1000`?

react_on_rails_pro/packages/node-renderer/src/shared/licenseValidator.ts
AbanoubGhadban: Licenses can contain additional fields, including organization, etc. Therefore, I incorporated this into the TypeScript code so that anyone working with it will be informed.

react_on_rails_pro/spec/dummy/spec/helpers/react_on_rails_pro_helper_spec.rb
```diff
@@ -13,7 +13,7 @@ def render_to_string(*args); end
   def response; end
 end
 
-describe ReactOnRailsProHelper, type: :helper do
```
AbanoubGhadban, line 16: The `type: :helper` metadata was removed by RuboCop (`RSpecRails/InferredSpecType cop`). RSpec Rails automatically infers the spec type from the file path (`spec/helpers/`), so the explicit type metadata is redundant and was removed.

PR #1868, opened 2025-10-20
Add performance tests

### LINKED ISSUE #1777: Add performance tests
Add performance tests for the dummy app using K6

alexeyr-ci2:
### Summary

Add performance tests. Fixes #1777.

### Pull Request checklist

- [ ] Update documentation
- [ ] Update CHANGELOG file

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1868)

## Summary by CodeRabbit

* **New Features**
  * End-to-end performance benchmarking: local scripts, k6/vegeta workloads, HTML report splitting, and CI-driven benchmark reporting with regression detection.

* **Documentation**
  * Added benchmarking strategy and CONTRIBUTING guidance; changelogs updated with benchmarking notes.

* **Chores**
  * CI benchmark workflow, tooling and test-run helpers, ignore rules for bench results, production/start/build helper scripts, and benchmark utility libraries.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -548,11 +548,17 @@ jobs:
             START_POINT_HASH="${{ github.event.pull_request.base.sha }}"
             EXTRA_ARGS="--start-point-reset"
           elif [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
-            # Find where branch diverged from master
-            git fetch origin master --depth=100
+            # Get merge-base from GitHub API (avoids needing deep fetch)
+            # See: https://stackoverflow.com/a/74710919
             BRANCH="${{ github.ref_name }}"
             START_POINT="master"
-            START_POINT_HASH=$(git merge-base HEAD origin/master)
+            START_POINT_HASH=$(gh api "repos/${{ github.repository }}/compare/master...$BRANCH" --jq '.merge_base_commit.sha' || true)
+
+            if [ -n "$START_POINT_HASH" ]; then
+              echo "Found merge-base via API: $START_POINT_HASH"
+            else
+              echo "⚠️ Could not find merge-base with master via GitHub API, continuing without it"
+            fi
             EXTRA_ARGS=""
           else
             echo "❌ ERROR: Unexpected event type: ${{ github.event_name }}"


alexeyr-ci2:
@coderabbitai review

alexeyr-ci2:
@coderabbitai Review.

benchmarks/lib/benchmark_helpers.rb
```diff
@@ -0,0 +1,124 @@
...
+  params.each do |key, value|
+    puts "  - #{key}: #{value}"
+  end
+end
+
+# Display summary using column command
+def display_summary(summary_file)
+  puts "\nSummary saved to #{summary_file}"
+  system("column", "-t", "-s", "\t", summary_file)
+end
```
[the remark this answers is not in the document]
alexeyr-ci2: There is already a newline there.

benchmarks/bench.rb
```diff
@@ -0,0 +1,224 @@
...
+  end
+  puts "Warm-up complete for #{route}"
+
+  route_name = sanitize_route_name(route)
+  metrics = run_k6_benchmark(target, route_name)
+  add_to_summary(route, *metrics)
+end
+
+puts "\nSummary saved to #{SUMMARY_TXT}"
+system("column", "-t", "-s", "\t", SUMMARY_TXT)
```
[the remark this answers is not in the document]
alexeyr-ci2: Again, it's there already.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,540 @@
...
+      - master
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+env:
```
ihabadham, line 62: should we add
```
concurrency:
    group: benchmark
    cancel-in-progress: true
```

to this? to prevent multiple simultaneous run all trying to post results to the same benchmark-data branch. see https://github.com/shakacode/react_on_rails/actions/runs/20232642214/job/58079207830 and https://github.com/shakacode/react_on_rails/actions/runs/20232643917/job/58079213536

benchmarks/convert_to_benchmark_json.rb
```diff
@@ -0,0 +1,173 @@
...
+    cols = line.split("\t")
+    row = header.zip(cols).to_h
+
+    # Determine the name based on available columns
+    name = row["Route"] || row["Test"] || "unknown"
+    bundle_suffix = row["Bundle"] ? " (#{row['Bundle']})" : ""
+    full_name = "#{prefix}#{name}#{bundle_suffix}"
+
+    # Skip if we got FAILED values
+    next if row["RPS"] == "FAILED"
```
ihabadham: should we also skip missing values?
`next if row["RPS"] == "missing"`
so they don't get silently converted to 0.0 next which would be misleading

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,540 @@
...
+          echo "✅ Server stopped"
+
+      # ============================================
+      # STEP 7: STORE BENCHMARK DATA
+      # ============================================
+      - name: Store all benchmark results
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: React on Rails Benchmarks
+          tool: customSmallerIsBetter
```
ihabadham: Instead of negating RPS to unify everything, why not use customBiggerIsBetter for it? The current approach completely breaks the alert system for RPS ratio calculations.

Example: 500 → 400 RPS (20% regression), stored as -500 → -400.

[github-action-benchmark calculates](https://github.com/benchmark-action/github-action-benchmark/blob/master/src/write.ts#L547-L552): ratio = current / prev = -400 / -500 = 0.8

Since 0.8 < 1.5 threshold, no alert fires - but this is a regression that should alert!

Conversely, improvements trigger false alerts: 500 → 1000 RPS stored as -500 → -1000, ratio = -1000 / -500 = 2.0 > 1.5 → alert fires incorrectly.

The original two-file approach in commit [f0f2692d](https://github.com/shakacode/react_on_rails/pull/1868/changes/f0f2692d3e1d44e4ef3c13b6e9d4aee4e3106c27) was correct (benchmark_rps.json with customBiggerIsBetter, benchmark_latency.json with customSmallerIsBetter). The Git conflict issue from multiple benchmark steps could be solved by using different benchmark-data-dir-path values instead of negating RPS.

benchmarks/convert_to_benchmark_json.rb
```diff
@@ -0,0 +1,173 @@
...
+
+# Parse Node Renderer benchmark
+node_renderer_summary = File.join(BENCH_RESULTS_DIR, "node_renderer_summary.txt")
+if File.exist?(node_renderer_summary)
+  all_results.concat(parse_summary_file(node_renderer_summary, prefix: "#{PREFIX}NodeRenderer: "))
+end
+
+if all_results.empty?
+  puts "No benchmark results found to convert"
+  exit 0
```
ihabadham:
```suggestion
  puts "ERROR: All benchmarks failed - no valid results to convert"
  exit 1
```

When all benchmarks fail, all_results is empty (FAILED rows are skipped at line 47). The script prints "No benchmark results found" and exits 0 without creating benchmark.json. Then github-action-benchmark runs and fails with a confusing "file not found" error. The clear message gets buried in logs.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,540 @@
...
+              echo "✅ Port 3001 is now free"
+              exit 0
+            fi
+            sleep 1
+          done
+
+          echo "❌ ERROR: Port 3001 is still in use after 10 seconds"
+          echo "Processes using port 3001:"
+          lsof -i:3001 || true
+          exit 1
```
ihabadham, line 369: This `exit 1` can fail a successful benchmark job. If Pro needs the port, failure will surface when Pro can't start. If running core_only, this fails the job unnecessarily.

.github/workflows/benchmark.yml
alexeyr-ci2: I'll see if different `benchmark-data-dir-path`s work, but I think it was unhappy just about the new commit being present. I agree your issue is correct, but if this doesn't work, the alternative is switching to 1/RPS, which should avoid it. I reported https://github.com/benchmark-action/github-action-benchmark/issues/328 but the author didn't respond; maybe I'll make a PR instead, it should be easy enough and the author ignores _some_ of them but merges other.

.github/workflows/benchmark.yml
alexeyr-ci2: The reason this was added is that Pro was reporting a successful start even though it actually was using the unstopped Core server. I'll try to see how to fix the issue there, but at the least this should only fail if `$RUN_PRO` is true.

.github/workflows/benchmark.yml
alexeyr-ci2: The issues there are not from simultaneous runs but from trying to use the action twice in the same run (note both failed on the second benchmark). That is, same as discussed in https://github.com/shakacode/react_on_rails/pull/1868#discussion_r2639874471.

benchmarks/convert_to_benchmark_json.rb
alexeyr-ci2: Probably, will fix (maybe check if it can be parsed as a number instead).

.github/workflows/benchmark.yml
alexeyr-ci2: Added the condition, though not pushed yet because there will be other changes to replace github-action-benchmark.

alexeyr-ci2: Fixed.

benchmarks/convert_to_benchmark_json.rb
alexeyr-ci2: Fixed.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,668 @@
...
+        run: |
+          # CPU pinning: server gets CPUs 1-(nproc-1), benchmark script gets CPU 0
+          # This avoids contention and improves consistency on GitHub Actions runners
+          NPROC=$(nproc)
+          echo "Available CPUs: $NPROC"
+
+          # Server: high priority (-20), pinned to CPUs 1+
+          # Benchmark: elevated priority (-10), pinned to CPU 0
+          SERVER_CMD="nice -n -20 taskset -c 1-$((NPROC-1)) bin/prod"
+          BENCH_CMD="nice -n -10 taskset -c 0 ruby"
```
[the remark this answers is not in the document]
alexeyr-ci2: This is wrong, it works on GitHub-hosted runners.

benchmarks/split_html_report.rb
```diff
@@ -0,0 +1,90 @@
...
+    if current_size + row_size > MAX_CHUNK_SIZE && current_rows.any?
+      # Save current chunk
+      chunks << build_chunk(header, current_rows, footer)
+      current_rows = []
+      current_size = chunk_overhead
+    end
+
+    current_rows << row
+    current_size += row_size
+  end
```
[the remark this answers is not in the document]
alexeyr-ci2: Not a realistic concern.

.github/workflows/benchmark.yml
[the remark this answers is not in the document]
alexeyr-ci2: This runs successfully in the CI environment.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,668 @@
...
+            START_POINT="$GITHUB_BASE_REF"
+            START_POINT_HASH="${{ github.event.pull_request.base.sha }}"
+            EXTRA_ARGS="--start-point-reset"
+          elif [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            # Find where branch diverged from master
+            git fetch origin master --depth=100
+            BRANCH="${{ github.ref_name }}"
+            START_POINT="master"
+            START_POINT_HASH=$(git merge-base HEAD origin/master)
+            EXTRA_ARGS=""
```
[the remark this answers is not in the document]
alexeyr-ci2: Should be fixed now.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ After a release, run `/update-changelog` in Claude Code to analyze commits, writ
 
 Changes since the last non-beta release.
 
+#### Developer (Contributors Only)
+
+- **Benchmarking in CI**: A benchmark workflow will now run on all pushes to master, as well as PRs with `benchmark` or `full-ci` labels. [PR 1868](https://github.com/shakacode/react_on_rails/pull/1868) by [alexeyr-ci2](https://github.com/alexeyr-ci2)
+
```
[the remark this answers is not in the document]
alexeyr-ci2: We have other "Developer (Contributors Only)" entries. In this case, knowing about the labels is useful.

PR #1925, opened 2025-11-05
Fix Node.js 22 V8 cache crash by reverting to Node 20

justin808:
## Summary

The previous attempt to disable yarn caching for Node 22 using `cache: ''` caused CI failures because GitHub Actions setup-node doesn't accept empty strings for the cache parameter and still attempts to initialize caching.

This PR reverts all workflows to use Node 20 until the V8 bug is resolved (https://github.com/nodejs/node/issues/56010).

## Key Changes

- Revert `lint-js-and-ruby.yml` from Node 22 to Node 20 with normal caching
- Revert `main.yml` matrix to use Node 20 for both latest and minimum dependency levels  
- Simplify `package-js-tests.yml` to only test Node 20
- Remove conditional cache logic that was causing failures
- Remove conditional frozen-lockfile flag that depended on Node 22

## CI Failures Fixed

This PR addresses the following CI failures:
- https://github.com/shakacode/react_on_rails/actions/runs/19094420849/job/54551179246
- https://github.com/shakacode/react_on_rails/actions/runs/19094420849/job/54551179250
- https://github.com/shakacode/react_on_rails/actions/runs/19094420881/job/54551179365

## Test Plan

- [x] RuboCop passes locally
- [x] Git hooks pass (prettier, trailing newlines check)
- [ ] CI workflows pass with Node 20 configuration

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1925)

### DIFF
--- a/.github/workflows/lint-js-and-ruby.yml
+++ b/.github/workflows/lint-js-and-ruby.yml
@@ -54,11 +54,8 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 22
-          # TODO: Re-enable cache when Node.js 22 V8 bug is fixed
-          # Disable cache for Node 22 due to V8 bug in 22.21.0
-          # Track: https://github.com/nodejs/node/issues/56010
-          cache: ''
+          node-version: 20
+          cache: 'yarn'
           cache-dependency-path: '**/yarn.lock'
       - name: Print system information
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
         include:
           # Always run: Latest versions (fast feedback on PRs)
           - ruby-version: '3.4'
-            node-version: '22'
+            node-version: '20'
             dependency-level: 'latest'
           # Master only: Minimum supported versions (full coverage)
           - ruby-version: '3.2'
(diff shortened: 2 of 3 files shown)


justin808:
Closing this PR as the **current workaround is working well** and reverting to Node 20 is not the best solution.

## Current State ✅

Master is successfully using **Node 22 with disabled cache workaround**:
- Recent CI runs are **passing consistently**
- Lint workflow: ✅ Success
- Main test workflow: ✅ Success  
- Package tests: ✅ Success

The workaround (disabling cache for Node 22) is effective and stable.

## Why Not Revert to Node 20

1. **Testing Latest Node Version** - Node 22 is the current LTS and we should test against it
2. **Minimal Performance Impact** - Disabled cache slightly slows CI but it's acceptable
3. **Bug May Be Resolved** - The V8 cache bug was specific to Node 22.5.0/22.21.0; newer Node 22.x versions may have fixes
4. **Current Workaround Works** - No need to downgrade when we have a working solution

## Better Strategy

Instead of reverting to Node 20:
- ✅ **Keep current workaround** (Node 22 with disabled cache)
- 🔍 **Periodically test** if cache can be re-enabled with newer Node 22.x versions
- 📊 **Monitor** the upstream Node.js issue for V8 cache fixes
- ⏰ **Re-enable cache** when the bug is confirmed fixed

## Related

The workaround was implemented in these commits:
- 43508580 - Initial fix for Node.js 22 V8 cache crash
- a1c71ea1 - PR #1922 Fix Node.js 22 V8 crash in GitHub Actions workflows

Thanks for creating this PR, but the current approach (disabled cache) is preferable to downgrading Node versions.

PR #1841, opened 2025-09-29
PR #4: Split JS Pro Code to Separate Package

AbanoubGhadban:
## Summary

This PR implements the separation of JavaScript Pro functionality from the core React-on-Rails package into a separate `react-on-rails-pro` package, as outlined in Phase 3 of the monorepo merger plan.

## Architecture Decisions

- **Package Strategy**: Pro package uses core as dependency (not peer dependency)
- **Versioning**: Caret range strategy (`^16.1.0`) following React model
- **Registry Architecture**: Dual system with direct imports (MIT→MIT, Pro→Pro)
- **Code Reuse**: Pro features layered over Core functionality for DRY

## Implementation Plan

Created comprehensive implementation plan in `docs/JS_PRO_PACKAGE_SEPARATION_PLAN.md` with:
- 10 major steps broken into 30+ checkpoints
- Detailed success criteria and validation steps
- Comprehensive testing strategy
- Rollback procedures

### New Structure:
```
packages/
├── react-on-rails/ (MIT - Core functionality)
│   ├── src/ComponentRegistry.ts (simple sync)
│   ├── src/StoreRegistry.ts (simple sync) 
│   └── src/ClientRenderer.ts (simple sync)
└── react-on-rails-pro/ (Pro - Enhanced functionality)
    ├── src/ComponentRegistry.ts (advanced async)
    ├── src/StoreRegistry.ts (advanced async)
    ├── src/ClientSideRenderer.ts (advanced async)
    └── src/index.ts (exports all core + pro)
```

### User Experience:
- **Core users**: `import ReactOnRails from 'react-on-rails'` (sync features only)
- **Pro users**: `import ReactOnRails from 'react-on-rails-pro'` (all features)

## Implementation Status

✅ Comprehensive implementation plan created  
⏳ Step-by-step implementation in progress

## License Compliance

- Core package: MIT licensed (simple registries, basic features)
- Pro package: Pro licensed (advanced registries, async features, immediate hydration)
- Clean separation maintained throughout

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1841)

## Summary by CodeRabbit

* **New Features**
  * Adds a standalone Pro package with server-component support, streaming helpers, Pro client/server entry points, and utility helpers; new client rendering APIs (render single, render all, component-loaded promise) and automatic unmount/cleanup.
  * Public component and store registries for runtime registration and hydration.

* **Refactor**
[...]

PR #1988, opened 2025-11-12
Only initiate hydration after all subresources loaded

Judahmeek:
Since async hydration is only supported by ReactOnRails Pro now that [the Callback Registry has been moved to ReactOnRails Pro only](https://github.com/shakacode/react_on_rails/pull/1841/files), we need to either [revert back to ReactOnRails only hydrating after all subresources are loaded](https://github.com/shakacode/react_on_rails/pull/1540/files) or return the CallbackRegistry & async support back to ReactOnRails in order to prevent component registration race conditions like [this one](https://github.com/shakacode/react_on_rails/actions/runs/19276955847/job/55119074912#step:24:1156).

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/1988)

## Summary by CodeRabbit

* **Refactor**
  * Page lifecycle now uses a readiness-driven flow to wait for full document readiness before attaching navigation listeners, ensuring initialization occurs at the correct time.

* **Tests**
  * Updated tests to use readiness-driven semantics: callbacks run immediately when the document is already ready or otherwise wait for the readiness event.
  * Removed tests and shared setup that relied on pro-mode immediate hydration.

### DIFF
--- a/packages/react-on-rails/tests/pageLifecycle.test.js
+++ b/packages/react-on-rails/tests/pageLifecycle.test.js
@@ -77,12 +77,12 @@ describe('pageLifecycle', () => {
 
     // Since no navigation library is mocked, callbacks should run immediately
     expect(callback).toHaveBeenCalledTimes(1);
-    // Should not add DOMContentLoaded listener since readyState is not 'loading'
+    // Should not add readystatechange listener since readyState is not 'loading'
     expect(addEventListenerSpy).not.toHaveBeenCalledWith('readystatechange', expect.any(Function));
   });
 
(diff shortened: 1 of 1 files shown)


Judahmeek:
I disagree with Claude & CodeRabbit regarding the recursion. I don't care how many event listeners are created. I only care that they don't actually do anything once `isPageLifecycleInitialized` is `true`.

Judahmeek:
Closed as shouldn't be necessary if all deferred javascript runs before the DOMContentLoaded event, ass specifications insist.

PR #2023, opened 2025-11-16
Fix bin/dev --verbose option causing error

### LINKED ISSUE #1849: bin/dev --verbose option causes error despite being listed in help
## Description

The `--verbose` option is documented in the `bin/dev --help` output but causes an error when used.

## Steps to Reproduce

1. Run `bin/dev --help` - it shows:
   ```
   ⚙️  OPTIONS:
     --route ROUTE        Specify route to display in URLs (default: root)
     --rails-env ENV      Override RAILS_ENV for assets:precompile step only (prod mode only)
     --verbose, -v        Enable verbose output for pack generation
   ```

2. Try to use the option: `bin/dev --verbose`

## Expected Behavior

The command should run with verbose output enabled.

## Actual Behavior

```
/Users/justin/.local/share/mise/installs/ruby/3.4.3/lib/ruby/gems/3.4.0/gems/react_on_rails-16.1.1/lib/react_on_rails/dev/server_manager.rb:139:in 'ReactOnRails::Dev::ServerManager.run_from_command_line': invalid option: --verbose (OptionParser::InvalidOption)
	from bin/dev:34:in '<main>'
```

## Environment

- react_on_rails version: 16.1.1
- Ruby version: 3.4.3

## Suggested Fix

Either:
1. Implement the `--verbose` option in the ServerManager's OptionParser, or
2. Remove it from the help documentation if it's not intended to be supported

justin808:
## Summary
- Implements the `--verbose`/`-v` flag that was documented but not functional
- Fixes OptionParser to properly handle the verbose flag
- Updates all start() calls to pass through the verbose option

## What was the problem?
The `--verbose` option was shown in `bin/dev --help` output but caused an `OptionParser::InvalidOption` error when used. This happened because:
1. The help text documented `--verbose, -v`
2. The `start()` method accepted a `verbose:` parameter
3. But the OptionParser in `run_from_command_line()` didn't parse the flag
4. All calls to `start()` hardcoded `verbose: false`

## Changes
- Added `verbose: false` to the options hash initialization
- Implemented `--verbose`/`-v` flag parsing in OptionParser
- Updated three `start()` calls to use `options[:verbose]` instead of hardcoded `false`
- Added RuboCop disable/enable comments for AbcSize metric (slightly increased by parameter handling)

## Test Plan
- ✅ Manually tested `--verbose` and `-v` flags with all commands
- ✅ All existing RSpec tests pass
- ✅ RuboCop passes with no violations
- ✅ Pre-commit and pre-push hooks pass

## Testing
```bash
# All of these now work without error:
bin/dev --verbose
bin/dev -v
bin/dev static --verbose
bin/dev prod -v
```

Closes #1849

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/2023)

### DIFF
--- a/lib/react_on_rails/dev/server_manager.rb
+++ b/lib/react_on_rails/dev/server_manager.rb
@@ -144,10 +144,11 @@ module ReactOnRails
           puts help_troubleshooting
         end
 
+        # rubocop:disable Metrics/AbcSize
         def run_from_command_line(args = ARGV)
           require "optparse"
 
-          options = { route: nil, rails_env: nil }
+          options = { route: nil, rails_env: nil, verbose: false }
 
           OptionParser.new do |opts|
             opts.banner = "Usage: dev [command] [options]"
@@ -158,6 +159,10 @@ module ReactOnRails
 
             opts.on("--rails-env ENV", "Override RAILS_ENV for assets:precompile step only (prod mode only)") do |env|
               options[:rails_env] = env
+            end
+
+            opts.on("-v", "--verbose", "Enable verbose output for pack generation") do
+              options[:verbose] = true
             end
 
             opts.on("-h", "--help", "Prints this help") do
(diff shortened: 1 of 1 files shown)


justin808:
@claude review this PR

PR #2029, opened 2025-11-16
Fix spec/dummy bin/dev to use correct route

justin808:
## Summary

Replaces the `spec/dummy/bin/dev` symlink with a wrapper script that passes `--route=/` to the base dev script.

## Problem

The `spec/dummy/bin/dev` was a symlink to the template script at `lib/generators/react_on_rails/templates/base/base/bin/dev`, which defaults to `DEFAULT_ROUTE = "hello_world"`. This is incorrect for the dummy app which needs to use the root route `/`.

## Solution

Replace the symlink with a wrapper script that:
- Calls the base dev script
- Forces `--route=/` for the dummy app
- Passes through all other command-line arguments

## Test Plan

- Verified `bundle exec rubocop` passes with no offenses
- Verified `bin/dev --help` works correctly in spec/dummy
- Verified file is executable and has correct permissions

Generated with Claude Code (https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react_on_rails/2029)

## Summary by CodeRabbit

* **Chores**
  * Updated internal development script configuration to improve development environment setup.

### DIFF
--- a/react_on_rails_pro/spec/dummy/bin/dev
+++ b/react_on_rails_pro/spec/dummy/bin/dev
@@ -2,4 +2,4 @@
 # frozen_string_literal: true
 
 # This script calls the base dev script with a fixed route for the dummy app
-exec File.expand_path("../../../../lib/generators/react_on_rails/templates/base/base/bin/dev", __dir__), "--route=/", *ARGV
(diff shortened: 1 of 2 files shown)


justin808:
@claude review this PR

PR #2136, opened 2025-11-26
Add suppress_unused_component_warnings config option

### LINKED ISSUE #2117: Ability to suppress `console.warn`s
A lot of console warnings are logged to the browser's console when a registered bundle is not used on a loaded page. 

We have a codebase where many React components are registered with `ReactOnRails.register()`, but not all the registered components are used on every page. The components which are not used on a particular page throw a warning in the browser's console to suggest cleaning them up. While I appreciate the intent behind having these messages during development, they're very noisy in production environments where one gets a ton of them.

It would be nice to have the ability to turn off console logging via a configuration option or to suppress them automatically in production environments.

_None_

**Additional context**
_None_

justin808:
## Summary

- Adds `suppress_unused_component_warnings` configuration option to disable console warnings about registered components that are not used on a page
- This is useful in production environments where many React components are registered globally but only a subset are used on each page
- Set `config.suppress_unused_component_warnings = true` in your React on Rails initializer to suppress these warnings

## Changes

- Added `suppress_unused_component_warnings` config option to Ruby configuration (default: false)
- Pass option to railsContext as `suppressUnusedComponentWarnings`
- Updated CallbackRegistry in Pro package to check the option before logging warnings
- Updated TypeScript types (`RailsContext`) and RBS signatures
- Added documentation in `docs/api-reference/configuration.md`
- Updated test fixtures to include the new property

## Test plan

- [x] Run `bundle exec rubocop` - passed
- [x] Run `bundle exec rake run_rspec:gem` - 290 examples, 0 failures
- [x] Run `bundle exec rake rbs:validate` - passed
- [x] Run `pnpm run type-check` - passed for both react-on-rails and react-on-rails-pro packages

Closes #2117

### DIFF
--- a/react_on_rails_pro/spec/dummy/spec/helpers/react_on_rails_pro_helper_spec.rb
+++ b/react_on_rails_pro/spec/dummy/spec/helpers/react_on_rails_pro_helper_spec.rb
@@ -323,7 +323,7 @@ describe ReactOnRailsProHelper do
     end
     let(:rails_context_tag) do
       <<-SCRIPT.strip_heredoc
-        <script type="application/json" id="js-react-on-rails-context">{"componentRegistryTimeout":5000,"railsEnv":"test","inMailer":false,"i18nLocale":"en","i18nDefaultLocale":"en","rorVersion":"#{ReactOnRails::VERSION}","rorPro":true,"rorProVersion":"#{ReactOnRailsPro::VERSION}","rscPayloadGenerationUrlPath":"rsc_payload/","href":"http://foobar.com/development","location":"/development","scheme":"http","host":"foobar.com","port":null,"pathname":"/development","search":null,"httpAcceptLanguage":"en","somethingUseful":null,"serverSide":false}</script>
+        <script type="application/json" id="js-react-on-rails-context">{"componentRegistryTimeout":5000,"suppressUnusedComponentWarnings":false,"railsEnv":"test","inMailer":false,"i18nLocale":"en","i18nDefaultLocale":"en","rorVersion":"#{ReactOnRails::VERSION}","rorPro":true,"rorProVersion":"#{ReactOnRailsPro::VERSION}","rscPayloadGenerationUrlPath":"rsc_payload/","href":"http://foobar.com/development","location":"/development","scheme":"http","host":"foobar.com","port":null,"pathname":"/development","search":null,"httpAcceptLanguage":"en","somethingUseful":null,"serverSide":false}</script>
       SCRIPT
     end
     let(:react_component_div_with_initial_chunk) do


justin808:
## Addressed Review Feedback

I've addressed the review comments in commit 60885767f:

### ✅ Fixed Issues

1. **CHANGELOG PR number corrected**: Changed `[PR 2125]` → `[PR 2136]`

2. **Added unit tests**: New tests in `react_on_rails/spec/react_on_rails/configuration_spec.rb`:
   - Test that default value is `false`
   - Test that it can be set to `true`
   - Test that it can be set to `false` explicitly

### ℹ️ Notes on Other Suggestions

**Re: default already in configuration hash**: The default is already present at lines 67-70 in `configuration.rb`:
```ruby
# Set to true to suppress warnings about registered components not being used on a page.
# This can be useful in production environments where many components are registered globally
# but only a subset are used on each page.
suppress_unused_component_warnings: false,
```

**Re: explicit null coalescing in TypeScript**: The current code `getRailsContext()?.suppressUnusedComponentWarnings` already returns `undefined` when context is unavailable, and the `if (!suppressWarnings)` check treats both `false` and `undefined`/`null` as "warnings enabled" which is the correct behavior. Adding `?? false` would be slightly more explicit but functionally identical.

---

Monitoring CI for the updated commit.

justin808:
## CI Fix Pushed

Pushed commit 76a642990 to fix the failing Pro integration test.

**Issue**: The test fixture in `react_on_rails_pro/spec/dummy/spec/helpers/react_on_rails_pro_helper_spec.rb` was checking for an exact match of the `rails_context` JSON structure, which didn't include the new `suppressUnusedComponentWarnings` field.

**Fix**: Added `suppressUnusedComponentWarnings:false` to the test fixture to match the updated rails_context structure.

---

**Note on other CI failure**: The "JS unit tests for Renderer package" workflow is a **pre-existing failure** on master (failing since Nov 23rd), not related to this PR.

justin808:
## Response to Code Review

Thank you for the thorough review! Here's my response to the observations:

### ✅ Observation #1: Default in `self.configuration` - **Already Addressed**

The default value is already present in `self.configuration` at lines 67-70:

```ruby
component_registry_timeout: DEFAULT_COMPONENT_REGISTRY_TIMEOUT,
# Set to true to suppress warnings about registered components not being used on a page.
# This can be useful in production environments where many components are registered globally
# but only a subset are used on each page.
suppress_unused_component_warnings: false,
generated_component_packs_loading_strategy: nil,
```

The diff may have truncated this section, but it was included in the initial implementation.

### ✅ Observations #2-5 - All Verified

- TypeScript types properly synchronized with `helper.rb`
- `getRailsContext()` is safely accessed after `onPageLoaded()` 
- `attr_accessor` properly includes the new config option
- RBS signatures are consistent with the implementation

### ℹ️ Re: Integration Test for Warning Suppression

Adding an integration test for the actual warning suppression behavior would require mocking the timeout and console.warn in the Pro package tests. While this is a good suggestion, the current implementation is straightforward enough that the unit tests provide sufficient coverage. The behavior is:
1. Check `suppressUnusedComponentWarnings` from `getRailsContext()`
2. If falsy, log warnings; if truthy, skip warnings

If desired, I can add such a test in a follow-up PR to avoid scope creep.

---

**Summary**: All items in the review are either already addressed or are nice-to-haves for future improvements. The PR is ready for merge.

justin808:
Closing b/c this one will get an attempt without AI.

PR #2163, opened 2025-12-03
Add loading time check on CI

AbanoubGhadban:
# Summary
- This PR started as an attempt to make CI use the `@size-limit/time` plugin to measure the execution time of each client bundle and ensure it doesn't exceed the base branch execution time by more than 20%.
- Later, found that the `@size-limit/time` plugin is not good at all at detecting that because of the runtime noise on CI.
- So, CI will only look for the `download time` of chunks at the browser for now, until finding another tool that run the code multiple time on CI and get average execution time to get rid of runtime noise.
- However, the following changes are made:
  - The feature of measring the execution time and comparing it to the base branch is already added to the `bin/compare-bundle-sizes` script, so if you can run it locally with a noise mitigated environment you can do that by changing the `running` values at `.size-limit.json` file to `true`.
  - Removed the `compare` command from the `scripts/bundle-size.mjs` script as it's not needed and used the default comparison behavior at the `size-limit` package.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -33,7 +33,7 @@ jobs:
             echo "skip=false" >> $GITHUB_OUTPUT
           fi
 
-  size:
+  check-bundle-size:
     needs: check-skip
     if: needs.check-skip.outputs.skip != 'true'
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


AbanoubGhadban:
Seems that `@size-limit/time` is not accurate at all in detecting the changes in runtime. We need to use another solution that the code multiple times and get the average runtime value to avoid the runtime noise. So, we only used the `@size-limit/time` for measuring the download time which seems that the plugin is good at measring it.

PR #2179, opened 2025-12-07
use right versions of react and react-on-rails-rsc package at react-on-rails-pro tests

AbanoubGhadban:
`react-on-rails-pro` package tests require the following versions of the packages:
- `react`: v19.0.0
- `react-dom`: v19.0.0
- `react-on-rails-rsc`: v19.0.2
Moved the `react-on-rails-rsc` package to the `react-on-rails-pro` dev dependencies instead of keeping it at root's package.json file to avoid making it use `react v18.0.0` at the root.

### DIFF
--- a/script/convert
+++ b/script/convert
@@ -45,6 +45,8 @@ gsub_file_content("../package.json", /,(\s*})/, "\\1")
 # Switch to minimum supported React version (React 18 since we removed PropTypes)
 gsub_file_content("../package.json", /"react": "[^"]*",/, '"react": "18.0.0",')
 gsub_file_content("../package.json", /"react-dom": "[^"]*",/, '"react-dom": "18.0.0",')
+gsub_file_content("../packages/react-on-rails-pro/package.json", /"react": "[^"]*",/, '"react": "18.0.0",')
+gsub_file_content("../packages/react-on-rails-pro/package.json", /"react-dom": "[^"]*",/, '"react-dom": "18.0.0",')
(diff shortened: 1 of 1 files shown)


.github/workflows/package-js-tests.yml
```diff
@@ -126,5 +126,7 @@ jobs:
           pnpm add -g yalc
       - name: Build Renderer package
         run: pnpm build
-      - name: Run JS unit tests for Renderer package
-        run: pnpm test
```
AbanoubGhadban, line 130: This command runs all tests, including react-on-rails-pro-node-renderer tests. We need to run react-on-rails and react-on-rails-pro package tests only as node renderer tests run on another workflow

packages/react-on-rails/package.json
```diff
@@ -59,13 +59,7 @@
   },
   "peerDependencies": {
     "react": ">= 16",
-    "react-dom": ">= 16",
```
AbanoubGhadban, line 62: not related to this PR, this peer dependency should be removed since we separated `react-on-rails` package into `react-on-rails` and `react-on-rails-pro`

.github/workflows/package-js-tests.yml
alexeyr-ci2: My preference would be to go ahead and remove them from that workflow instead, even if that would be in the last merger step otherwise. Any reason to prefer this approach?

PR #2197, opened 2025-12-09
Upgrade to react@19.2.1

AbanoubGhadban:
### Summary

Used `react-on-rails-rsc@19.2.1` and `react v19.2.1` as dev dependencies and as dependencies for the dummy apps. Updated the tests to work well with the new behavior. The change in the tests happens because v19.2.1 emits much more debugging chunks for debugging, so we need to merge chunks together before asserting their content.

### Next Step
Implement the react partial-prerendering (PPR) at react-on-rails-pro. This is a smaple of how to use the partial prerendering API at react@19.2.1

```tsx
import { Suspense, use } from 'react';
import { prerender } from 'react-dom/static';
import { resume } from 'react-dom/server';

const AsyncComponent = ({ id, promise }: { id: string, promise: Promise<string>}) => {
  console.log(`Awaiting "${id}"`);
  const value = use(promise);
  console.log(`Awaited "${id}"`);
  return <p>Value of Component {id} is "{value}"</p>
};

const delayedValue = (value: string, delayMs: number) => new Promise<string>((resolve) => {
  setTimeout(() => {
    resolve(value);
  }, delayMs);
})

const App = ({ value1, value2 }: { value1: string, value2: string }) => {
  const promise1 = delayedValue(value1, 1000);
  const promise2 = delayedValue(value2, 2000);
  console.log('Executing the main App')

  return (
    <div>
      <h1>App Header</h1>
      <ol>
        <li>
          <Suspense fallback={"Loading First Item..."}>
            <AsyncComponent id='First Component' promise={promise1} />
          </Suspense>
        </li>
        <li>
          <Suspense fallback={"Loading Second Item..."}>
            <AsyncComponent id='Second Component' promise={promise2} />
          </Suspense>
        </li>
      </ol>
    </div>
  )
}

const decoder = new TextDecoder();

const printReaderChunks = async (type: string, stream: ReadableStream<Uint8Array<ArrayBufferLike>>) => {
  const reader = stream.getReader();
  let done = false;
  while (!done) {
    const result = await reader.read();
    done = result.done;
    if (!done) {
      console.log(`${type} Chunk: ${decoder.decode(result.value)}`)
    }
  }
}

test('fake test', async() => {
  const controller = new AbortController();
  setTimeout(() => {
    controller.abort(new Error('eeeeee'));
  }, 1100);
  // @ts-ignore
  const { prelude, postponed } = await prerender(<App />, {
    signal: controller.signal,
    onError: (e) => {
      console.log(`Error: ${e}`);
    }
  });
  await printReaderChunks('prelude', prelude);

  if (!postponed) {
    throw new Error('postponed is null');
  }

  await delayedValue('', 1000);
  console.log("\n\n**************Resuming**************\n\n");

  const resumeStream = await resume(<App value1='' value2='' />, postponed);
  await printReaderChunks('Resume', resumeStream);
})
```

### Problem that wee need to focus on fixing

The React API is limited; you can't choose which components to prerender and which to resume later. You need to select a specific point at time and abort the rendering at it. Hence, all components that are resolved before that minute are prerendered, and all components that are not resolved yet are resumed later.

### Possible Solutions

- Abort rendering when user uses async props (assuming that the Async Props PR is merged)
```tsx
import { Suspense, use, ReactNode } from 'react';
import { prerender } from 'react-dom/static';
import { resume } from 'react-dom/server';

const AsyncComponent = ({ id, promise, children }: { id: string, promise: Promise<string>, children?: () => ReactNode }) => {
  console.log(`Awaiting "${id}"`);
  const value = use(promise);
  console.log(`Awaited "${id}"`);
  return (
    <div>
      <p>Value of Component {id} is "{value}"</p>
      {children?.()}
    </div>
  )
};

const delayedValue = (value: string, delayMs: number) => new Promise<string>((resolve) => {
  setTimeout(() => {
    resolve(value);
  }, delayMs);
})

type WithAsyncProps<Props extends Record<string, unknown>> = Props & {
  getAsyncProp: (propName: string) => Promise<string>;
}

const InnerComponent = ({ getAsyncProp }: WithAsyncProps<{}>) => {
  const asyncPropPromise = getAsyncProp('asyncProp1');

  return (
    <Suspense fallback={<p>Loading Inner Async Component...</p>}>
      <AsyncComponent id='async prop component' promise={asyncPropPromise} />
    </Suspense>
  )
}

const App = ({ value1, value2, getAsyncProp }: WithAsyncProps<{ value1: string, value2: string }>) => {
  const promise1 = delayedValue(value1, 1000);
  const promise2 = delayedValue(value2, 2000);
  console.log('Executing the main App')

  return (
    <div>
      <h1>App Header</h1>
      <ol>
        <li>
          <Suspense fallback={"Loading First Item..."}>
            <AsyncComponent id='First Component' promise={promise1} />
          </Suspense>
        </li>
        <li>
          <Suspense fallback={"Loading Second Item..."}>
            <AsyncComponent id='Second Component' promise={promise2}>
              {() => (
                <InnerComponent getAsyncProp={getAsyncProp} />
              )}
            </AsyncComponent>
          </Suspense>
        </li>
      </ol>
    </div>
  )
}

const decoder = new TextDecoder();

const printReaderChunks = async (type: string, stream: ReadableStream<Uint8Array<ArrayBufferLike>>) => {
  const reader = stream.getReader();
  let done = false;
  while (!done) {
    const result = await reader.read();
    done = result.done;
    if (!done) {
      console.log(`${type} Chunk: ${decoder.decode(result.value)}`)
    }
  }
}

test('fake test', async () => {
  const controller = new AbortController();
  // setTimeout(() => {
  //   controller.abort(new Error('eeeeee'));
  // }, 1100);

  const getAsyncProp = async (propName: string) => `Async Prop Value of "${propName}"`;
  const getAsyncPropPreludeImplementation = () => new Promise<string>((resolve) => {
    setTimeout(() => {
      controller.abort();
      resolve('');
    }, 0);
  });

  const { prelude, postponed } = await prerender(
    <App
      getAsyncProp={getAsyncPropPreludeImplementation}
      value1='First Unique Value'
      value2='Second Unique Value'
    />,
    {
      signal: controller.signal,
      onError: (e) => {
        console.log(`Error: ${e}`);
      }
    },
  );
  await printReaderChunks('prelude', prelude);

  if (!postponed) {
    throw new Error('postponed is null');
  }

  await delayedValue('', 1000);
  console.log("\n\n**************Resuming**************\n\n");

  const resumeStream = await resume(
    <App
      getAsyncProp={getAsyncProp}
      value1='First Unique Value'
      value2='Second Unique Value'
    />, postponed);
  await printReaderChunks('Resume', resumeStream);
}, 10000)
```

But it causes a problem that all components that use async props will be suspended and not preluded, instead it will be resumed later.

- Another solution, is letting user decide which async props to be suspended and which are provided on prelude. This makes us need to put a high timeout that get awaited before aborting the rendering, to ensure that all components that use the resolved async props are rendered before aborting the rendering.

### DIFF
--- a/packages/react-on-rails-pro/tests/concurrentRSCPayloadGeneration.rsc.test.tsx
+++ b/packages/react-on-rails-pro/tests/concurrentRSCPayloadGeneration.rsc.test.tsx
@@ -98,7 +98,7 @@ const createParallelRenders = (size: number) => {
 };
 
 test('Renders concurrent rsc streams as single rsc stream', async () => {
-  // expect.assertions(258);
+  expect.assertions(258);
   const asyncQueue = new AsyncQueue();
   const stream = renderComponent({ asyncQueue });
   const reader = new StreamReader(stream);


packages/react-on-rails-pro/tests/concurrentRSCPayloadGeneration.rsc.test.tsx
```diff
@@ -101,8 +98,8 @@ const createParallelRenders = (size: number) => {
 };
 
 test('Renders concurrent rsc streams as single rsc stream', async () => {
-  expect.assertions(258);
-  const asyncQueue = new AsyncQueue<string>();
+  // expect.assertions(258);
```
alexeyr-ci2: Let's delete, not uncomment. `expect.assertions` is useful in the first place when using callbacks instead of `async/await`, to check callbacks actually get called.
```suggestion
```

packages/react-on-rails-pro/tests/AsyncQueue.ts
```diff
@@ -1,22 +1,29 @@
 import * as EventEmitter from 'node:events';
 
-class AsyncQueue<T> {
-  private eventEmitter = new EventEmitter();
+const debounce = <T extends unknown[]>(callback: (...args: T) => void, delay: number) => {
```
alexeyr-ci2, line 3: Strange we don't already have it somewhere, but apparently not.

packages/react-on-rails-pro/tests/concurrentRSCPayloadGeneration.rsc.test.tsx
AbanoubGhadban: I commented it by mistake. At these types of tests, I have to put the number of assertions, so for any reason if the `createParallelRenders` is broken, the tests don't pass without noticing.

alexeyr-ci2: Can we show how it's calculated then, since 258 looks like a random number?

PR #2201, opened 2025-12-10
Fix broken pnpm lock file for dummy apps

AbanoubGhadban:
### Summary

Spec dummy apps (core and pro dummy apps) didn't set up well when we migrated to pnpm. `react_on_rails/spec/dummy` and `react_on_rails_pro/spec/dummy` apps were not part of the monorepo packages at `<root>/pnpm-workspace.yaml`. This configuration made some dependencies of dummy apps not installed. Also, it installed different versions of `react`, which caused a problem with the `react-redux` package.

Also, we currently use yalc to make dummy apps use the packages. So, I added `pnpm-workspace.yaml` file to both dummy apps, so each of them have its own `pnpm-lock.yaml` file and fix all problems. In the future, if we add the dummy apps to the workspace, we can get rid of the yalc and make the dummy apps use packages directly.

Seems that claude discovered that problem when it migrated to `pnpm`, but it made a work around at CI to make it pass. It used the `--ignore-workspace` argument while installing the dummy apps packages.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflows to improve dependency resolution and package linking for test environments.
  * Added workspace configuration files to standardize how dependencies are handled during builds and testing.
  * Refined how local packages are linked to ensure consistency across different dependency scenarios.


.github/workflows/integration-tests.yml
```diff
@@ -153,9 +153,8 @@ jobs:
         # Use --link to maintain the link: protocol that matches pnpm-lock.yaml
         run: cd react_on_rails/spec/dummy && yalc add --link react-on-rails
       - name: Install Node modules with pnpm for dummy app
-        # --ignore-workspace prevents pnpm from treating this as part of the parent workspace
         # minimum config changes package.json (shakapacker version), so can't use frozen lockfile
-        run: cd react_on_rails/spec/dummy && pnpm install --ignore-workspace ${{ matrix.dependency-level == 'minimum' && '--no-frozen-lockfile' || '' }}
+        run: cd react_on_rails/spec/dummy && pnpm install ${{ matrix.dependency-level == 'minimum' && '--no-frozen-lockfile' || '' }}
```
alexeyr-ci2: Probably not worth changing if there are no other changes, but I'd be explicit and use `--frozen-lockfile` instead of `''` even though it's the default in CI (similarly in other places).

.github/workflows/pro-integration-tests.yml
```diff
@@ -136,9 +136,8 @@ jobs:
           pnpm install --frozen-lockfile
 
       - name: Install Node modules with pnpm for Pro dummy app
-        # --ignore-workspace prevents pnpm from treating this as part of the parent workspace
-        # The Pro dummy app has its own dependencies and uses yalc links
-        run: cd spec/dummy && pnpm install --ignore-workspace
+        run: |
+          cd spec/dummy && pnpm install
```
alexeyr-ci2:
```suggestion
        run: cd spec/dummy && pnpm install
```
for consistency with the below changes.

react_on_rails/spec/dummy/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - '!**/*'
```
alexeyr-ci2: Should it just be `'.'`?

PR #2211, opened 2025-12-11
Fix hydration mismatch when reactOnRailsPageLoaded() called multiple times

### LINKED ISSUE #2210: Hydration mismatch with React 19 and the latest beta version of this gem
After upgrading from react_on_rails 16.1.2 to 16.2.0.beta.19, we're experiencing hydration mismatch errors even though we're using client-side only rendering (no prerender: true, server_bundle_js_file is empty). We are not on pro version.

**We are using `ReactOnRails.reactOnRailsPageLoaded();` when loading components asynchronously.**

The error shows that React is attempting to hydrate existing DOM content with kebab-case CSS properties, when it should be doing a fresh client-side render with camelCase properties.

## Environment

1. Ruby version: 3.4.7
2. Rails version: 8.1
3. Shakapacker/Webpacker version: 9.4.0
4. React on Rails version: 16.2.0.beta.19

## Expected behavior
No hydration, and no mismatch errors.

## Actual behavior
React is attempting to hydrate existing DOM content. The hydration comparison shows:
- Server/existing DOM: kebab-case CSS (align-items, flex-direction, margin-right)
- Client React: camelCase CSS (alignItems, flexDirection, marginRight)

## Notes

Possibly these changes are related:
- `reactOnRailsPageLoaded` used to call `hydrateAllStores()`, now it does not anymore? (https://github.com/shakacode/react_on_rails/blame/c3ad7a9b37ff4e98f968fbf572b6cd79424ed666/packages/react-on-rails/src/clientStartup.ts#L11-L14)
  - The current version just calls ´renderAllComponents` (https://github.com/shakacode/react_on_rails/blame/master/packages/react-on-rails/src/clientStartup.ts#L10)

- https://github.com/shakacode/react_on_rails/blob/16.1.2/node_package/src/pro/ClientSideRenderer.ts#L138-L139 the `supportsHydrate` is no longer part of the check?
  - As it has been changed to `const shouldHydrate = !!domNode.innerHTML;`
(https://github.com/shakacode/react_on_rails/blob/master/packages/react-on-rails/src/ClientRenderer.ts#L86-L87)

## Workaround

If you use `reactOnRailsComponentLoaded(domId)` per component, instead of `reactOnRailsPageLoaded()`, it works without any issues.

justin808:
## Summary

Fixes #2210 - When `reactOnRailsPageLoaded()` is called multiple times (e.g., for asynchronously loaded content), previously rendered client-side components were being re-hydrated, causing hydration mismatch errors in React 19.

**Root cause**: The `shouldHydrate` check (`!!domNode.innerHTML`) was true for previously client-rendered components because they already had HTML content. This caused React to try to hydrate content that wasn't server-rendered.

**Fix**: Skip rendering for components that are already tracked in the `renderedRoots` Map. This prevents re-hydration of already-rendered components while still allowing new components to be rendered.

## Changes

- **packages/react-on-rails/src/ClientRenderer.ts**: Add check to skip already-rendered components in `renderElement()`
- **packages/react-on-rails/tests/ClientRenderer.test.ts**: Add unit test for the fix
- **react_on_rails/spec/dummy/**: Add test page and E2E Playwright test for the scenario

## Test plan

- [x] Unit test passes: Component should not be re-rendered when `renderComponent()` is called twice
- [ ] Playwright E2E test: `async_page_loaded_test` page should show no hydration errors when adding components dynamically
- [ ] Manual test: Go to `/async_page_loaded_test`, click "Add Component" multiple times, verify no console errors

## Workaround mentioned in issue

The user noted that using `reactOnRailsComponentLoaded(domId)` per component works without issues. This fix makes `reactOnRailsPageLoaded()` work correctly as well.

## Release Notes

* **Bug Fixes**
  * Prevents duplicate renders when `reactOnRailsPageLoaded()` is called multiple times, eliminating hydration errors.
  * Properly handles DOM node replacement scenarios during component re-rendering.

* **Tests**
  * Added comprehensive test suite for multiple `reactOnRailsPageLoaded()` invocation scenarios.
  * Added end-to-end tests validating hydration behavior with dynamic component additions and node replacements.

### DIFF
--- a/packages/react-on-rails/src/ClientRenderer.ts
+++ b/packages/react-on-rails/src/ClientRenderer.ts
@@ -83,7 +83,9 @@ function renderElement(el: Element, railsContext: RailsContext): void {
       // (e.g., for asynchronously loaded content)
       const existing = renderedRoots.get(domNodeId);
       if (existing) {
-        // Only skip if it's the exact same DOM node and it's still connected to the document
+        // Only skip if it's the exact same DOM node and it's still connected to the document.
+        // If the node was replaced (e.g., via innerHTML or Turbo), we need to unmount the old
+        // root and re-render to the new node to prevent memory leaks and ensure rendering works.
         const sameNode = existing.domNode === domNode && existing.domNode.isConnected;
         if (sameNode) {
           if (trace) {


justin808:
@AbanoubGhadban: 
* Verified failure and the fix manually

justin808:
@claude review this PR

justin808:
@claude review this PR.

PR #2213, opened 2025-12-12
Fix Procfile.dev to use pnpm instead of yarn

justin808:
## Summary
- Fix rescript build command in Procfile.dev to use `pnpm` instead of `yarn`

The project uses pnpm as its package manager, but the Procfile.dev was still using yarn for the rescript build command, causing corepack to fail with:
```
UsageError: This project is configured to use pnpm because package.json has a "packageManager" field
```

## Test plan
- [x] Run `bin/dev` in spec/dummy and verify rescript process starts without corepack errors

### DIFF
--- a/bin/ci-local
+++ b/bin/ci-local
@@ -153,7 +153,7 @@ ensure_pro_dependencies() {
 
   if [ ! -d "react_on_rails_pro/node_modules" ] || [ ! -d "react_on_rails_pro/vendor/bundle" ]; then
     echo -e "${YELLOW}Installing React on Rails Pro dependencies...${NC}"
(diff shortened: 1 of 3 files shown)


justin808:
Trying out this prompt with Opus 4.5:

> Can you verify that all the changes you have made work? Could you add some sort of print message that would print when each one of your scripts or proc files is executed and then remove those print statements once you verify that you tested each one of those changes? Basically, add debug statements just so you can see the output and verify that you knew the right script to run and ensure that the script ran. Otherwise, you may think you are running the right program but you are not, and the code you change is never executed.

PR #2219, opened 2025-12-12
Migrate from markdown-link-check to lychee

### LINKED ISSUE #2217: Link checker CI passes despite broken links on master
## Summary

The markdown link checker CI workflow passes in ~5 seconds despite master having 21+ broken GitHub URLs that return HTTP 404.

### Broken URLs on master (all return 404)

```
https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/install_generator.rb
https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/config/shakapacker.yml
https://github.com/shakacode/react_on_rails/tree/master/spec/dummy
https://github.com/shakacode/react_on_rails/tree/master/lib/react_on_rails/controller.rb
... (21+ total)
```

These should be prefixed with `react_on_rails/` after the monorepo restructure in PR #2114.

### Local vs CI behavior

| Environment | Time | Result |
|-------------|------|--------|
| Local (npx markdown-link-check) | ~4 minutes | ❌ Finds 404 errors correctly |
| CI (GitHub Actions) | ~5 seconds | ✅ "All links are good!" |

### Timeline

- **Nov 4**: wretry.action wrapper added (PR #1899)
- **Nov 16**: CI failed and caught broken relative link `../CHANGELOG.md` (took 96-99 seconds)
- **Nov 24**: Monorepo restructure merged (PR #2114), breaking many GitHub URLs
- **Nov 24+**: All CI runs pass in ~5 seconds despite broken external URLs

### CI Run Comparison

**Failed run (Nov 16)** - [Run #19398864335](https://github.com/shakacode/react_on_rails/actions/runs/19398864335)
- find command took 96-99 seconds
- Found error: `ERROR: 1 dead links found in docs/api-reference/configuration-deprecated.md !`

**Recent "successful" run on master** - [Run #20147672041](https://github.com/shakacode/react_on_rails/actions/runs/20147672041)  
- find command took only 5 seconds
- Reports "All links are good!" despite 21+ broken URLs

### Local verification

```bash
# This returns 404
curl -sI "https://github.com/shakacode/react_on_rails/blob/master/lib/generators/react_on_rails/install_generator.rb" | head -1
# HTTP/2 404

# Local tool catches it
npx markdown-link-check docs/api-reference/redux-store-api.md --config .github/markdown-link-check-config.json
# ERROR: 4 dead links found!
```

## Root Cause Theory

Something changed in the GitHub Actions environment between Nov 16 and Nov 24 that causes the Docker container running markdown-link-check to complete almost instantly (~5 seconds for 102 files) without making actual HTTP requests.

The 5-second completion time is impossible if network requests are being made - each external link check takes 0.5-1 second minimum.

## Workflow Configuration

`.github/workflows/check-markdown-links.yml`:
- Uses `Wandalen/wretry.action@v3.5.0` wrapper
- Wraps `tcort/github-action-markdown-link-check@a800ad5f1c35bf61987946fd31c15726a1c9f2ba`
- Config file: `.github/markdown-link-check-config.json`

## Impact

The link checker CI is not functioning as a safety net - broken links can merge to master undetected.

## Possible Solutions

1. Investigate why Docker container isn't making HTTP requests
2. Switch to a different link checker (e.g., [Linkspector](https://github.com/UmbrellaDocs/action-linkspector) recommended in the deprecation notice)
3. Add verbose logging to diagnose the issue
4. Test without the wretry.action wrapper

## Related

- PR #2216 fixes the broken URLs introduced by the monorepo restructure
- PR #2114 was the monorepo restructure that broke the URLs
- PR #1899 added the wretry.action wrapper

🤖 Generated with [Claude Code](https://claude.ai/code)

ihabadham:
## Summary

Replaces the broken `markdown-link-check` setup with `lychee` for link checking.

## Problem

The previous setup using `Wandalen/wretry.action` wrapper with `tcort/markdown-link-check` was **silently failing** due to a Node.js 21 ESM compatibility bug (see #2217). The tool crashed on every file, but the crash message didn't match the error detection pattern, so CI falsely reported "All links are good!" - while actually checking zero links.

## Solution

Migrate to [lychee](https://github.com/lycheeverse/lychee), a Rust-based link checker that:

- ✅ **Built-in retry for ALL transient errors** (502, 503, 504) - what PR #1899 wanted but couldn't achieve
- ✅ **Proper exit codes** (2 = link failures, not relying on string matching)
- ✅ **Faster execution** (~12 seconds vs 4+ minutes for 550 links)
- ✅ **Result caching** for faster repeated runs
- ✅ **No Docker/Node.js compatibility issues**

## Changes

- Replace `.github/markdown-link-check-config.json` with `.lychee.toml`
- Update workflow to use `lycheeverse/lychee-action@v2`
- Add `.lycheecache` to `.gitignore`

## Test Plan

- [x] CI link checker workflow runs successfully
- [x] lychee correctly identifies broken links (will show ~26 errors from known broken URLs that need separate fixes)

## Note

This PR will **fail CI** because it correctly detects ~26 broken links that the previous setup missed. Those links need to be fixed in a separate PR (like #2216).

Fixes #2217

## Summary by CodeRabbit

* **Chores**
  * Migrated CI markdown-link checking to Lychee with caching and updated workflow steps.
  * Removed the previous markdown-link-check configuration and added a new Lychee configuration.
  * Updated ignore rules to stop tracking Lychee cache artifacts.

* **Documentation**
  * Rewrote and refreshed release notes and migration docs; updated tutorial links and formatting.
  * Clarified contributing guidance, changelog entry, and project listing content.

### DIFF
--- a/react_on_rails_pro/CONTRIBUTING.md
+++ b/react_on_rails_pro/CONTRIBUTING.md
@@ -178,7 +178,9 @@ To do this, follow the instructions in the
    It's run automatically when you run `yarn install`.
 3. Refresh the browser if the server is already running or start the server using `foreman start -f Procfile.dev` from `spec/dummy` and navigate to `http://localhost:3000/`. You will now see the `Hello!` message printed in the browser's console.
 
-> [!NOTE] > `yalc` makes the NPM package available globally on the machine.
+<!-- prettier-ignore -->
+> [!NOTE]
+> `yalc` makes the NPM package available globally on the machine.
 > So, if you have the repo checked out more than once to compare behavior between branches,
 > make sure to run `yarn install` every time you switch to a new copy.
 


.lychee.toml
```diff
@@ -0,0 +1,44 @@
+# Lychee link checker configuration
+# https://lychee.cli.rs
+
+# Accept successful responses and permanent redirects
+# 200 = OK, 206 = Partial Content, 308 = Permanent Redirect
+accept = ["200", "206", "308"]
```
alexeyr-ci2: Why is 206 OK?

justin808:
@ihabadham can you have CC post a reply here if you want to do the suggestions in https://github.com/shakacode/react_on_rails/pull/2219#issuecomment-3648568432 in this PR or make a follow up or not do it.

.lychee.toml
ihabadham: It was in the old config. Not sure why I added it then but we don't actually need it. Removed

ihabadham:
## Response to Claude Code Review Suggestions

We investigated each suggestion and addressed them as follows:

### ✅ Addressed

| Suggestion | Action Taken |
|------------|--------------|
| **1. Limited Scope** | Expanded to check `docs/`, `*.md`, `react_on_rails_pro/docs/`, `react_on_rails_pro/*.md`. Also added path exclusions for `node_modules` and `spec` directories. |
| **3. Retry Configuration** | Changed to 4 retries with 15s wait - balanced between speed and resilience. |
| **4. medium.com Context** | Added inline comment explaining why it's excluded (returns 403 for bots). |

### ❌ Not Needed (Investigated)

| Suggestion | Why Not Needed |
|------------|----------------|
| **2. Missing 301/302 in accept** | **Incorrect suggestion.** Lychee follows redirects automatically via `max_redirects` (default 10). The `accept` list is for the *final* status after following redirects. Our CI output showed "🔀 54 Redirects" as successful. |
| **2. Add 429 to accept** | Decided against. Our retries handle transient 429s. Persistent 429s (anti-bot) should fail so we know to add explicit exclusions. |
| **5. GitHub Blob URLs (base)** | Only links starting with `/` are in `docs/planning/` which is already excluded via `exclude_path`. Verified with `grep -rE '\]\(/' docs/`. |
| **6. HTTP Headers** | Verified GitHub returns HTML without Accept header. Tested with `curl -sI "https://github.com/shakacode/react_on_rails"` → `content-type: text/html`. |

### Additional Changes Made

- **Removed 206 from accept**: No URLs in our docs return 206. Lychee doesn't send Range headers.
- **Added docs.google.com exclusion**: Private docs require auth, will always 401.
- **Fixed 4 broken links in react_on_rails_pro/**: Updated paths after monorepo restructure (helper.ts, RSC tutorial links).

All changes are in commits:
- `06ce277` - Remove 206 from accept
- `1b4d734` - Expand scope, document medium.com
- `e8acb5c` - Full monorepo scope, fix broken links, add exclusions

alexeyr-ci2:
> 4. medium.com Context | Added inline comment explaining why it's excluded (returns 403 for bots).

If the link is actually wrong, does it still return 403? Or 404 so we can distinguish? I don't know if we can allow 403 only for medium.com even in that case.

ihabadham:
> If the link is actually wrong, does it still return 403?

@alexeyr-ci2 yes. medium uses cloudflare, so the checker won't really reach medium to get the 404

> I don't know if we can allow 403 only for medium.com even in that case.

I don't think we can, lychee doesn't support per-domain accept configurations

react_on_rails_pro/docs/release-notes/4.0.md
```diff
@@ -1,94 +1,103 @@
-# 4.0 Release Notes
```
alexeyr-ci2, line 1: Doesn't matter much but your last commit should probably have been a separate PR.

react_on_rails_pro/CONTRIBUTING.md
```diff
@@ -167,31 +174,32 @@ To do this, follow the instructions in the
...
-   It's run automatically when you run `yarn install`. 
+2. The `preinstall` script of `spec/dummy` builds the NPM package and sets up `yalc` to use it for the renderer.
+   It's run automatically when you run `yarn install`.
 3. Refresh the browser if the server is already running or start the server using `foreman start -f Procfile.dev` from `spec/dummy` and navigate to `http://localhost:3000/`. You will now see the `Hello!` message printed in the browser's console.
 
-> [!NOTE]
-> `yalc` makes the NPM package available globally on the machine.
+> [!NOTE] > `yalc` makes the NPM package available globally on the machine.
 > So, if you have the repo checked out more than once to compare behavior between branches,
 > make sure to run `yarn install` every time you switch to a new copy.
```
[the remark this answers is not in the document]
alexeyr-ci2: This is weird, yes, I saw Prettier do the same locally as well.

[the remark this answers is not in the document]
alexeyr-ci2: @claude Please add `<!-- prettier-ignore -->` comment here as described.

alexeyr-ci2:
@coderabbitai Review and approve if the issue is fixed.

react_on_rails_pro/docs/release-notes/4.0.md
ihabadham: Yeah but coderabbit flagged it here since the PR changed the file, so I thought why not

PR #2271, opened 2025-12-31
Investigation: Partial bundle error analysis

AbanoubGhadban:
## Summary
- Investigation scripts for reported "Invalid or unexpected token" / "missing ) after argument list" errors during SSR

### Scenarios Tested

| Scenario | Result | Can Cause Error? |
|----------|--------|------------------|
| Concurrent uploads of same bundle content | Files end up valid | ❌ No |
| File descriptor race with O_TRUNC | Null bytes appear | ⚠️ Only with different content |
| Non-atomic move (cross-device) | Partial file readable | ✅ Yes |
| Direct deploy file writes | Partial file during write | ✅ Yes (if not using HTTP upload) |
| Interrupted HTTP upload | Partial file saved | ✅ Yes (but persistent, not transient) |

### Root Cause Analysis

The **only scenario** that matches the reported "transient" error pattern (first few requests fail, then works) is:

**Non-atomic move operation** when bundle path is on a different filesystem than the uploads directory.

When `fs.rename()` fails with `EXDEV` (cross-device link), `fs-extra.move()` falls back to `copy()` + `unlink()`. During the copy operation, a concurrent request can read a partial file.

### Evidence

1. **Null byte corruption requires different content** - When two requests upload the same bundle, the file ends up valid even with race conditions.

2. **Same-filesystem moves are atomic** - `fs.rename()` is atomic on the same filesystem, so no partial reads are possible.

3. **Cross-device moves are NOT atomic** - The copy fallback creates a window where partial files can be read.

### Conclusion

This appears to be a **project-specific configuration issue** rather than a bug in react_on_rails:

- The bundle path (`bundlePath: path.resolve(__dirname, 'tmp/bundles')`) must be on a different filesystem than the uploads directory
- This is an unusual configuration that causes cross-device moves
- Only one project has reported this issue

## Test Scripts

22 test scripts were created to investigate various race condition scenarios. These can be run to reproduce and understand the behavior:

```bash
cd packages/react-on-rails-pro-node-renderer
node scripts/test-fd-truncation.mjs      # Proves null byte corruption mechanism
node scripts/test-concurrent-race.mjs    # Tests non-atomic copy race
node scripts/test-multipart-race.mjs     # Tests multipart upload race
```

## Recommendation

For the affected project:
1. Ensure `serverBundleCachePath` is on the same filesystem as the application
2. Or use a RAM disk / tmpfs for both uploads and bundles directories
3. Check if there are any NFS or network mounts involved

## Tests
- Added comprehensive test suite for file operation atomicity, concurrent requests, race conditions, truncation scenarios, and upload interruptions to enhance system reliability.

### DIFF
--- a/packages/react-on-rails-pro-node-renderer/scripts/test-atomicity-parallel.mjs
+++ b/packages/react-on-rails-pro-node-renderer/scripts/test-atomicity-parallel.mjs
@@ -1,0 +1,239 @@
+#!/usr/bin/env node
+/**
+ * Test atomicity using child processes for TRUE parallelism.
+ *
+ * Usage:
+ *   node scripts/test-atomicity-parallel.mjs
+ */
+
+import { fork, spawn } from 'node:child_process';
+import fs from 'node:fs';
+import fsp from 'node:fs/promises';
+import path from 'node:path';
+import os from 'node:os';
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const TEST_FILE_SIZE_MB = 200;
+const TEST_DIR = path.join(os.tmpdir(), `atomicity-parallel-${Date.now()}`);
+
+async function generateTestFile(filePath, sizeMB) {
+  console.log(`Generating ${sizeMB}MB test file...`);
+  const start = Date.now();
+
+  const fd = fs.openSync(filePath, 'w');
+  const chunk = Buffer.alloc(1024 * 1024, 'x'); // 1MB of 'x'
+
+  for (let i = 0; i < sizeMB; i++) {
+    fs.writeSync(fd, chunk);
+  }
+  fs.writeSync(fd, '\n// COMPLETE_MARKER_12345_END //\n');
+  fs.closeSync(fd);
+
+  console.log(`Generated in ${Date.now() - start}ms`);
+}
+
+// Reader script that runs in a child process
+const READER_SCRIPT = `
+const fs = require('fs');
+const destPath = process.argv[2];
+const expectedSize = parseInt(process.argv[3], 10);
+
+let readAttempts = 0;
+let partialReads = 0;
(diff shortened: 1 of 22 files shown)


AbanoubGhadban:
## Closing PR - Investigation Complete

This PR contains investigation scripts only (no code changes to merge). The findings are documented below.

---

## Investigation Conclusion

After extensive testing with 22 different race condition scenarios, we've determined that:

### The Only Scenario That Can Cause Transient Partial Bundle Errors

**Non-atomic `move` operation** - This occurs when the bundle path (`serverBundleCachePath`) is on a **different filesystem** than the uploads directory.

When `fs.rename()` fails with `EXDEV` (cross-device link error), `fs-extra.move()` falls back to a non-atomic `copy()` + `unlink()` operation. During the copy, concurrent requests can read a partial file, causing:

- `SyntaxError: Invalid or unexpected token`
- `SyntaxError: missing ) after argument list`

### Why Other Scenarios Don't Apply

| Scenario | Why It Doesn't Cause This Bug |
|----------|------------------------------|
| Concurrent uploads of same content | Both write identical bytes → file ends up valid |
| Direct deploy file writes | User confirmed they use HTTP upload |
| Interrupted HTTP upload | Would cause **persistent** errors, not transient |
| File descriptor O_TRUNC race | Only corrupts with **different** content |

### Root Cause: Project Configuration

Since:
1. This has only been reported by **one project**
2. The local bundle path (`tmp/bundles`) should normally be on the same filesystem
3. Same-filesystem moves are atomic (no partial file race possible)

**The issue appears to be project-specific configuration** where the bundle path is on a different filesystem (NFS mount, Docker volume, etc.) causing cross-device moves.

### Recommendation for Affected Project

1. Verify `serverBundleCachePath` is on the same filesystem as the uploads directory
2. Check for NFS mounts or Docker volume configurations
3. Consider using ephemeral storage (tmpfs/emptyDir) for both directories

---

The investigation scripts are preserved in this branch for future reference.

PR #2273, opened 2026-01-01
Fix bin/dev failing with 'Unknown argument' when using --route flag

ihabadham:
## Summary

Fixes `bin/dev` failing with "Unknown argument: hello_world" immediately after running `rails generate react_on_rails:install`.

## Problem

The generator creates a `bin/dev` script that passes `["--route", "hello_world"]` to `ServerManager.run_from_command_line`. The early command extraction (added in PR #2092) was incorrectly picking up "hello_world" as the command instead of recognizing it as the value for `--route`.

```bash
$ ./bin/dev
Unknown argument: hello_world
Run 'dev help' for usage information
```

## Root Cause

PR #2092 moved command extraction before OptionParser to enable skipping precompile hooks for `kill`/`help` commands:

```ruby
# This finds "hello_world" as the "command" because it doesn't start with "-"
command = args.find { |arg| !arg.start_with?("--") && !arg.start_with?("-") }
```

## Solution

Added `extract_command_from_args` method that properly skips values following flags like `--route` and `--rails-env`:

```ruby
def extract_command_from_args(args)
  skip_next = false
  args.each do |arg|
    if skip_next
      skip_next = false
      next
    end
    if FLAGS_WITH_VALUES.include?(arg)
      skip_next = true
      next
    end
    next if arg.start_with?("-")
    return arg
  end
  nil
end
```

## Test Plan

- [x] All 35 existing `server_manager_spec.rb` tests pass
- [x] Manual testing: Created fresh Rails app, ran generator, `./bin/dev` works
- [x] Verified `bin/dev help`, `bin/dev kill`, `bin/dev static` still work correctly

## Summary by CodeRabbit

* **Refactor**
  * Improved development-server CLI parsing so flags that take values (e.g., route or rails environment) are ignored when determining the command, preventing flag values from being misinterpreted as commands.

* **Tests**
  * Added comprehensive tests for flag/value forms, flag ordering, verbosity flags, default behavior, and unknown-command handling to validate parsing.

### DIFF
--- a/react_on_rails/spec/react_on_rails/dev/server_manager_spec.rb
+++ b/react_on_rails/spec/react_on_rails/dev/server_manager_spec.rb
@@ -440,4 +440,160 @@ RSpec.describe ReactOnRails::Dev::ServerManager do
       end
     end
   end
+
+  # These tests verify argument parsing works correctly, following Rails' CLI testing pattern
+  # See: https://github.com/rails/rails/blob/main/railties/test/commands/server_test.rb
+  describe ".run_from_command_line argument parsing" do
+    before do
+      mock_system_calls
+      allow(ReactOnRails::PackerUtils).to receive(:shakapacker_precompile_hook_value).and_return(nil)
+      allow(ReactOnRails::Dev::ServiceChecker).to receive(:check_services).and_return(true)
+    end
+
+    context "with --route flag" do
+      # This test would have caught the bug fixed in PR #2273
+      # The generator creates bin/dev with: argv_with_defaults.push("--route", DEFAULT_ROUTE)
+      # which passes ["--route", "hello_world"] to run_from_command_line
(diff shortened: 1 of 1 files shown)


justin808:
@claude is this handled by https://github.com/shakacode/react_on_rails/pull/2309

ihabadham:
@justin808 @claude PR #2309 fixes the default case (the generated `bin/dev` template now uses `--route=hello_world`).

However, the parser bug still exists. Per https://docs.ruby-lang.org/en/3.0/OptionParser.html, both `--route=value` and `--route value` are valid formats, but only the first works currently:

```bash
bin/dev --route=hello_world  # works
bin/dev --route hello_world  # fails with "Unknown argument: hello_world"
```

This PR fixes the early command extraction to handle both formats correctly, and adds unit tests for argument parsing.

PR #2354, opened 2026-02-05
Only warn on duplicate registration when component/store differs

### LINKED ISSUE #2117: Ability to suppress `console.warn`s

justin808:
## Summary
- Instead of always warning on duplicate component/store registration, only warn when a **different** component or store generator is registered under an already-used name
- Silently accepts re-registration of the same component (common with HMR)
- No configuration option needed — the behavior is always correct

## Motivation
Closes #2117

The previous approach added a `warn_on_duplicate_registration` config option with Ruby config, Rails context propagation, and JS option handling. This simpler approach makes the detection smarter instead: compare references and only warn when it matters (a different component under the same name, which is likely a bug).

## Changes
- **ComponentRegistry.ts**: Compare component reference before warning. Only warn when `existing.component !== component`
- **StoreRegistry.ts**: Compare store generator reference before warning. Only warn when `existing !== store`
- **Tests**: Updated to verify both cases (same component = no warning, different component = warning)

## Test plan
- [x] All 112 react-on-rails JS tests pass
- [x] ESLint passes
- [x] Prettier passes
- [x] Pre-commit hooks pass
- [ ] CI

duplicate-registration-tests-pro

## Summary by CodeRabbit

* **Bug Fixes**
  * Silent re-registration for identical components/stores; registering a different component/store under an existing name now logs a clear console error to surface conflicts.

* **New Features**
  * Added a non-throwing lookup to safely retrieve callback registry entries if present.

* **Tests**
  * Tests updated to cover silent re-registration and to assert error logging on name conflicts.

### DIFF
--- a/packages/react-on-rails-pro/src/ComponentRegistry.ts
+++ b/packages/react-on-rails-pro/src/ComponentRegistry.ts
@@ -29,6 +29,8 @@ export function register(components: Record<string, ReactComponentOrRenderFuncti
       throw new Error(`Called register with null component named ${name}`);
     }
 
+    // Reference comparison lets HMR re-register the same component silently
+    // while still catching bugs where different components share a name.
     const existing = componentRegistry.getIfExists(name);
     if (existing && existing.component !== component) {
       console.error(
--- a/packages/react-on-rails-pro/src/StoreRegistry.ts
+++ b/packages/react-on-rails-pro/src/StoreRegistry.ts
@@ -33,6 +33,8 @@ export function register(storeGenerators: Record<string, StoreGenerator>): void 
       );
     }
 
+    // Reference comparison lets HMR re-register the same store silently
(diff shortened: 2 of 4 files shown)


justin808:
@claude review this PR

justin808:
@claude review this PR

PR #2416, opened 2026-02-15
Fix StreamResponse status fallback for non-streaming errors

### LINKED ISSUE #2408: Fix NoMethodError on HTTPX::StreamResponse.status for non-streaming error responses
## Problem

When the Node renderer returns a non-streaming error response (e.g., a plain HTTP 400 from a synchronous throw before any stream is created), HTTPX's stream plugin wraps it in a `StreamResponse`. The `StreamResponse` uses `method_missing` to delegate to the underlying response, but `.status` delegation fails with:

```
NoMethodError: undefined method 'status' for an instance of HTTPX::StreamResponse
```

This occurs in `StreamRequest#process_response_chunks` where `stream_response.status >= 400` is called inline. The `NoMethodError` is raised inside the `ensure` block of `loop_response_lines`, which replaces the original `HTTPX::HTTPError` exception — a classic Ruby ensure gotcha.

## Root Cause

In `stream_request.rb`, `process_response_chunks` calls `stream_response.status >= 400` directly. When the response is a non-streaming error wrapped in `StreamResponse`, `method_missing` cannot delegate `.status`, raising `NoMethodError`. This error propagates through the `ensure` block in `loop_response_lines`, replacing the original `HTTPX::HTTPError` that would have been properly handled by `handle_http_error`.

## Proposed Fix

Extract an `error_status?` helper method that rescues `NoMethodError`:

```ruby
def error_status?(response)
  response.status >= 400
rescue NoMethodError
  # HTTPX::StreamResponse may fail to delegate .status for non-streaming
  # error responses. Treat as error when status is unavailable, and let
  # the original HTTPX::HTTPError propagate to handle_http_error.
  true
end
```

## Reproduction

Add a synchronous `throw new Error(...)` at the top of `streamRenderReactComponent` (before any stream is created) in `packages/react-on-rails-pro/src/streamServerRenderedReactComponent.ts`, then load a page that uses streaming SSR.

## Related

This was discovered while working on #2402 / PR #2407 (streaming hang fix) but deferred to avoid scope creep.

justin808:
## Summary
- Fix `StreamRequest#process_response_chunks` to treat responses as errors when `status` delegation raises `NoMethodError`
- Avoid masking the original HTTPX error path when a `StreamResponse` cannot answer `#status`
- Add a regression spec that exercises the missing-status delegation path

Closes #2408

## Test plan
- Proposed fix (UNTESTED in this environment)
- Attempted: `bundle exec rspec react_on_rails_pro/spec/react_on_rails_pro/stream_request_spec.rb`
- Blocked by local Ruby version (`2.6.10`) vs project requirement (`>= 3.0.0`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for streaming responses to properly detect and handle error conditions in edge cases.

* **Tests**
  * Added test coverage for streaming response error scenarios.


justin808:
@AbanoubGhadban @ihabadham @alexeyr-ci2 This was done independently by codex.

PR #2429, opened 2026-02-16
Fix RSC install generator layout wiring and post-install helper hint

justin808:
## Summary
- always copy `app/views/layouts/hello_world.html.erb` during install so `--rsc` apps have a pack-driven layout available
- set generated `HelloServerController` to `layout "hello_world"` so fresh Rails apps do not depend on `javascript_pack_tag "application"`
- update post-install messaging to show `stream_react_component(...)` when `--rsc` is used
- extend generator specs to assert the RSC layout/controller behavior and install message helper output

## Why
Generating a fresh Rails app with `rails g react_on_rails:install --rsc --typescript` produced a runtime `Shakapacker::Manifest::MissingEntryError` because `application` layout references `javascript_pack_tag "application"` while no `application` pack is generated in this flow.

## Verification
- `mise exec ruby@3.3 -- bundle exec rspec react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb react_on_rails/spec/react_on_rails/generators/generator_messages_spec.rb`
- `mise exec ruby@3.3 -- bundle exec rubocop react_on_rails/lib/generators/react_on_rails/base_generator.rb react_on_rails/lib/generators/react_on_rails/templates/rsc/base/app/controllers/hello_server_controller.rb react_on_rails/lib/generators/react_on_rails/generator_messages.rb react_on_rails/lib/generators/react_on_rails/install_generator.rb react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb react_on_rails/spec/react_on_rails/generators/generator_messages_spec.rb`

## Summary by CodeRabbit

* **New Features**
  * Post-install messages now show the appropriate render example for React Server Components (RSC) vs non-RSC installs.

* **Improvements**
  * Installer guidance dynamically reflects RSC vs non-RSC setups for clearer next steps.

* **Bug Fixes**
  * Installer now ensures the app layout and controller wiring for server-rendered setups are created to prevent missing-layout errors on fresh installs.

* **Tests**
  * Added tests verifying RSC and non-RSC post-install messages and layout/controller behavior.

---

> [!NOTE]
> **Low Risk**
> Generator/template and messaging changes with limited runtime impact beyond newly generated apps; risk is mainly incorrect layout selection affecting fresh RSC installs.
> 
> **Overview**
> Fixes a fresh `rails g react_on_rails:install --rsc` runtime failure where `HelloServerController` could fall back to Rails’ default `application.html.erb` (referencing a non-existent `application` pack), causing `Shakapacker::Manifest::MissingEntryError`.
> 
> The generator now **always copies** `app/views/layouts/hello_world.html.erb`, sets the RSC `HelloServerController` template to `layout "hello_world"`, and updates the post-install message to show `stream_react_component(...)` for RSC (with new specs covering the updated messaging and RSC file generation).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 95306eb5df9ea9f4de9c85697bd2ab05b826e662. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
+++ b/react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
@@ -840,6 +840,14 @@ describe InstallGenerator, type: :generator do
         expect(deps).to include("redux")
       end
     end
+
+    it "creates HelloServer controller with hello_world layout" do
+      assert_file "app/controllers/hello_server_controller.rb" do |content|
+        expect(content).to include("class HelloServerController")
+        expect(content).to include('layout "hello_world"')
+        expect(content).to include("ReactOnRailsPro::Stream")
+      end
+    end
   end
 
   context "with --rsc --typescript" do
@@ -884,6 +892,14 @@ describe InstallGenerator, type: :generator do
         expect(dev_deps).to include("@types/react")
       end
     end
+
+    it "creates HelloServer controller with hello_world layout" do
+      assert_file "app/controllers/hello_server_controller.rb" do |content|
+        expect(content).to include("class HelloServerController")
+        expect(content).to include('layout "hello_world"')
+        expect(content).to include("ReactOnRailsPro::Stream")
+      end
+    end
   end
 
   context "with --rsc --rspack" do
@@ -915,6 +931,14 @@ describe InstallGenerator, type: :generator do
         deps = package_json["dependencies"] || {}
         expect(deps).to include("react-on-rails-rsc")
         expect(deps).to include("@rspack/core")
+      end
+    end
+
+    it "creates HelloServer controller with hello_world layout" do
+      assert_file "app/controllers/hello_server_controller.rb" do |content|
+        expect(content).to include("class HelloServerController")
+        expect(content).to include('layout "hello_world"')
+        expect(content).to include("ReactOnRailsPro::Stream")
       end
     end
   end


react_on_rails/lib/generators/react_on_rails/generator_messages.rb
```diff
@@ -83,6 +89,15 @@ def helpful_message_after_installation(component_name: "HelloWorld", route: "hel
...
     private
 
+    def build_render_example(component_name:, route:, rsc:)
+      helper_name = rsc ? "stream_react_component" : "react_component"
+      if rsc
+        "<%= #{helper_name}(\"#{component_name}\", props: @#{route}_props) %>"
+      else
+        "<%= #{helper_name}(\"#{component_name}\", props: @#{route}_props, prerender: true) %>"
+      end
+    end
```
ihabadham, line 98: `helper_name` is derived from `rsc` on this line, then `if rsc` is evaluated again in the very next line. The variable is used exactly once per branch and adds no clarity. Each branch can embed the name directly:

```suggestion
    def build_render_example(component_name:, route:, rsc:)
      if rsc
        "<%= stream_react_component(\"#{component_name}\", props: @#{route}_props) %>"
      else
        "<%= react_component(\"#{component_name}\", props: @#{route}_props, prerender: true) %>"
      end
    end
```

react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -695,7 +695,8 @@
       %w[config/initializers/react_on_rails.rb
          Procfile.dev
          Procfile.dev-static-assets
-         Procfile.dev-prod-assets].each { |file| assert_file(file) }
+         Procfile.dev-prod-assets
+         app/views/layouts/hello_world.html.erb].each { |file| assert_file(file) }
```
ihabadham, line 699: The `--rsc` context now asserts `hello_world.html.erb`, but the [`--rsc --redux` context](https://github.com/shakacode/react_on_rails/blob/95306eb5df9ea9f4de9c85697bd2ab05b826e662/react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb#L806-L811) and [`--rsc --typescript` context](https://github.com/shakacode/react_on_rails/blob/95306eb5df9ea9f4de9c85697bd2ab05b826e662/react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb#L847-L852) have the same "copies common files" block without this assertion. The fix in `base_generator.rb` is unconditional so the file is always copied for all install types. Both contexts should also verify it:

```ruby
      %w[config/initializers/react_on_rails.rb
         Procfile.dev
         Procfile.dev-static-assets
         Procfile.dev-prod-assets
         app/views/layouts/hello_world.html.erb].each { |file| assert_file(file) }
```

react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -773,6 +774,7 @@
     it "creates HelloServer controller and view" do
       assert_file "app/controllers/hello_server_controller.rb" do |content|
         expect(content).to include("class HelloServerController")
+        expect(content).to include('layout "hello_world"')
```
ihabadham, line 777: This assertion was added for `--rsc`, but `--rsc --redux` also creates `HelloServerController` and has no assertion on its content. If the `layout "hello_world"` declaration were missing or regressed in the Redux combination, no test would catch it. Suggest adding to the `--rsc --redux` context:

```ruby
    it "creates HelloServer controller with hello_world layout" do
      assert_file "app/controllers/hello_server_controller.rb" do |content|
        expect(content).to include("class HelloServerController")
        expect(content).to include('layout "hello_world"')
        expect(content).to include("ReactOnRailsPro::Stream")
      end
    end
```

ihabadham (changes requested):
Fix itself looks right. Before this merges though:

**1. `hello_world.html.erb` naming debt**

The fix is correct and low-risk. The layout file needs to exist and the controller needs to declare it explicitly. But in a `--rsc`-only install there is no HelloWorld controller, component, or route anywhere. A developer opening their generated app will find `app/views/layouts/hello_world.html.erb` with no HelloWorld anything, and `HelloServerController` declaring `layout "hello_world"`. The name does not reflect what the file actually does (inject pack tags via the empty `javascript_pack_tag` / `stylesheet_pack_tag` pattern). Worth a follow-up to introduce a neutral layout name so the layout is decoupled from the HelloWorld demo.

**2. `prerender: true` in the generated view template is a no-op**

`react_on_rails/lib/generators/react_on_rails/templates/rsc/base/app/views/hello_server/index.html.erb#L19` (not changed in this PR) generates:

```erb
<%= stream_react_component('HelloServer', props: @hello_server_props, prerender: true) %>
```

But `stream_react_component` always forces `options[:prerender] = true` internally (see `react_on_rails_pro/app/helpers/react_on_rails_pro_helper.rb#L127-L129`) and its own comment says it "doesn't have the prerender option." This PR correctly drops `prerender: true` from the post-install hint but leaves it in the generated view file, so a developer reading the template is still taught a no-op option. The comment above it on line 13 even says "prerender: true enables server-side rendering with streaming," which is misleading since the flag does nothing. Worth cleaning up the template in a follow-up.

react_on_rails/lib/generators/react_on_rails/generator_messages.rb
justin808: agree

react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
justin808: Fine, although I'm not sure we need the support for RSC and Redux, which is adding a bunch more stuff for us to test. I'm not sure it makes sense in the context of RSC.

ihabadham: I did consider blocking this combination, but kept it for the upgrade path: existing apps with Redux that want to add RSC on top. For new installs it doesn't make much sense, agreed.

ihabadham (approved):

react_on_rails/spec/react_on_rails/generators/generator_messages_spec.rb
```diff
@@ -24,4 +24,27 @@
     expect(described_class.messages)
       .to contain_exactly(described_class.format_info("Test info message"))
   end
+
+  it "shows stream_react_component in RSC install message" do
```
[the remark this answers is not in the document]
ihabadham: No action needed — `generator_spec_helper.rb` has a global `config.after { GeneratorMessages.clear }` that clears state after every example.

react_on_rails/lib/generators/react_on_rails/generator_messages.rb
```diff
@@ -38,11 +38,17 @@ def clear
...
       process_manager_section = build_process_manager_section
       testing_section = build_testing_section
       package_manager = detect_package_manager
       shakapacker_status = build_shakapacker_status_section
+      render_example = build_render_example(component_name: component_name, route: route, rsc: rsc)
+      render_label = if rsc
+                       "• Streaming server rendering in app/views/#{route}/index.html.erb:"
+                     else
+                       "• Server-side rendering - Enabled with prerender option in app/views/#{route}/index.html.erb:"
+                     end
```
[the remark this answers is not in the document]
ihabadham: Pre-existing, not introduced in this PR. Agreed it's worth cleaning up — will track as a follow-up.

react_on_rails/spec/react_on_rails/generators/install_generator_spec.rb
```diff
@@ -837,6 +840,14 @@
         expect(deps).to include("redux")
       end
     end
+
+    it "creates HelloServer controller with hello_world layout" do
```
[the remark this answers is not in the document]
ihabadham: The `--rsc` context test at line 774 is different (also asserts the view file), so there are 3 identical blocks, not 4. Agreed on the style point — will track as a follow-up.

PR #2444, opened 2026-02-18
Fix RSC stream tee backpressure deadlock for large payloads

### LINKED ISSUE #2443: Fix RSC payload stream backpressure deadlock in RSCRequestTracker
## Problem

`RSCRequestTracker.getRSCPayloadStream()` tees the RSC Flight stream into two PassThrough streams via `pipe()`:

- **stream1** — consumed by React's SSR runtime (immediate consumer)
- **stream2** — consumed by `injectRSCPayload` to embed RSC payloads as `<script>` tags in HTML (delayed consumer)

`injectRSCPayload` only starts reading stream2 **after the first HTML chunk arrives from SSR**. If the RSC Flight payload exceeds the default PassThrough `highWaterMark` (16KB), stream2's buffer fills before anyone reads it. When `pipe()` detects a full destination buffer, it pauses the **source stream**. Since both stream1 and stream2 share the same source, pausing it also starves stream1 — SSR stalls, no HTML is produced, `injectRSCPayload` never starts, and the render deadlocks.

### Why this wasn't caught earlier

Previous demos (e.g., search pages) had small RSC payloads (~5KB of data-driven UI) that fit within the 16KB buffer. The blog post demo renders a full markdown article with syntax highlighting (marked + highlight.js), producing ~60KB+ of Flight data — enough to overflow the buffer.

### Deadlock cycle

```
stream2 buffer fills (>16KB) → pipe() pauses source stream
→ stream1 starved → SSR stalls → no HTML produced
→ injectRSCPayload never starts consuming stream2
→ stream2 stays full → permanent deadlock
```

## Proposed Fix

Replace `pipe()` with manual event-based forwarding using `stream.on('data')` + `push()`. Since `push()` never blocks (it only returns a boolean advisory), neither destination can backpressure the source:

```ts
const stream1 = new PassThrough();
const stream2 = new PassThrough();

stream.on('data', (chunk: Buffer) => {
  stream1.push(Buffer.from(chunk));
  stream2.push(Buffer.from(chunk));
});
stream.on('end', () => {
  stream1.push(null);
  stream2.push(null);
});
stream.on('error', (err: Error) => {
  stream1.destroy(err);
  stream2.destroy(err);
});
```

This eliminates the deadlock regardless of payload size, without relying on oversized buffer configuration.

## Current Workaround

Setting `highWaterMark: 64 * 1024 * 1024` on stream2 prevents the buffer from filling for typical payloads, but wastes memory and doesn't fix the architectural issue.

AbanoubGhadban:
## Summary

- Replace `pipe()`-based stream teeing with manual `on('data')` + `push()` forwarding in `RSCRequestTracker.getRSCPayloadStream()`
- With `pipe()`, if either tee destination's buffer fills (~32KB), the source is paused, coupling both consumers' backpressure and deadlocking the render when payloads exceed the default highWaterMark
- With `push()`, each destination buffers independently — stream1 keeps receiving data regardless of stream2's consumption state
- Add unit, integration, and e2e tests covering payloads exceeding the default highWaterMark (sync + async chunk delivery, delayed SSR shell threshold, full `streamServerRenderedReactComponent` pipeline)

## Test plan

- [x] Unit tests: tee delivers data to both streams for small and large payloads
- [x] Integration tests: RSCRequestTracker + injectRSCPayload with simulated SSR (immediate and delayed shell)
- [x] E2E tests: full `streamServerRenderedReactComponent` pipeline with render functions consuming RSC streams
- [x] All 12 new tests pass (previously 7 deadlocked with 5s timeout)
- [x] All 21 existing tests in `streamServerRenderedReactComponent.test.jsx` and `injectRSCPayload.test.ts` continue to pass

Closes #2443

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for RSC streaming under varied backpressure, delayed-shell, and chunked scenarios
  * Added end-to-end tests validating RSC payload delivery, embedding, and large-payload resilience

* **Refactor**
  * Improved internal stream handling for more reliable streaming and backpressure resilience

* **Bug Fixes**
  * Prevented premature stream closure on error to improve robustness during streaming

### DIFF
--- a/packages/react-on-rails-pro/src/transformRSCStreamAndReplayConsoleLogs.ts
+++ b/packages/react-on-rails-pro/src/transformRSCStreamAndReplayConsoleLogs.ts
@@ -78,11 +78,10 @@ export default function transformRSCStreamAndReplayConsoleLogs(stream: ReadableS
           // eslint-disable-next-line no-await-in-loop
           ({ value, done } = await reader.read());
         }
+        controller.close();
       } catch (error) {
         console.error('Error transforming RSC stream:', error);
         controller.error(error);
-      } finally {
-        controller.close();
       }
     },
   });


packages/react-on-rails-pro/src/transformRSCStreamAndReplayConsoleLogs.ts
```diff
@@ -78,11 +78,10 @@ export default function transformRSCStreamAndReplayConsoleLogs(stream: ReadableS
           // eslint-disable-next-line no-await-in-loop
           ({ value, done } = await reader.read());
         }
+        controller.close();
```
AbanoubGhadban, line 81: This change is unrelated to the backpressure fix in this PR. It fixes a pre-existing bug where `controller.close()` in the `finally` block would throw a TypeError if `controller.error()` was already called in the `catch` block, since the Web Streams spec requires the stream to be in the `'readable'` state for `close()` to succeed.

justin808:
@claude review this PR.

PR #2473, opened 2026-02-22
Add PropsTransformer to reduce RSC props embedded in HTML

AbanoubGhadban:
## Summary
- Adds a `props_transformer` configuration option to ReactOnRailsPro that allows expanding compact "key props" into full rendering props on the server side
- Only the small key props appear in HTML output (RSC cache keys, hydration script), while full props are used for server-side rendering
- Reduces redundant prop data in HTML from ~112KB to a few hundred bytes for large-prop RSC components

## Test plan
- [x] RSCRequestTracker: keyPropsOverride set correctly, consumed-on-first-use, undefined when no override
- [x] injectRSCPayload: cache key uses keyProps when provided, falls back to props when undefined
- [x] Configuration: nil default, accepts valid transformer, rejects invalid transformer
- [x] All existing tests pass (backward compatible)

## Summary by CodeRabbit

* **New Features**
  * Added props_transformer configuration option for customizing server-side props handling
  * Introduced keyProps support to optimize RSC payload caching by selectively including props in cache keys
  * Extended RSC streaming with per-request keyProps override mechanism for improved caching efficiency

### DIFF
--- a/packages/react-on-rails-pro/src/RSCRequestTracker.ts
+++ b/packages/react-on-rails-pro/src/RSCRequestTracker.ts
@@ -52,8 +52,24 @@ class RSCRequestTracker {
 
   private railsContext: RailsContextWithServerComponentMetadata;
 
+  private keyPropsOverride?: unknown;
+
   constructor(railsContext: RailsContextWithServerComponentMetadata) {
     this.railsContext = railsContext;
+  }
+
+  /**
+   * Sets a key props override for RSC payload cache key generation.
+   *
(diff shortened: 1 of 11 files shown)


AbanoubGhadban:
Closing this PR — after further discussion, the recommended approach is to use **async props** (passing props as a block via `cached_stream_react_component` or similar patterns) to avoid embedding large props in the HTML at all. This removes the need for a PropsTransformer abstraction.

If you have large props that bloat the HTML output, consider restructuring your component to fetch data asynchronously on the server side rather than passing it all through the view layer.

PR #2477, opened 2026-02-23
Fix bin/setup failing on pnpm workspace member directories

justin808:
## Summary
- `bin/setup` was failing on workspace member directories (e.g., `spec/dummy`) because it ran `pnpm install --frozen-lockfile` in directories without a `pnpm-lock.yaml`
- Workspace members' dependencies are already installed by the root `pnpm install`, so they don't need their own install step
- Now checks for the lockfile before running `pnpm install`, and prints a message that dependencies are managed by the workspace root

## Test plan
- [x] Ran `bin/setup --skip-pro` locally — completes successfully
- [ ] CI passes

---

> [!NOTE]
> **Low Risk**
> Small, localized change to setup scripting that only relaxes a directory-level precondition for running `pnpm install`.
> 
> **Overview**
> Fixes `bin/setup` failures in pnpm workspace member directories by **only running `pnpm install --frozen-lockfile` when a local `pnpm-lock.yaml` exists**.
> 
> If a directory has a `package.json` but no lockfile (typical for workspace members like dummy apps), the script now skips the install step and reports that JS dependencies are managed by the workspace root.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3d38d3c0e1db2ab6d0eaf7bb409f8dbd489e406. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Chores**
  * Setup now installs JavaScript dependencies only when both a package manifest and a matching lockfile are present. If a manifest exists without a lockfile, installation is skipped and a clear notice indicates dependencies are managed at the workspace root, avoiding unnecessary local installs and improving setup clarity.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -165,7 +165,7 @@ install_dependencies() {
     fi
     print_success "JavaScript dependencies installed"
   elif [ -f "package.json" ]; then
-    print_warning "Skipping JavaScript dependencies in $name (no pnpm-lock.yaml); assuming workspace root manages them"
+    echo "  JavaScript dependencies in $name are managed by the workspace root (no local pnpm-lock.yaml)"
   fi
 
   cd "$ROOT_DIR"


justin808:
Addressed both open review points in `bin/setup`:
- Changed the skip message from success to warning.
- Clarified the skip condition as an assumption (`no pnpm-lock.yaml` => managed by workspace root).

This keeps behavior the same while avoiding false "success" signaling and making the heuristic explicit.

justin808:
Updated as requested: the workspace-member case now uses plain informational output instead of warning semantics.

PR #2529, opened 2026-03-04
Fix streaming deadlock with Rack compression middleware

### LINKED ISSUE #2519: RSC payload streaming deadlocks with Rack::Deflater and Rack::Brotli compression middleware
## Description

The RSC payload endpoint (`/rsc_payload/:component_name`) hangs indefinitely when `Rack::Deflater` or `Rack::Brotli` compression middleware is present in the Rack middleware stack. The endpoint never sends an HTTP response to the client, despite Rails logging `Completed 200 OK` and the Node renderer successfully returning data.

## Root Cause

The `RscPayloadController` uses `ActionController::Live` for streaming responses (via the `Stream` concern). Compression middleware like `Rack::Deflater` and `Rack::Brotli` typically have an `if` condition that checks the response body size before compressing:

```ruby
if: lambda do |env, *, body|
  sum = 0
  body.each { |i| sum += i.length }
  sum > 512
end
```

With `ActionController::Live`, the response body is a `Live::Buffer`. Calling `body.each` on a `Live::Buffer` **blocks** waiting for stream data to be written. But the controller action (running inside `Sync do` from the `async` gem) is also blocked waiting for the middleware to complete. This creates a **deadlock**: the middleware waits for body data, and the controller waits for the middleware.

## Steps to Reproduce

1. Create a Rails app with `react_on_rails_pro` and RSC enabled
2. Add `Rack::Deflater` or `Rack::Brotli` to the middleware stack (very common in Rails apps):
   ```ruby
   config.middleware.use(Rack::Deflater, ...)
   config.middleware.use(Rack::Brotli, ...)
   ```
3. Navigate to any RSC-enabled page or directly request `/rsc_payload/ComponentName`
4. The request hangs indefinitely — no response is sent

## Expected Behavior

The RSC payload endpoint should return the RSC payload JSON stream regardless of compression middleware.

## Workaround

Skip compression for RSC payload routes in the `if` lambda:

```ruby
if: lambda do |env, *, body|
  return false if env["PATH_INFO"]&.start_with?("/rsc_payload/")
  sum = 0
  body.each { |i| sum += i.length }
  sum > 512
end
```

## Suggested Fix

The gem should handle this automatically. Options:

1. **Set response headers** in the RSC payload controller that tell compression middleware to skip (e.g., `Content-Encoding` or `Cache-Control: no-transform`)
2. **Document the incompatibility** prominently in the RSC migration guide, since `Rack::Deflater` is extremely common
3. **Include a Rack middleware** in the gem that skips compression for `/rsc_payload/` routes

## Environment

- react_on_rails_pro: 16.4.0.rc.5
- Rails: 7.2
- Ruby: 3.4.8
- Puma: 7.1.0
- async gem: 2.35.0

justin808:
## Summary

- Sets `Content-Encoding: identity` on all streaming responses in the `Stream` concern to prevent `Rack::Deflater` and `Rack::Brotli` from iterating the `Live::Buffer` response body, which causes a deadlock with `ActionController::Live`

Fixes #2519

## Test plan

- [ ] Verify existing streaming specs still pass
- [ ] Test with `Rack::Deflater` in middleware stack — RSC payload endpoint should respond without hanging
- [ ] Test with `Rack::Brotli` in middleware stack — same expected behavior

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 3a633ea5f31ad961e8f180bbe4a9b704202f7621. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming deadlocks with HTTP compression middleware and prevented SSR streaming hangs and silent error absorption; streaming now sets encoding early and handles client disconnects gracefully.
* **Documentation**
  * Changelog updated with entries describing the streaming/compression and RSC payload fixes.
* **Tests**
  * Added tests covering Accept-Encoding behavior, Content-Encoding responses, and compressed vs. identity streaming.


justin808:
Closing in favor of #2530, which takes a better approach:

- **Opt-in `compress:` parameter** instead of always-on — only RSC payload opts in, doesn't affect other streaming endpoints
- **Clean `GzipOutputStream` class** instead of `define_singleton_method` monkey-patching
- **Proper `Accept-Encoding` parsing** via `Rack::Utils.q_values` instead of simple regex
- **Sets `Vary: Accept-Encoding`** for correct HTTP caching behavior
- **No `Content-Encoding: identity` fallback** — when compression isn't used, leaves headers untouched so Rack middleware can still compress non-streaming responses
- **Explicit `output_stream` plumbing** through `drain_streams_concurrently` instead of patching `response.stream` in place

AbanoubGhadban:
This was addressed via documentation in #2544 — the root cause is user-provided `:if` conditions on compression middleware calling `body.each` on streaming responses (a Rack SPEC violation), not a bug in React on Rails. See the comment on #2530 for full details.

PR #2565, opened 2026-03-08
Add test for empty async_props_block (fixes #2269)

### LINKED ISSUE #2269: `stream_react_component_with_async_props` doesn't render when an empty block is passed to it
A bug is a crash or incorrect behavior. If you have a debugging or troubleshooting question, please open [a discussion](https://github.com/shakacode/react_on_rails/discussions).

## Environment

1. Ruby version:
2. Rails version:
3. Shakapacker/Webpacker version:
4. React on Rails version:

AbanoubGhadban:
## Summary
- Adds a minimum reproduction test for issue #2269 where `stream_react_component_with_async_props` with an empty block caused a streaming deadlock
- Adds test view, controller action, route, and integration test for the empty block scenario
- This issue was already fixed by PR #2297 which changed fire-and-forget (`void`) calls to `await`ed promises in `handleIncrementalRenderStream.ts`

## Context

Issue #2269 reported that passing an empty block to `stream_react_component_with_async_props` would cause the stream to hang or return `nil`. The root cause was in `handleIncrementalRenderStream.ts` where `void onResponseStart()` and `void onRequestEnded()` were fire-and-forget — with an empty block, `request.close` fires immediately (END_STREAM) before the response was fully set up.

PR #2297 fixed this by changing to `await onRequestEnded()` and `await onResponseStartPromise`. This PR adds a regression test to ensure this scenario continues to work.

Closes #2269

## Test plan
- [ ] Integration test `completes without hanging when async_props_block emits nothing` passes with node renderer running
- [ ] Existing incremental rendering tests still pass

### DIFF
--- a/react_on_rails_pro/spec/dummy/app/controllers/pages_controller.rb
+++ b/react_on_rails_pro/spec/dummy/app/controllers/pages_controller.rb
@@ -48,6 +48,10 @@ class PagesController < ApplicationController # rubocop:disable Metrics/ClassLen
     stream_view_containing_react_components(template: "/pages/test_incremental_rendering")
   end
 
+  def test_empty_async_props_block
+    stream_view_containing_react_components(template: "/pages/test_empty_async_props_block")
+  end
+
   def cached_stream_async_components_for_testing
     stream_view_containing_react_components(template: "/pages/cached_stream_async_components_for_testing")
   end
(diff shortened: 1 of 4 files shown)


AbanoubGhadban:
Closing — the underlying bug was already fixed in PR #2297. The regression test here is useful but not critical enough to merge separately.

PR #2787, opened 2026-03-21
Document view helper pattern for disabling SSR in CI/test (#2786)

### LINKED ISSUE #2763: Add REACT_ON_RAILS_PRERENDER_OVERRIDE env var to globally disable prerendering
## Problem

There is no way to globally force prerendering off when views explicitly set `prerender: true` on individual components.

**Use case:** In CI/test environments (e.g., CircleCI), there is no SSR server available. When specs render views containing components with `prerender: true`, they throw errors. Setting `config.prerender = false` in the initializer only changes the **default** — it doesn't override component-level settings.

### Root cause

`RenderOptions#prerender` delegates to `retrieve_configuration_value_for`, which uses `options.fetch(:prerender) { config.prerender }`. When a view passes `prerender: true` explicitly, the config default is never consulted.

## Current Workaround

This monkeypatch in an initializer works today:

```ruby
# config/initializers/react_on_rails_test.rb
if Rails.env.test?
  ReactOnRails::ReactComponent::RenderOptions.prepend(Module.new do
    def prerender
      false
    end
  end)
end
```

### Option A: Document the monkeypatch (minimal approach)

Add a "Testing without an SSR server" section to the docs with the monkeypatch above as the recommended approach.

**Pros:**
- Zero gem code to maintain
- The monkeypatch is clean — standard Ruby `prepend` on a stable public method
- Keeps the gem simple; no new precedence rules to reason about

**Cons:**
- Technically brittle if `RenderOptions#prerender` changes signature or location
- Feels unofficial

### Option B: Add `REACT_ON_RAILS_PRERENDER_OVERRIDE` env var

A small change to `RenderOptions#prerender`:

```ruby
def prerender
  env_override = ENV["REACT_ON_RAILS_PRERENDER_OVERRIDE"]
  return env_override == "true" if env_override.present?

  retrieve_configuration_value_for(:prerender)
end
```

**Pros:**
- Official, supported solution
- Natural for CI — just set one env var
- Eliminates monkeypatch brittleness

**Cons:**
- Adds a hidden precedence layer (ENV > component option > config default) that can be surprising when debugging
- Requires documentation and tests
- Small but real maintenance surface in the gem

## Recommendation

Looking for input on which approach is preferred. Option A is lower cost; Option B is more user-friendly but adds complexity.

### LINKED ISSUE #2786: Document view helper pattern for disabling SSR in CI/test environments
## Context

Issue #2763 identified a real pain point: when CI/test environments lack an SSR server, components with explicit `prerender: true` throw errors. PR #2779 explored adding `config.prerender_override` but was closed because it adds config complexity for an unusual use case that is arguably an anti-pattern (disabling SSR hides SSR-specific bugs).

## Proposed documentation

Document a simple view helper pattern as the recommended approach:

```ruby
# app/helpers/prerender_helper.rb
module PrerenderHelper
  def use_prerendering
    return false if ENV["DISABLE_SSR"].present?
    true
  end
end
```

Then replace hardcoded `prerender: true` with the helper:

```ruby
react_component("MyComponent", prerender: use_prerendering)
```

### Why this is better than a gem-level config override

- **Zero gem changes** — no new config keys, no precedence rules, no maintenance burden
- **Per-component granularity** — SEO-critical components can still hardcode `prerender: true` if they must always SSR
- **Standard Rails pattern** — view helpers are idiomatic; any Rails developer understands this immediately
- **Users own the logic** — teams can customize (check `Rails.env.test?`, read different env vars, etc.)
- **Explicit** — the environment-awareness is visible at each call site, not hidden behind a global override

### Where to document

A short section under the prerender configuration docs (e.g., "Disabling SSR in CI or test environments") or in a troubleshooting/recipes section.

Closes #2763

justin808:
## Summary
- Add a "Disabling SSR in CI or Test Environments" section to `server-rendering-tips.md` showing a simple Rails view helper pattern
- Add a cross-reference entry in the troubleshooting guide's SSR section

Closes #2786
Closes #2763

## Test plan
- [ ] Verify markdown renders correctly on GitHub
- [ ] Verify internal link from troubleshooting guide resolves to the new anchor

---

> [!NOTE]
> **Low Risk**
> Documentation-only updates; no runtime code changes, so risk is limited to link/anchor correctness and potential doc confusion.
> 
> **Overview**
> Adds a new docs section describing a Rails view-helper pattern to *disable server-side rendering* in CI/test environments to avoid failures when the renderer isn’t available.
> 
> Updates the troubleshooting documentation to cross-reference this new guidance from the SSR-related troubleshooting section.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75c164b308fe1c7a7730398472a58013cbf8b2ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Documentation**
  * Added deployment guidance for disabling server-side rendering in CI and test environments to prevent failures when rendering infrastructure is unavailable.
  * Includes step-by-step instructions for configuring environment variables and component settings.
  * Added troubleshooting section documenting SSR-related errors in CI/test runs and remediation steps.

### DIFF
--- a/docs/oss/deployment/server-rendering-tips.md
+++ b/docs/oss/deployment/server-rendering-tips.md
@@ -20,31 +20,6 @@ For the best performance with Server Rendering, consider using [React on Rails P
 2. Set `config.trace` to true. You will get the server invocation code that renders your component. If you're not using Shakapacker, you will also get the whole file used to set up the JavaScript context.
 3. If streaming SSR requests hang indefinitely, check whether your compression middleware (`Rack::Deflater`, `Rack::Brotli`) has an `:if` condition that calls `body.each`. This causes deadlocks with streaming responses. See the [Compression Middleware Compatibility](../building-features/streaming-server-rendering.md#compression-middleware-compatibility) section in the Streaming Server Rendering guide.
 
-## Disabling SSR in CI or Test Environments
-
-If your CI or test environment lacks an SSR server (e.g., no Node.js runtime available), components with `prerender: true` will throw errors. Rather than adding gem-level configuration, use a standard Rails view helper to control prerendering per-environment:
-
-```ruby
-# app/helpers/prerender_helper.rb
-module PrerenderHelper
-  def use_prerendering
-    return false if ENV["DISABLE_SSR"].present?
-
-    true
-  end
-end
-```
-
-Then replace hardcoded `prerender: true` calls with the helper:
-
-```erb
-<%= react_component("MyComponent", prerender: use_prerendering) %>
-```
-
-Set `DISABLE_SSR=1` in your CI environment to skip server rendering.
-
-This approach gives you per-component granularity (SEO-critical components can still hardcode `prerender: true`), and you can customize the logic however you like — for example, checking `Rails.env.test?` or reading a different env var.
-
 ## CSS
 
 Server bundles must always have CSS extracted.
--- a/docs/oss/deployment/troubleshooting.md
+++ b/docs/oss/deployment/troubleshooting.md
@@ -370,12 +370,6 @@ rule.use = rule.use.filter((item) => {
 
 3. **No browser-only code in component?** (see "window is not defined" above)
 
-### "SSR errors in CI or test environments"
-
-**Symptoms:** Components with `prerender: true` throw errors because no SSR server is available
-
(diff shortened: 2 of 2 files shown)


justin808:
Closing — the team decided that disabling SSR in CI via a view helper is an anti-pattern. The root cause (ensuring an SSR server is available) should be fixed instead.

PR #2816, opened 2026-03-24
Add REACT_ON_RAILS_PRERENDER_OVERRIDE global prerender override

### LINKED ISSUE #2763: Add REACT_ON_RAILS_PRERENDER_OVERRIDE env var to globally disable prerendering

justin808:
## Summary
- add `REACT_ON_RAILS_PRERENDER_OVERRIDE` support in `RenderOptions#prerender`
- make env override precedence explicit: env > component option > config default
- add spec coverage for true/false overrides and invalid-value fallback
- document the env override in view helper docs

Closes #2763

## Testing
- `bundle exec rspec react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`
- `bundle exec rubocop react_on_rails/lib/react_on_rails/react_component/render_options.rb react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`

---

> [!NOTE]
> **Medium Risk**
> Introduces a new environment-variable-driven override that can force SSR on/off across all `react_component` renders, which could change rendering behavior in any environment where the variable is set. Default behavior is unchanged when the env var is absent, and invalid values fall back with a logged warning.
> 
> **Overview**
> Adds a global prerender override via `REACT_ON_RAILS_PRERENDER_OVERRIDE=true|false`, making prerender precedence **env > per-component option > initializer default**.
> 
> Implements a thread-safe cached parser for the env value (including normalization and one-time warnings on invalid values), expands `RenderOptions` specs to cover override/precedence behavior, and documents the new override in the view helper docs and `CHANGELOG`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit daa8421281ae7c61dd3ef984968acb97d1800cde. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a global prerender override via REACT_ON_RAILS_PRERENDER_OVERRIDE with precedence: environment variable > component option > initializer default.

* **Documentation**
  * Updated API docs to describe the new prerender override, expected values, normalization, and evaluation order.

* **Tests**
  * Added tests for env-var override precedence, normalization, invalid-value warning + fallback, and cache/reset behavior.

* **Chores**
  * Added an unreleased changelog entry documenting the change.

### DIFF
--- a/react_on_rails/lib/react_on_rails/react_component/render_options.rb
+++ b/react_on_rails/lib/react_on_rails/react_component/render_options.rb
@@ -15,8 +15,8 @@ module ReactOnRails
       PRERENDER_OVERRIDE_CACHE_MUTEX = Mutex.new
       class << self
         def prerender_env_override
-          raw_value = ENV.fetch(PRERENDER_OVERRIDE_ENV_KEY, nil)
           PRERENDER_OVERRIDE_CACHE_MUTEX.synchronize do
+            raw_value = ENV.fetch(PRERENDER_OVERRIDE_ENV_KEY, nil)
             cached_override = @prerender_env_override_cache
             return cached_override[:value] if cached_override && cached_override[:raw_value] == raw_value
 


justin808:
Addressed the blocking review items in latest push:\n\n- Added class-level caching of  parsing so invalid values no longer warn on every render path.\n- Kept cache keyed by raw env value so changed env values are re-parsed, while stable values avoid repeated warning/log overhead.\n- Scoped env cleanup into helper usage +  block, and added explicit config restore in  to avoid global-state leakage.\n- Updated invalid-value spec to assert warning fires once and added explicit no-env-override precedence coverage.\n\nLocal verification:\n- Inspecting 2 files
..

2 files inspected, no offenses detected\n- Run options: include {focus: true}

All examples were filtered out; ignoring {focus: true}
...........................

Finished in 0.00869 seconds (files took 0.82004 seconds to load)
27 examples, 0 failures

Run options: --seed 1121

# Running:



Finished in 0.000455s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips

justin808:
Correction to previous comment (shell formatting mangled markdown):

Addressed in latest push:
- Added class-level caching for `REACT_ON_RAILS_PRERENDER_OVERRIDE` parsing so invalid values do not warn on every render path.
- Cache is keyed by raw env value, so a changed env value is re-parsed while stable values avoid repeated warning/log overhead.
- Scoped env cleanup to helper usage and the `#prerender env override` block, with explicit config restore to avoid global-state leakage.
- Updated invalid-value spec to assert warning fires once and added explicit no-env-override precedence coverage.

Local verification:
- `bundle exec rubocop react_on_rails/lib/react_on_rails/react_component/render_options.rb react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`
- `bundle exec rspec react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`

justin808:
Follow-up push addresses the latest review pass:
- Fixed test helper lifecycle bug by clearing env/cache for the full `:prerender` assertions, not just object construction.
- Added cache reset in helper teardown to avoid cross-example state leakage.
- Added a missing spec for env override of config default when no explicit per-component option is set.
- Added mutex-protected access around class-level env override cache.

Re-verified locally:
- `bundle exec rubocop react_on_rails/lib/react_on_rails/react_component/render_options.rb react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`
- `bundle exec rspec react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`

justin808:
Addressed one additional thread-safety refinement in latest push (e8b16752): moved  inside the cache mutex so read/compare/write happen in one synchronized section.\n\nRe-verified locally:\n- Inspecting 2 files
..

2 files inspected, no offenses detected\n- Run options: include {focus: true}

All examples were filtered out; ignoring {focus: true}
.................................

Finished in 0.00571 seconds (files took 0.49942 seconds to load)
33 examples, 0 failures

Run options: --seed 34733

# Running:



Finished in 0.000407s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips

justin808:
Correction to previous comment (shell formatting stripped inline code):

Addressed one additional thread-safety refinement in latest push (e8b16752): moved `ENV.fetch(REACT_ON_RAILS_PRERENDER_OVERRIDE)` inside the cache mutex so read/compare/write happen in one synchronized section.

Re-verified locally:
- `bundle exec rubocop react_on_rails/lib/react_on_rails/react_component/render_options.rb react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`
- `bundle exec rspec react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb`

PR #3215, opened 2026-04-29
feat(pro): add imperative refetch API to RSCRoute (#3106)

### LINKED ISSUE #3106: Add imperative refetch API to RSCRoute: `ref` handle and `useCurrentRSCRoute` hook
## Context

Today, the only way to refetch an RSC mounted via `<RSCRoute>` is to call `useRSC().refetchComponent(componentName, componentProps)` (defined in `packages/react-on-rails-pro/src/RSCProvider.tsx`). The caller must supply **both** the component name and the exact props that were used to render it, because the internal cache key is built from `componentName + JSON.stringify(componentProps)` (see `packages/react-on-rails-pro/src/utils.ts`).

This works well for the "retry after fetch error" case: `ServerComponentFetchError` (at `packages/react-on-rails-pro/src/ServerComponentFetchError.ts`) carries the original props, so an error boundary can read them off the error and call `refetchComponent`. See the existing working example at `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentWithRetry.tsx`.

**The gap:** any other caller that wants to trigger a refetch — a parent/sibling refresh button, a retry UI rendered inside the server component's own subtree, a toolbar control, etc. — currently has no way to do it without being manually passed the props.

## Goal

Add **two** new ergonomic ways to refetch an `<RSCRoute>` instance without requiring the caller to know the component's props:

1. An **imperative ref handle** on `<RSCRoute>`. A parent (or any component that holds the ref) can call `ref.current.refetch()`.
2. A **`useCurrentRSCRoute()` hook** for client components that are rendered **inside** the server component's subtree. The hook returns a `refetch()` function tied to the nearest ancestor `<RSCRoute>`.

Both APIs should refetch the current RSCRoute instance using the component name and props it is currently rendered with — the caller should not need to know or pass them.

**Critical behavioral requirement:** Calling `ref.current.refetch()` or `useCurrentRSCRoute().refetch()` must cause the `<RSCRoute>` to **automatically re-render with the new content** from the server — the caller must not have to trigger any additional state change, `key` bump, remount, or manual re-render.

Background: the existing `useRSC().refetchComponent(name, props)` only replaces the internal promise in the provider's cache; it does **not** by itself cause the currently-mounted `<RSCRoute>` to re-render with the new payload. The existing retry example at `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentWithRetry.tsx` works around this by calling `setKey((k) => k + 1)` after refetching to force a re-render (see lines 43 and 57 of that file). The APIs added by this issue must **not** require that workaround — once the new RSC payload is available, the corresponding `<RSCRoute>` must refresh on its own.

## Public API

The feature should expose the following public surface from `react-on-rails-pro/RSCRoute`:

```ts
export type RSCRouteHandle = {
  /**
   * Force-refetches this RSCRoute instance using its currently-rendered
   * componentName and componentProps. Resolves with the new rendered
   * ReactNode. Rejects if the fetch fails.
   */
  refetch: () => Promise<ReactNode>;
};

export type RSCRouteProps = {
  componentName: string;
  componentProps: unknown;
  ref?: Ref<RSCRouteHandle>;
};

export function useCurrentRSCRoute(): RSCRouteHandle;
```

### Usage Example 1 — parent/sibling holds the ref

```tsx
import { useRef } from 'react';
import RSCRoute, { type RSCRouteHandle } from 'react-on-rails-pro/RSCRoute';

function Dashboard() {
  const cardRef = useRef<RSCRouteHandle>(null);

  return (
    <>
      <Toolbar>
        <button onClick={() => cardRef.current?.refetch()}>Refresh</button>
      </Toolbar>
      <RSCRoute
        ref={cardRef}
        componentName="UserCard"
        componentProps={{ userId: 123 }}
      />
    </>
  );
}
```

### Usage Example 2 — descendant inside the server component uses the hook

```tsx
// A client component registered in the client bundle, rendered by the
// server component inside an <RSCRoute>:
import { useCurrentRSCRoute } from 'react-on-rails-pro/RSCRoute';

export function InlineRefreshButton() {
  const { refetch } = useCurrentRSCRoute();
  return (
    <button onClick={() => refetch().catch(console.error)}>
      Refresh
    </button>
  );
}
```

The descendant does **not** need to know the server component's name or props — it just asks: "refetch the `<RSCRoute>` I'm rendered inside of".

### Behavior when used outside an `<RSCRoute>`

Calling `useCurrentRSCRoute()` outside of any `<RSCRoute>` ancestor should throw a clear error with the message:

```
useCurrentRSCRoute must be used inside an <RSCRoute>
```

## Affected Files

The following files are expected to be touched. This is guidance, not a prescription — the implementer is free to add new files if that's cleaner.

**Source (expected to change):**
- `packages/react-on-rails-pro/src/RSCRoute.tsx` — the component and the new hook live here.

**Source (for reference only, not expected to change):**
- `packages/react-on-rails-pro/src/RSCProvider.tsx` — exposes `useRSC()` and `refetchComponent`, which the feature builds on.
- `packages/react-on-rails-pro/src/ServerComponentFetchError.ts` — related error type.
- `packages/react-on-rails-pro/src/utils.ts` — where the cache key helper lives.
- `packages/react-on-rails-pro/package.json` — the `react-on-rails-pro/RSCRoute` export already exists; no new export entry should be required.

**Tests (new):**
- `packages/react-on-rails-pro/tests/RSCRoute.client.test.tsx` — new test file.
- Reference for existing test setup patterns: `packages/react-on-rails-pro/tests/registerServerComponent.client.test.jsx`.

**Dummy app (new + one edit):**
- `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentWithManualRefetch.tsx` — new example component.
- `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentRouter.tsx` — register the new example as a route (see how `ServerComponentWithRetry` is wired in this file for reference).
- Existing reference: `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentWithRetry.tsx` — the existing error-retry example, kept as-is.

**Docs:**
- `docs/pro/react-server-components/inside-client-components.md` — add a new section covering the two APIs.

**Changelog:**
- `CHANGELOG.md` — add an entry under `## [Unreleased]` → `#### Added`, tagged `**[Pro]**`.

> **Note:** `packages/react-on-rails-pro/src/RSCRoute.tsx` carries a Pro license header at the top. Preserve it.

## Note on `forwardRef`

As of **React 19 (released December 2024)**, the React team has announced that `forwardRef` will be deprecated in a future release, and function components can now receive `ref` as a regular prop. The implementer should decide whether to use `forwardRef` for this feature.

## Tests to Cover

Add tests for the following behaviors (see `packages/react-on-rails-pro/tests/registerServerComponent.client.test.jsx` for existing RSC test setup patterns — mock the fetcher, do not make real HTTP requests):

1. **Ref handle triggers a refetch.** Mounting `<RSCRoute ref={ref} ... />` and calling `ref.current.refetch()` results in a second fetch of the server component with the enforce-refetch behavior (same as the existing `refetchComponent`).

2. **Ref handle reflects the current props after a prop change.** When `componentProps` changes between renders, a subsequent call to `ref.current.refetch()` uses the **latest** props — not the props from when the ref was first assigned. This is a regression test; please do not skip it.

3. **`useCurrentRSCRoute` works from a client descendant.** A client component rendered inside an `<RSCRoute>` subtree can call `useCurrentRSCRoute().refetch()` and triggers a refetch of the correct RSCRoute instance with its current name + props.

4. **`useCurrentRSCRoute` throws outside any `<RSCRoute>`.** Calling the hook in a component that has no `<RSCRoute>` ancestor (but is still inside an `RSCProvider`) throws with the exact error message above.

5. **`ref.current` becomes `null` after unmount.** Standard React ref lifecycle — verify it still holds after these changes.

6. **Refetching automatically updates the rendered content.** Mount an `<RSCRoute>` and confirm the first payload is visible. Trigger `ref.current.refetch()` (or the hook equivalent in a separate test) and verify that — **without any external state change, `key` prop change, or remount by the test** — the `<RSCRoute>` eventually displays the content from the second (refetched) payload. This test exists specifically to guard against the stale-promise pitfall in the existing `refetchComponent` API; it must fail if the new APIs behave like the current workaround-dependent `refetchComponent`.

## Dummy App Example

Add a new file `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentWithManualRefetch.tsx` that demonstrates **both** patterns in a single page:

- A parent component with a `useRef<RSCRouteHandle>()` and a button that refetches via `ref.current?.refetch()` **only** — no surrounding `useState` / `setKey` workaround like the one in `ServerComponentWithRetry.tsx`. The example must visibly update with new content after the button click without any other caller-side action.
- An `<RSCRoute>` whose rendered subtree includes a small client component that calls `useCurrentRSCRoute().refetch()` on click.

You can reuse an existing registered server component (e.g. `SimpleComponent`) — a new server component is not required.

Wire it into `react_on_rails_pro/spec/dummy/client/app/components/ServerComponentRouter.tsx` as a new navigable route, following how `ServerComponentWithRetry` is wired in that file.

**Manual verification:**

```bash
cd react_on_rails_pro/spec/dummy && ./bin/dev
```

Navigate to the server component router, open the new example, and confirm that both the ref-based button and the hook-based button trigger an RSC payload refetch (check the browser Network tab) **and** that the rendered content visibly updates.

## Documentation

Update `docs/pro/react-server-components/inside-client-components.md` by adding a new `## Manually Refetching a Server Component` section that includes:

- Both usage examples from the Public API section above.
- A short "when to use which" paragraph explaining:
  - `ref` handle — when a parent or a component that shares an ancestor with the `<RSCRoute>` owns the refetch trigger.
  - `useCurrentRSCRoute()` — when the trigger lives inside the server component's own subtree (e.g., an inline refresh button rendered by the server component itself).
  - The existing `useRSC().refetchComponent(name, props)` — kept as-is, primarily for error-retry flows where the props come off a `ServerComponentFetchError`.

## Changelog

Add an entry in `CHANGELOG.md` under `## [Unreleased]` → `#### Added`, tagged `**[Pro]**`. Follow the existing entry style in that file: one-line summary, link to PR, link to this GitHub issue, author handle.

## Acceptance Criteria

- [ ] `<RSCRoute>` accepts an optional `ref` typed as `Ref<RSCRouteHandle>`, and `ref.current.refetch()` refetches the RSCRoute using its currently-rendered name and props.
- [ ] `useCurrentRSCRoute()` is exported from `react-on-rails-pro/RSCRoute` and returns `{ refetch }` when called inside an `<RSCRoute>` subtree.
- [ ] `useCurrentRSCRoute()` throws the exact error message specified above when called outside any `<RSCRoute>`.
- [ ] Both APIs use the **latest** rendered name/props — not stale values — when props change between renders.
- [ ] Calling `ref.current.refetch()` or `useCurrentRSCRoute().refetch()` causes the `<RSCRoute>` to re-render with the new content **automatically**. No caller-side `setState`, `key` change, remount, or other manual re-render is required. The new dummy example must also demonstrate this — its buttons should call `refetch()` directly, with **no** surrounding `setKey` / `useState` hack.
- [ ] All six test cases listed above pass.
- [ ] The new dummy app example demonstrates both patterns and is reachable from the server component router nav.
- [ ] Docs section added with both examples and the "when to use which" guidance.
- [ ] Changelog entry added under `[Unreleased]` with the `[Pro]` tag.
- [ ] `pnpm --filter react-on-rails-pro run test` passes.
- [ ] `pnpm --filter react-on-rails-pro run type-check` passes.
- [ ] `cd react_on_rails_pro && pnpm run eslint .` passes.
- [ ] `cd react_on_rails_pro && pnpm run prettier --check .` passes.
- [ ] The existing `refetchComponent(name, props)` API, cache-key logic, and error-retry flow are **unchanged**.

## Out of Scope

The following are **explicitly not** part of this task. They may be addressed in follow-up issues:

- A string-based `refetchKey` / `refetchByKey(key)` registry API for faraway or global callers (e.g. websocket-triggered refetches from outside React).
- A coarse "refetch all mounted RSCRoutes" API (similar to Next.js `router.refresh()`).
- Tag-based invalidation (similar to Next.js `revalidateTag`).
- Any changes to the existing `refetchComponent(name, props)` signature or behavior.
- Any changes to the `ServerComponentFetchError`-based retry flow or the example at `ServerComponentWithRetry.tsx`.
- Any changes to cache-key generation, the preloaded payload cache, or the fetch pipeline.
- Any changes to server-side RSC code.

## Dev Commands

From `react_on_rails_pro/CLAUDE.md`:

- JS tests: `pnpm --filter react-on-rails-pro run test`
- Type-check: `pnpm --filter react-on-rails-pro run type-check`
- Lint: `cd react_on_rails_pro && pnpm run eslint .`
- Prettier: `cd react_on_rails_pro && pnpm run prettier --check .`
- Dummy app: `cd react_on_rails_pro/spec/dummy && ./bin/dev`

AbanoubGhadban:
## Summary

Closes [Issue #3106](https://github.com/shakacode/react_on_rails/issues/3106). Adds two ergonomic ways to refetch a `<RSCRoute>` without the caller having to know its `componentName` or `componentProps`:

- **Ref handle** — `<RSCRoute ref={...} />` exposes `RSCRouteHandle` with `refetch()`. A parent or sibling can drive the refresh.
- **`useCurrentRSCRoute()` hook** — a client component rendered inside the server component's subtree can refetch the surrounding `<RSCRoute>` without being passed any props. Throws the spec-mandated message `useCurrentRSCRoute must be used inside an <RSCRoute>` if called outside one.

Both APIs auto-update the rendered tree — no caller-side `setKey`/`useState` workaround. The internal cache invalidation runs inside a React transition so old content stays visible while the new RSC payload streams in (no Suspense-fallback flash). Refetching propagates to every `<RSCRoute>` instance bound to the same cache key.

The existing `useRSC().refetchComponent(name, props)` API and the `ServerComponentFetchError`-based retry flow are unchanged — purely additive.

## Implementation highlights

- **`RSCProvider`** is now a real React component. The cache lives in a `useRef` (not reactive) and a per-cache-key `versions: Record<string, number>` map lives in `useState`. `refetchComponent` writes the new promise into the cache synchronously, then bumps `versions[key]` inside `startTransition`. Listing `versions` in the context value's `useMemo` deps is what propagates the re-render to every `useRSC()` consumer; the surrounding `startTransition` makes those re-renders transition commits.
- **`<RSCRoute>`** is converted to `forwardRef` and holds a `latestPropsRef` so a captured-then-stale `refetch` callback always reads the current name+props. It provides its handle via a private `CurrentRSCRouteContext` so descendants in the RSC subtree can read it via `useCurrentRSCRoute()`.

See [Issue #3106](https://github.com/shakacode/react_on_rails/issues/3106) for the full acceptance criteria.

## Stress demo

A new dummy-app page at `/server_router/refetch-stress` exercises 8 scenarios end-to-end:

1. Ref handle (parent triggers refetch).
2. `useCurrentRSCRoute()` from inside the RSC subtree.
3. Multi-instance fan-out (same name + same props → both update on one refetch).
4. Independent siblings (same name, different props → each refresh isolated).
5. Captured handle survives prop change (latest props win).
6. Rapid double-click (concurrent refetches).
7. Many siblings (5 distinct keys, refresh-all).
8. Mount/unmount cycle (`ref.current` null after unmount, set after re-mount).

None of the scenarios use a caller-side `setKey`/`useState` workaround — that is the whole point.

## Test plan

- [x] `pnpm --filter react-on-rails-pro run type-check` — passes.
- [x] `pnpm --filter react-on-rails-pro run test:non-rsc` — **44 / 44** pass (9 new, 35 pre-existing).
- [x] Unit tests cover all 7 acceptance criteria from the issue plus multi-instance fan-out plus a deterministic last-write-wins concurrency test under out-of-order resolution.
- [x] Webpack: all 3 bundles (client + server + RSC) compile clean.
- [x] Playwright e2e (`react_on_rails_pro/spec/dummy/e2e-tests/refetch-stress.spec.ts`) covers all 8 stress scenarios — **8 / 8** pass on Chromium against the local dev stack.
- [x] Root prettier + root ESLint + Pro ESLint — clean on all changed files.
- [x] Manual visual verification on the dummy app stress page.
- [x] Two-round AI code review against codex (gpt-5.4, high reasoning) — first round flagged 1 blocker (deterministic concurrency test) + 2 nits (docs note about error-boundary case, scenario-8 tautology fix); all addressed in commit `67500f13f`; round 2 verdict: **APPROVED**.

## Side fixes

- **`packages/react-on-rails-pro/jest.config.js`** — added `moduleNameMapper` to dedupe `react`/`react-dom` to the workspace root copy. Without it, hooks called from the Pro package source see a null React dispatcher because `@testing-library/react` resolves React from root (19.2.3) while the package source resolves it from `packages/react-on-rails-pro/node_modules` (19.2.0). Latent bug introduced by [PR #2155](https://github.com/shakacode/react_on_rails/pull/2155) when `react`/`react-dom` were added as devDependencies of the Pro package — the mismatch wasn't surfaced because the `test:non-rsc` jest pattern excludes every test that combines `@testing-library/react` rendering with hook-bearing source paths.
- **`packages/react-on-rails-pro/README.md`** — fixed pre-existing import-style bug (named-import syntax for what are actually default exports).

## Out of scope (worth follow-up issues, not this PR)

- Removing the duplicate React install at the project level (`pnpm.overrides` or dropping the redundant Pro devDep) so the `jest.config.js` mapper becomes unnecessary.
- Fixing the `test:non-rsc` regex hole that has been silently skipping `registerServerComponent.client.test.jsx`, `SuspenseHydration.test.tsx`, `streamServerRenderedReactComponent.test.jsx`, and others — including, ironically, several tests that would have caught the React-duplication issue in CI before this PR.
- `registerServerComponent.client.test.jsx` is independently broken since [PR #2074](https://github.com/shakacode/react_on_rails/pull/2074) (Nov 20, 2025) which renamed `testUtils.js` → `testUtils.ts` but didn't update the test's import path.

## Summary by CodeRabbit

* **New Features**
  * Imperative refetch for server-component routes via ref handles and a subtree hook to trigger refreshes.
  * Inline refresh button component and a Refetch Stress demo page.

* **Documentation**
  * Updated guides and API docs clarifying ref-based refresh vs. error-recovery flows and usage examples.

* **Tests**
  * Added unit and end-to-end tests covering imperative refetch, shared-cache fan-out, lifecycle, and stress scenarios.

* **Changelog**
  * New unreleased changelog entry documenting the feature.


justin808:
Opened #3552 as a maintainer-owned replacement because this branch is conflicting and `maintainerCanModify=false`. The replacement rebases this approved refetch API work onto current `main`, preserves the newer `RSCRoute ssr={false}` path, and is now waiting on fresh CI/review.

justin808:
Closing as superseded by #3552, the maintainer-owned replacement PR. #3552 rebased this imperative RSCRoute refetch work onto current main, addressed the blocking review/CI issues, and has now merged.

PR #3217, opened 2026-04-29
perf(pro): enable TCP_NODELAY on httpx (remove 40ms HTTP/2 tail)

AbanoubGhadban:
## Summary

- httpx (1.7.x) does not call \`setsockopt(TCP_NODELAY, 1)\` on its sockets, while net-http and async-http both do. With Nagle on, small HTTP/2 frame writes interact with Linux's delayed-ACK timer (\`tcp_delack_min = 40ms\`), producing a bimodal latency distribution with a second mode at ~40ms.
- This PR prepends a tiny monkey patch onto \`HTTPX::TCP#build_socket\` from inside react_on_rails_pro that enables \`TCP_NODELAY\` on every socket httpx opens. Loaded alongside \`require "httpx"\` in \`react_on_rails_pro/request.rb\`, so it applies before any session is constructed.
- The patch is self-removing once upstream httpx ships TCP_NODELAY by default.

## Why

The Rails → node-renderer transport uses httpx with `fallback_protocol: "h2"` and `persistent: true`. Every HTTP/2 frame is its own write, so Nagle + delayed-ACK aligns badly and pads ~9% of single-call samples by 40ms. Removing this restores parity with net-http/async-http.

## Measured impact

**Microbenchmark** (single thread, 800 persistent calls, h2):

| | p50 | p99 | max |
|---|---|---|---|
| httpx default | 6.63 ms | **44.80 ms** | 46.44 ms |
| **with TCP_NODELAY** | **1.35 ms** | **2.50 ms** | **3.06 ms** |

Bimodal 40ms tail eliminated entirely.

**Throughput** (4 threads, 20s load, h2 + Pro plugins):

| | p50 | p99 | RPS |
|---|---|---|---|
| httpx default | 5.31 ms | 45.20 ms | 426 |
| **with TCP_NODELAY** | 6.46 ms | **21.41 ms** | **559 (+31%)** |

**Page-level k6** (25 SSR calls/page, VU=4): p50 −5.7%, p99 −3.1%, +3.2% iterations. Page-level gain is smaller than the microbench because the connection pool (default 10) lets concurrent requests piggyback ACKs.

For reference, async-http h2 (which does set TCP_NODELAY) measures p50 0.84 ms / p99 3.55 ms — the patched httpx now matches it.

## Verification

End-to-end: with the patch loaded, `Socket#getsockopt(IPPROTO_TCP, TCP_NODELAY).int` on a socket built by `HTTPX::TCP#build_socket` returns `1`. Without it, `0`.

## Test plan

- [x] `rubocop` clean on changed files
- [x] Unit spec verifies the prepend ordering and that `setsockopt(IPPROTO_TCP, TCP_NODELAY, 1)` is invoked
- [x] Direct sanity check: `HTTPX::TCP.new(...).build_socket.getsockopt(...)` returns `1` (and `0` without the patch)
- [ ] CI green
- [ ] Manual smoke: hit the node-renderer from a Pro app and confirm no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * [Pro] Improved Rails→node-renderer transport performance by eliminating delayed-ACK latency (~40ms). Benchmarks show throughput ~426→559 RPS and p99 latency ~45ms→~21ms.
  * Applied a temporary transport-level patch to enable TCP_NODELAY for HTTP connections; remove once upstream fixes are available.

* **Tests**
  * Added test coverage validating the networking patch and its socket behavior.


AbanoubGhadban:
Not needed any more as we are migrating to `async-http`

PR #3461, opened 2026-05-28
Improve RSC missing-client-boundary diagnostics

justin808:
### Summary

Improves React on Rails Pro RSC diagnostics for missing client boundaries. Client-hook runtime failures now name the registered component and hook, preserve the original error through `cause`, keep the wrapper stack before the original stack, and point users to the missing `"use client";` directive instead of surfacing only a low-level `is not a function` error.

This also handles direct hook imports compiled as `(0, react_1.useState) is not a function`, plus `React.useState` and bare `useState` forms. The hook matcher now covers additional React 18/19 client hooks including `useId`, `useDebugValue`, `useInsertionEffect`, `useOptimistic`, and `useActionState`.

### Review updates

- Expanded client-hook diagnostic coverage for newer React hooks.
- Preserved the wrapping diagnostic stack before appending the original stack.
- Made `reportError`'s `Error` return contract explicit.
- Added coverage for direct hook imports, the `cause` chain, and the `throwJsErrors: false` stream metadata path.
- Left React 19 `use` out of the client-hook list intentionally: `use` is valid in Server Components, so diagnosing `use is not a function` as a missing client boundary would be misleading.

### Pull Request checklist

- [x] ~Update documentation~
- [x] Update CHANGELOG file

### Validation

- `NODE_CONDITIONS=react-server pnpm --filter react-on-rails-pro exec jest tests/serverRenderRSCReactComponent.rsc.test.tsx --runInBand`
- `pnpm run lint`
- `pnpm exec prettier --check packages/react-on-rails-pro/src/capabilities/proRSC.ts packages/react-on-rails-pro/tests/serverRenderRSCReactComponent.rsc.test.tsx`
- `git diff --check`

Note: `pnpm start format.listDifferent` currently fails only on the gitignored local Conductor attachment `.context/attachments/7UCCM9/PR instructions.md`; the PR files pass Prettier.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error diagnostics for React Server Components to provide clearer, more targeted error messages when client hooks are used in server components, including guidance on establishing proper component boundaries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shakacode/react_on_rails/pull/3461?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ After a release, run `/update-changelog` in Claude Code to analyze commits, writ
 
 ### [Unreleased]
 
+#### Fixed
+
+- **[Pro]** **RSC client-hook runtime errors now explain the missing client boundary**: React on Rails Pro now rewrites RSC runtime hook failures such as `useState is not a function` with a diagnostic that names the registered component, points to the likely missing `"use client";` directive, and clarifies that `.client`/`.server` suffixes only control bundle placement. Fixes [Issue 3184](https://github.com/shakacode/react_on_rails/issues/3184).
+
 ### [16.7.0.rc.3] - 2026-05-25
 
 #### Breaking Changes
--- a/packages/react-on-rails-pro/src/capabilities/proRSC.ts
+++ b/packages/react-on-rails-pro/src/capabilities/proRSC.ts
@@ -37,8 +37,28 @@ import loadJsonFile from '../loadJsonFile.ts';
 
 let serverRendererPromise: Promise<ReturnType<typeof buildServerRenderer>> | undefined;
 
-const CLIENT_HOOK_RUNTIME_ERROR_REGEX =
-  /\b(?:React\.)?(useState|useEffect|useReducer|useCallback|useMemo|useRef|useLayoutEffect|useImperativeHandle|useContext|useSyncExternalStore|useTransition|useDeferredValue) is not a function\b/;
+const CLIENT_HOOK_NAMES = [
+  'useState',
+  'useEffect',
+  'useReducer',
+  'useCallback',
+  'useMemo',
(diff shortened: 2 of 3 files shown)


justin808:
Addressed review feedback after scanning PR #3461 with no prior address-review summary cutoff.

### Mattered

- Addressed: client-hook detection now covers direct hook imports compiled as `(0, react_1.useState) is not a function`, plus bare and `React.` forms.
- Addressed: the diagnostic hook list now includes additional React 18/19 hooks: `useId`, `useDebugValue`, `useInsertionEffect`, `useOptimistic`, and `useActionState`.
- Addressed: enhanced errors now keep the wrapper stack before appending `Caused by:` and the original stack.
- Addressed: `reportError` now has an explicit `Error` return type.
- Addressed: tests now assert the `cause` chain, direct hook import behavior, a newer hook (`useInsertionEffect`), wrapper stack preservation, and the `throwJsErrors: false` metadata path.
- Left for discussion: the suggestion to use `new Error(message, { cause })` instead of manual `cause` assignment. The repo TypeScript config uses `lib: ["dom", "es2020"]`, so the current manual assignment keeps type-check compatibility without widening the TS lib target.

### Optional

- Addressed inline: changed the test component to direct hook import usage, which better matches real app code and exposed the transpiled import error shape.

### Skipped

- Skipped status-only bot comments: CodeRabbit draft-skip notice and size-limit report.
- Skipped adding React 19 `use` to the client-hook matcher: `use` is valid in Server Components, so diagnosing `use is not a function` as a missing client boundary would be misleading without more evidence.

Future full-PR scans should start after this comment unless `check all reviews` is requested.

PR #3550, opened 2026-06-02
Make benchmark reports smaller and more reliable

### LINKED ISSUE #3520: Make benchmark reports smaller and more reliable
- [x] **Decrease number of alerts**: p99 are most volatile and can probably be dropped, thresholds for the others increased.
- [ ] ~**Convert failures to internal check instead of a metric**: They are always near 0% or 100%, in the second case we should verify the status is always the expected one.~ This makes the code significantly more complicated without a sufficient benefit.

alexeyr-ci2:
## Summary

Three focused changes to make the Bencher benchmark reports smaller and to stop them from crying wolf. Today a single benchmark CI run files a false regression on nearly every commit, because ~100 benchmarks × several measures are each threshold-checked at a 95% boundary, and several routes are tracked that can never pass.

1. **Exclude intentional error-demo routes from auto-discovery.** `react_helmet_broken`, `context_function_return_jsx`, and `pure_component_wrapped_in_function` exist only to demonstrate mistakes (they're linked from the dummy apps' sidebar as "Improperly defined…" / "Incorrectly …") and hard-500 in the Pro suite. They're demos, not workloads, so a `NON_BENCHMARK_ROUTES` skip-list keeps route auto-discovery from sweeping them in. Explicit `ROUTES=` requests are still honored.

2. **Drop `p99` and `max` from the reports.** `p99` was a tracked Bencher measure (the noisiest latency percentile, a steady source of false-positive alerts); `max` was display-only and even noisier. Both are removed from the BMF output, the regression thresholds, the summary tables, and the underlying k6/Vegeta stat requests.

3. **Tune per-measure boundaries and drop `p90` as a tracked measure.** Bencher's t-test threshold is a prediction interval, so a one-sided boundary `B` gives a per-test false-positive rate of ~`(1 - B)`. Across all benchmarks per commit that compounds, so a sweep of recent `main` reports was used to pick per-measure boundaries that keep false regressions under ~1/20 commits: `rps` lower=`0.9995`, `p50_latency` upper=`0.9999`, `failed_pct` upper=`0.95`. `p90`'s tail noise can't meet that bar at any usable boundary, so it's no longer tracked (still shown in the summary table for visibility).

Fixes #3520.

## Test plan

- `bundle exec rspec benchmarks/spec` — 48 examples, 0 failures. New coverage: the per-measure boundary/side contract emitted to `bencher run`, `BmfCollector`'s emitted measure shape + `failed_pct` logic, the explicit-`ROUTES=` skip-list bypass, and a cross-check that the tracked measures exactly match what `BmfCollector` emits (a mismatch would silently disable a threshold).
- `bundle exec rubocop` — clean on all changed files.

## Summary by CodeRabbit

* **Benchmark Updates**
  * Simplified performance metrics collection: removed p99/max latency from benchmarks, now focusing on RPS, p50, and p90
  * Streamlined benchmark data collection and aggregation pipeline
  * Improved route discovery filtering to exclude intentionally skipped benchmark routes
  * Enhanced threshold configuration for per-measure boundary management

### DIFF
--- a/benchmarks/bench-node-renderer.rb
+++ b/benchmarks/bench-node-renderer.rb
@@ -309,8 +309,8 @@ non_rsc_tests.each do |test_case|
   rps, p50, p90, status = run_vegeta_benchmark(test_case, non_rsc_bundle)
   add_to_summary(test_case[:name], "non-RSC", rps, p50, p90, status)
 
-  # Add to BMF collector for Bencher output
-  bmf_collector.add(name: "#{test_case[:name]} (non-RSC)", rps: rps, p50: p50, p90: p90, status: status)
+  # Add to BMF collector for Bencher output (p90 stays in the summary table only)
+  bmf_collector.add(name: "#{test_case[:name]} (non-RSC)", rps: rps, p50: p50, status: status)
 end
 
 # Run RSC benchmarks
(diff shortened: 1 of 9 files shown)


justin808:
Superseded by #3551, which rebased these benchmark-report changes onto current main, added the requested BMF p50 normalization fix, passed review/CI, and has now merged.

Issue #3594, opened 2026-06-04
Follow-up: harden RSC CSS manifest delivery after #3587


justin808:
## Context

PR #3587 fixed RSC CSS FOUC by:

- patching `react-on-rails-rsc@19.0.4` inside this monorepo so the RSC client manifest records `css` files
- resolving those manifest CSS entries in React on Rails Pro and emitting `<link rel="stylesheet" precedence="ror-rsc">` from the RSC payload

Post-merge review comments surfaced a few useful follow-ups. The PR is merged and CI was green, so this issue is for hardening and downstream delivery rather than blocking the merge.

### 1. Make the manifest CSS patch reach downstream installs

The current fix relies on the root `pnpm.patchedDependencies` entry:

```json
"react-on-rails-rsc@19.0.4": "patches/react-on-rails-rsc@19.0.4.patch"
```

That patch applies inside this repo, but the published `react-on-rails-pro` package ships only `README.md`, `lib/**/*.js`, and `lib/**/*.d.ts`, and declares `react-on-rails-rsc` as an optional peer. Customer/generated apps will not automatically inherit this repo-level pnpm patch.

Options:

- Publish or require a `react-on-rails-rsc` version that includes the CSS-manifest fix, likely via shakacode/react_on_rails_rsc#35 or a later release.
- Update Pro/generator dependency pins so downstream apps install the fixed package.
- Remove the temporary root patch once the fixed upstream package is in use.
- Coordinate with #3497, but do not make the larger monorepo move a prerequisite for this smaller delivery fix.

### 2. Scope emitted CSS to rendered client references, or document the limitation

`resolveCssHrefs()` currently walks every entry in `filePathToModuleMetadata`, so every RSC render can emit stylesheets for all `'use client'` modules in the manifest, not just modules encountered by the rendered payload.

This is mostly safe for scoped CSS modules, but an unrendered client component with global CSS, or CSS modules with `:global` rules, could affect a page that never rendered that component.

Possible directions:

- Investigate whether we can derive the client references actually used by a render and emit only those CSS hrefs.
- Prefer an upstream/wrapper-based approach if the fixed `react-on-rails-rsc` package can attach CSS links at the client-reference boundary.
- If full scoping is not practical yet, explicitly document the limitation and add tests that make the current all-manifest behavior intentional.

### 3. Avoid synchronous manifest signature checks on every RSC render

[...]

Issue #35, opened 2015-09-21
Add trace message to console that indicates server side rendering happened


justin808:
![2015-09-21_11-46-28](https://cloud.githubusercontent.com/assets/1118459/10006066/7ce5aa1e-6056-11e5-8885-ce514f48bbd9.png)

1. Check if trace is turned on
2. Add a message to the console that comes from the server messages: 

Before line: https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/react_renderer.rb#L52
```ruby
      js_code_wrapper = <<-JS
  (function () {
    var result = '';
#{ReactOnRails::ReactRenderer.wrap_code_with_exception_handler(result_js_code, component_name)}
#{after_render}
    return result;
  })()
      JS
```

Using this history.console (https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/react_renderer.rb#L21)

```ruby
    CONSOLE_REPLAY = <<-JS
    var history = console.history;
    if (history && history.length > 0) {
      result += '\\n<script>';
      history.forEach(function (msg) {
        result += '\\nconsole.' + msg.level + '.apply(console, ' + JSON.stringify(msg.arguments) + ');';
      });
      result += '\\n</script>';
    }
```

samnang:
Fix in #38 .

PR #3595, opened 2026-06-04
fix(pro): make RSC manifest signatures async

justin808:
## Summary

Follow-up to PR #3587 and issue #3594.

- Moves RSC manifest file signature checks from synchronous `fs.statSync` calls to async `fs.promises.stat`, so manifest cache invalidation no longer does blocking filesystem work on the RSC render path.
- Makes the current RSC CSS limitation explicit and tested: `resolveCssHrefs()` works from the manifest alone, so it intentionally emits CSS for every `'use client'` manifest entry rather than only the client references encountered by one render.
- Includes the small review-adjacent cleanups from #3587: spreading stylesheet link children into the React fragment and resetting manifest loader singleton state in focused tests.

## #3594 Scope

This PR covers #3594 items 2, 3, and 4.

It does not yet close item 1. I checked `react-on-rails-rsc@19.0.5-rc.5`; the published tarball still does not include the upstream PR #35 CSS-manifest plugin changes, so bumping to that RC would not make downstream installs receive the manifest writer fix. That part should wait for a `react-on-rails-rsc` release that actually includes the upstream CSS metadata support, then Pro/generator pins can move to it and the temporary root patch can be removed.

## Test Plan

- `pnpm --filter react-on-rails-pro exec jest tests/manifestLoader.test.ts tests/resolveCssHrefs.test.ts --runInBand`
- `pnpm --filter react-on-rails-pro exec jest tests/resolveCssHrefs.test.ts --runInBand`
- `pnpm --filter react-on-rails-pro run test:rsc`
- `pnpm start format.listDifferent`
- `pnpm run lint`
- `cd react_on_rails && bundle exec rubocop`
- `cd react_on_rails_pro && bundle exec rubocop --ignore-parent-exclusion`

Labels: full-ci, benchmark. This touches Pro RSC render/cache behavior and removes blocking I/O from a server-rendering path, so broader CI plus benchmarks are appropriate.

---

> [!NOTE]
> **Medium Risk**
> Touches the Pro RSC streaming render path and manifest cache invalidation; behavior is mostly equivalent but async timing and manifest-wide CSS semantics affect styling and dev hot reload.
> 
> **Overview**
> This PR removes **blocking filesystem work** from the Pro RSC render path by making manifest cache signatures async (`fs.promises.stat` instead of `statSync`) and awaiting `setManifestFileNames()` inside each RSC stream initialization so dev rebuilds still invalidate derived caches correctly.
> 
> It also **documents and locks in** the current RSC CSS behavior: `resolveCssHrefs()` intentionally emits stylesheet links for **every** `'use client'` entry in the client manifest (not only client references used on a given render), with a new unit test and a docs section on FOUC prevention and the global-CSS caveat.
> 
> Smaller follow-ups: stylesheet `<link>` children are spread into the React fragment, manifest-loader tests reset singleton state in `afterEach`, and the Playwright FOUC test fetches SSR HTML via `request.get` before navigating the page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48fd69337bc50b1869efde3d82eacbdfd53c0dae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Release Notes

* **Documentation**
  * Added documentation section explaining how stylesheets are managed for React Server Components with `'use client'` boundaries, including stylesheet loading behavior and flash-of-unstyled-content prevention.

### DIFF
--- a/docs/pro/react-server-components/how-react-server-components-work.md
+++ b/docs/pro/react-server-components/how-react-server-components-work.md
@@ -110,6 +110,12 @@ Let's search for the client component `ToggleContainer` that we built before in 
 
 This entry indicates that the `ToggleContainer` client component is included in the `client25` chunk. The `js/client25.js` file contains the client-side code for the `ToggleContainer` component. You can find the `client25` chunk in the `public/webpack/<environment>/js/client25.js` file. Also, the `id` field is the Webpack module ID for the `ToggleContainer` client component. It's used by react runtime to load and hydrate the component in the browser.
 
+### CSS Links for `'use client'` Boundaries
+
+React on Rails Pro reads CSS metadata from the RSC client manifest and emits `<link rel="stylesheet" precedence="ror-rsc">` entries into the RSC payload. React hoists those links into `<head>` and delays committing the corresponding streamed content until the stylesheets are ready, which prevents a flash of unstyled content for client components behind RSC boundaries.
+
+The current resolver works from the manifest alone, so it emits CSS for every `'use client'` module entry in the manifest instead of only the client references encountered by one render. This is intentional for now: CSS modules and component-scoped styles remain safe, but global CSS imported by a client component can apply on an RSC page that did not render that component. If that matters for your app, avoid importing page-specific global CSS from broad client-component entry points until `react-on-rails-rsc` can attach CSS at the rendered client-reference boundary.
+
 If you want to change the file name of the `react-client-manifest.json` file, you can do so by setting the `clientManifestFilename` option in the `react-on-rails-rsc/WebpackPlugin` plugin as follows:
 
 ```js
--- a/packages/react-on-rails-pro/src/cache/manifestLoader.ts
+++ b/packages/react-on-rails-pro/src/cache/manifestLoader.ts
@@ -24,8 +24,16 @@ let manifestCacheKey: string | undefined;
 let clientRendererPromise: Promise<ReturnType<typeof buildClientRenderer>> | undefined;
 let rscCssHrefsPromise: Promise<string[]> | undefined;
 
-const buildManifestCacheKey = (clientManifest: string, serverClientManifest: string): string =>
-  `${getJsonFileSignature(clientManifest)}\0${getJsonFileSignature(serverClientManifest)}`;
+const buildManifestCacheKey = async (
+  clientManifest: string,
+  serverClientManifest: string,
+): Promise<string> => {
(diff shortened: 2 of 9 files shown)


justin808:
## Manual verification: change no longer present at HEAD — superseded by the #3860 revert ⏭️

I verified **net current behavior** (per the #3587/#3860 caveat in the batch issue) rather than the original PR, and the result is that **#3595's change is no longer in the codebase**. #3595 made the RSC manifest file-signature checks async (`statSync` → `fs.promises.stat`, `setManifestFileNames`/`buildManifestCacheKey`/`getJsonFileSignature` made async) as a follow-up to #3587 (#3594 items 2–4). #3860 ("Revert Pro-side RSC FOUC CSS wrapping from #3587", `c2803ceba`) then removed the whole mechanism #3595 had modified.

### Evidence (HEAD)

```text
# #3860 (c2803ceba) deletions/edits:
 packages/react-on-rails-pro/src/loadJsonFile.ts          |  22 +--   (getJsonFileSignature removed)
 packages/react-on-rails-pro/src/resolveCssHrefs.ts       |  83 ---   (deleted)
 tests/manifestLoader.test.ts                             |  62 ---   (deleted)
 tests/manifestLoaderServerRSC.rsc.test.ts                |  65 ---   (deleted)
 tests/resolveCssHrefs.test.ts                            | 181 ---   (deleted)

# At HEAD:
git grep getJsonFileSignature packages/   -> (none)
git grep buildManifestCacheKey packages/  -> (none)
grep -E 'statSync|promises.stat' src/loadJsonFile.ts -> (none; no stat call at all)
src/cache/manifestLoader.ts:25  export function setManifestFileNames(...): void   # synchronous again
```

`getClientRenderer()` now simply memoizes a single `clientRendererPromise`; there is no manifest-signature cache key and no per-render `stat` of the manifests. So the blocking-I/O-on-the-render-path concern from #3594 that #3595 addressed is moot in the current tree — but so is #3595's fix: it was swept up in the #3587 revert, not preserved independently.

### Conclusion

No before/after reproduction is possible or meaningful: the async manifest-signature path #3595 introduced was removed by #3860. If the manifest-signature cache-invalidation mechanism is reintroduced later (e.g. when the upstream `react-on-rails-rsc` CSS-manifest support lands, per this PR's #3594 item 1 note), the async-`stat` behavior should be re-applied and verified at that time. Flagging so the checklist reflects net current state rather than implying the async path is live.

PR #3504, opened 2026-05-31
Auto-mount Pro rolling deploy bundle routes

justin808:
Fixes #3476

Summary:
- Auto-mounts rolling deploy bundle routes for the built-in HTTP adapter when the adapter mount path is configured.
- Uses an internal helper prefix and prepended routes so existing manual mounts and catch-all routes remain safe.
- Documents the default auto-mounted workflow and manual mount escape hatch.

Verification:
- bundle exec rspec spec/rolling_deploy_auto_routes_spec.rb (in react_on_rails_pro/spec/dummy)
- bundle exec rspec spec/react_on_rails_pro/rolling_deploy/bundles_controller_spec.rb (in react_on_rails_pro)
- bundle exec rubocop --cache-root /private/tmp/rubocop_cache --no-server --cache false react_on_rails_pro/lib/react_on_rails_pro/engine.rb react_on_rails_pro/app/controllers/react_on_rails_pro/rolling_deploy/bundles_controller.rb react_on_rails_pro/spec/dummy/spec/rolling_deploy_auto_routes_spec.rb
- git diff --check
- autoreview --mode local

---

> [!NOTE]
> **Medium Risk**
> Changes deploy-time HTTP routing and upgrade behavior for Pro rolling deploy; security still relies on existing bearer-token and hash allowlist on the controller, with added coverage but no change to auth semantics.
> 
> **Overview**
> **React on Rails Pro** now **auto-mounts** the HTTP rolling-deploy bundle endpoints when `config.rolling_deploy_adapter` is the built-in `Http` adapter, using `config.rolling_deploy_mount_path` (default `/react_on_rails_pro/rolling_deploy`). Apps no longer need a default `draw_routes` entry in `config/routes.rb`; set the mount path to `nil` or `""` to opt out and keep manual routing.
> 
> The Pro engine **prepends** these routes and registers them under an internal named-route prefix so they win over catch-all routes and do not clash with existing manual mounts at the same path (upgrade-safe). Docs and changelog describe the default flow, custom paths, and secondary manual mounts with a distinct `as_prefix:`.
> 
> `BundlesController` route constraints now use a **non-anchored** `ROUTE_HASH_PATTERN` derived from `SAFE_HASH_PATTERN`, matching how Rails applies segment constraints, with controller-level validation unchanged. New request and unit specs cover auto-mount, opt-out, custom paths, invalid hash segments, and route ordering.
> 
[...]

PR #3611, opened 2026-06-04
fix(pro): polish rolling-deploy auto-route prefix + hash constraint (#3548)

### LINKED ISSUE #3548: Follow up on rolling deploy auto-route polish
Follow-up from #3504 after readying the PR. These are optional review nits rather than merge blockers:

- Consider referencing `ROLLING_DEPLOY_AUTO_ROUTE_PREFIX` in the auto-route specs instead of hardcoding the private route prefix string, or keep the hardcoded expectation with the existing intent-revealing comment.
- Consider preserving regex options when rebuilding `SAFE_HASH_PATTERN`-derived route constraints, e.g. by passing `SAFE_HASH_PATTERN.options` to `Regexp.new`.

No correctness blocker was found in the current PR; CI was green when this issue was created.

justin808:
## Summary

Addresses the two optional review nits raised in #3548 (follow-up from #3504, "Auto-mount Pro rolling deploy bundle routes"). Both are polish-only — no user-visible behavior change.

1. **Auto-route spec references the engine constant instead of a hardcoded literal.** `react_on_rails_pro/spec/dummy/spec/rolling_deploy_auto_routes_spec.rb` now reads `ReactOnRailsPro::Engine::ROLLING_DEPLOY_AUTO_ROUTE_PREFIX` via `const_get` (the constant is `private_constant`, so `const_get` is required) and uses it both for the manual reproduction of the engine auto-mount and for the helper-name assertion. The spec now stays pinned to the real prefix instead of a copy that could silently drift.

2. **`ROUTE_HASH_PATTERN` preserves `SAFE_HASH_PATTERN.options`.** `BundlesController` derives the route constraint from `SAFE_HASH_PATTERN` by stripping the `\A`/`\z` anchors; it now passes `SAFE_HASH_PATTERN.options` to `Regexp.new` so any future flags (e.g. case-insensitivity) stay in sync between the anchored defense-in-depth validation and the route constraint. Today the pattern carries no flags, so the derived regexp is byte-identical.

## Testing

- `bundle exec rspec spec/react_on_rails_pro/rolling_deploy/bundles_controller_spec.rb` → 31 examples, 0 failures
- `cd spec/dummy && bundle exec rspec spec/rolling_deploy_auto_routes_spec.rb` → 10 examples, 0 failures
- `bundle exec rubocop --ignore-parent-exclusion` on both changed files → no offenses

No CHANGELOG entry per the project's guidelines (test/refactor changes with no user-visible behavior change).

Closes #3548

---

> [!NOTE]
> **Low Risk**
> Test and internal regexp derivation only; no user-visible behavior change with current patterns.
> 
> **Overview**
> **Rolling-deploy route hash constraint:** `BundlesController` now builds `ROUTE_HASH_PATTERN` with `SAFE_HASH_PATTERN.options` (not just the stripped source), so any future regexp flags on the safe pattern also apply to the Rails `:hash` segment constraint. Behavior is unchanged today because both patterns use default options.
> 
> **Specs:** Auto-route examples read `ReactOnRailsPro::Engine::ROLLING_DEPLOY_AUTO_ROUTE_PREFIX` via `const_get` instead of a duplicated string, with one example that still pins the published literal so renames are caught. A controller spec asserts the route and safe hash patterns keep matching **options**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33a1b9fc2e6e69bd8313316801264b9ff1bf4de6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured rolling-deploy route pattern preserves regex options so route matching and internal validation remain consistent (prevents mismatches when flags like case-sensitivity are added).

* **Tests**
  * Updated specs to derive the rolling-deploy route prefix dynamically and added a regression test asserting regex-option consistency, improving test reliability and future-proofing.

### DIFF
--- a/react_on_rails_pro/app/controllers/react_on_rails_pro/rolling_deploy/bundles_controller.rb
+++ b/react_on_rails_pro/app/controllers/react_on_rails_pro/rolling_deploy/bundles_controller.rb
@@ -63,8 +63,10 @@ module ReactOnRailsPro
       # matcher applies segment constraints to the full segment. Derived from
       # SAFE_HASH_PATTERN by stripping the \A/\z anchors; the controller still
       # performs the anchored defense-in-depth validation before any filesystem
-      # lookup.
-      ROUTE_HASH_PATTERN = Regexp.new(SAFE_HASH_PATTERN.source.delete_prefix("\\A").delete_suffix("\\z"))
+      # lookup. Carries SAFE_HASH_PATTERN.options forward so any future flags
+      # (e.g. case-insensitivity) stay in sync between the two patterns.
+      ROUTE_HASH_PATTERN = Regexp.new(SAFE_HASH_PATTERN.source.delete_prefix("\\A").delete_suffix("\\z"),
+                                      SAFE_HASH_PATTERN.options)
 
       class << self
         # Helper for manual route mounts. The Pro engine uses these same route
--- a/react_on_rails_pro/spec/dummy/spec/rolling_deploy_auto_routes_spec.rb
+++ b/react_on_rails_pro/spec/dummy/spec/rolling_deploy_auto_routes_spec.rb
@@ -6,6 +6,9 @@ RSpec.describe "rolling-deploy auto routes", type: :request do
   let(:config) { ReactOnRailsPro.configuration }
   let(:default_path) { ReactOnRailsPro::Configuration::DEFAULT_ROLLING_DEPLOY_MOUNT_PATH }
   let(:token) { "a" * 32 }
+  # Engine keeps this prefix private; read it through const_get so the spec
+  # stays pinned to the real value instead of a copy that can silently drift.
+  let(:auto_route_prefix) { ReactOnRailsPro::Engine.const_get(:ROLLING_DEPLOY_AUTO_ROUTE_PREFIX) }
 
   around do |example|
     original_adapter = config.rolling_deploy_adapter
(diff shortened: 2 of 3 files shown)


justin808:
## Manual verification: skipped — no reproducible runtime symptom (non-behavioral) ⏭️

Per the `verify-pr-fix` protocol, a before/after reproduction adds nothing here because this PR is **polish / future-proofing with byte-identical behavior today** (as the PR description itself states). The two changes are:

1. The auto-route spec reads `ROLLING_DEPLOY_AUTO_ROUTE_PREFIX` via `const_get` instead of a duplicated literal — a test-quality change.
2. `ROUTE_HASH_PATTERN` now passes `SAFE_HASH_PATTERN.options` to `Regexp.new` so future flags stay in sync — a future-proofing change.

Because `SAFE_HASH_PATTERN` carries no flags today, the derived route pattern is identical with or without the fix. Verified against the real pattern (`/\A[A-Za-z0-9_][A-Za-z0-9_.\-]*\z/`):

```text
SAFE.options          = 0
PRE-FIX  route.options = 0   (matches SAFE? true)   # Regexp.new(source)
POST-FIX route.options = 0   (matches SAFE? true)   # Regexp.new(source, SAFE.options)
pre.source == post.source ? true
pre == post (byte-identical)? true
```

So the merged regression test (`ROUTE_HASH_PATTERN.options == SAFE_HASH_PATTERN.options`) passes against pre-fix code too — there is no failing "before" to capture. The value of the change is realized only if a flag (e.g. case-insensitivity) is added to `SAFE_HASH_PATTERN` later; at that point the constraint would silently diverge without this fix. That future condition is the residual risk, and it is the right thing for the unit test to guard rather than to reproduce here.

Routing this PR to the standard local-checks lane (`$verify`) rather than behavioral reproduction.

PR #3716, opened 2026-06-06
Guard Pro React root unmount errors

### LINKED ISSUE #3618: Pro: modern React 18+ Root-API unmount path (this.root?.unmount()) has no error handling
## Summary

In the Pro client renderer, `ComponentRenderer.unmount()` guards **only** the React 16/17 legacy unmount path with a try/catch. The modern React 18+ Root-API path (`this.root?.unmount()`) is **unguarded**, so a throw there is both unlogged and able to abort cleanup of every other component on the page.

This is the dominant path for modern React users (the legacy `!supportsRootApi` branch is dead on React 18+), so it is the more important of the two teardown paths to make robust.

## Details

`packages/react-on-rails-pro/src/ClientSideRenderer.ts`:

```ts
unmount(): void {
  // ...
  if (supportsRootApi) {
    this.root?.unmount();   // L188 — no try/catch
    this.root = undefined;
  } else {
    // L196-209: legacy path IS wrapped in try/catch (now console.error after #3610)
  }
}
```

`unmountAllComponents()` (same file, L329-332) iterates and then performs cleanup:

```ts
function unmountAllComponents(): void {
  renderedRoots.forEach((root) => root.unmount()); // L330
  renderedRoots.clear(); // L331
  resetRailsContext(); // L332
}
```

If `this.root?.unmount()` throws on the modern path, the exception propagates out of `ComponentRenderer.unmount()`, aborts the `forEach` at L330, and **skips** `renderedRoots.clear()` (L331) and `resetRailsContext()` (L332). One failed teardown poisons the whole page's cleanup — remaining components never unmount (memory leak) and the Rails context is never reset — and nothing is logged.

## Precedent — the non-Pro renderer already handles this

`packages/react-on-rails/src/ClientRenderer.ts` (L207-222) wraps **both** branches in a single try/catch:

```ts
function unmountAllComponents(): void {
  renderedRoots.forEach(({ root, domNode }) => {
    try {
      if (supportsRootApi && root && typeof root === 'object' && 'unmount' in root) {
        root.unmount(); // React 18+ Root API
      } else {
        unmountComponentAtNode(domNode); // React 16-17 legacy API
      }
    } catch (error) {
      console.error('Error unmounting component:', error);
    }
  });
  renderedRoots.clear();
}
```

After PR #3610, the Pro renderer makes the legacy path _more_ robust than the modern path — the opposite of what we want.

## Proposed fix

Wrap `this.root?.unmount()` in the same log-and-swallow pattern as the legacy branch, and clear the root reference in a `finally` so it is reset even on throw:

```ts
if (supportsRootApi) {
  try {
    this.root?.unmount();
  } catch (e: unknown) {
    const error = e instanceof Error ? e : new Error('Unknown error');
    console.error(
      `Caught error calling root.unmount() for domNode ${this.domNodeId}: ${error.message}`,
      error,
    );
  } finally {
    this.root = undefined;
  }
}
```

Optionally, consider routing teardown failures through `globalThis.reportError` (with a `console.error` fallback) to match the existing client-error pattern in `packages/react-on-rails-pro/src/handleRecoverableError.client.ts` (L27-31), so failures reach error-tracking backends rather than only the live console. A small test that stubs `root.unmount()` / `unmountComponentAtNode` to throw and asserts the error-level log would lock in the intended behavior for both branches.

## Context

Surfaced by the `/pr-review-toolkit:review-pr` pass on PR #3610 (item 2 of #3592). PR #3610 itself only changed the legacy path's log level (`console.info` → `console.error`) and intentionally did not touch this pre-existing, unrelated gap, per the project's "don't refactor unrelated code" rule. Filing here so it is not lost.

Related: #3592 (teardown follow-ups), #3610.

justin808:
## Summary

Fixes #3618.

Hardens the Pro `ClientSideRenderer` modern React root teardown path:

- wraps `this.root?.unmount()` in `try/catch`
- logs root unmount failures at error level instead of letting them abort cleanup
- clears `this.root` in `finally`
- adds a targeted regression proving one throwing root unmount does not stop other roots from unmounting or prevent later renders

## Validation

- `script/ci-changes-detector origin/main` -> Pro JS/TS category detected
- `pnpm --filter react-on-rails-pro exec jest tests/ClientSideRenderer.test.ts --runInBand` -> passed, 22 tests
- `pnpm --filter react-on-rails-pro run type-check` -> passed
- `git diff --check` -> clean

## Self-review

- Diff is limited to `packages/react-on-rails-pro/src/ClientSideRenderer.ts` and the targeted Pro client renderer test.
- Legacy teardown behavior is preserved; only the modern React root path is guarded.

---

> [!NOTE]
> **Low Risk**
> Localized change to Pro client teardown error handling; behavior on success is unchanged and aligns with existing guarded teardown paths.
> 
> **Overview**
> Hardens Pro **`ClientSideRenderer`** teardown on the modern React root API path so a failing **`root.unmount()`** no longer aborts bulk cleanup (fixes #3618).
> 
> **`unmount()`** now wraps **`this.root?.unmount()`** in **try/catch/finally**: failures are logged with **`console.error`** (including dom node id), and **`this.root`** is always cleared in **`finally`**. Renderer-owned mounts and the legacy **`unmountComponentAtNode`** path are unchanged.
> 
> A new **`ClientSideRenderer`** test asserts that when one framework root throws on unmount, **`unmountAll()`** still unmounts other roots, logs the error, and a later render on the same dom id still works.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9f56327dc74c5ab71fe3c8bafaec0312dfd6f59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Bug Fixes
* Improved error handling during React component unmounting to ensure cleanup always completes, even if an error occurs. Errors are now logged while allowing subsequent cleanup operations to continue, preventing potential state inconsistencies.

### DIFF
--- a/packages/react-on-rails-pro/src/ClientSideRenderer.ts
+++ b/packages/react-on-rails-pro/src/ClientSideRenderer.ts
@@ -296,8 +296,13 @@ You should return a React.Component always for the client side entry point.`);
     }
 
     if (supportsRootApi) {
-      this.root?.unmount();
-      this.root = undefined;
+      try {
+        this.root?.unmount();
+      } catch (e: unknown) {
+        console.error(`Error calling root.unmount() for dom node "${this.domNodeId}":`, e);
+      } finally {
+        this.root = undefined;
+      }
     } else {
       // Use the stored node first. During same-id replacement, document.getElementById(this.domNodeId)
       // already points at the new node, but the old legacy React tree is attached to this.domNode.
--- a/packages/react-on-rails-pro/tests/ClientSideRenderer.test.ts
+++ b/packages/react-on-rails-pro/tests/ClientSideRenderer.test.ts
@@ -575,6 +575,46 @@ describe('ClientSideRenderer', () => {
     expect(rootUnmount).toHaveBeenCalledTimes(1);
   });
 
+  it('logs and continues cleanup when a framework-created React root throws during unmount', async () => {
+    const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+    try {
+      const unmountError = new Error('root unmount boom');
+      const firstRootUnmount = jest.fn(() => {
+        throw unmountError;
+      });
+      const secondRootUnmount = jest.fn();
+      const thirdRootUnmount = jest.fn();
+      mockReactHydrateOrRender
+        .mockReturnValueOnce({ render: jest.fn(), unmount: firstRootUnmount })
+        .mockReturnValueOnce({ render: jest.fn(), unmount: secondRootUnmount })
+        .mockReturnValueOnce({ render: jest.fn(), unmount: thirdRootUnmount });
+      ComponentRegistry.register({
+        TestComponent: ({ greeting }: { greeting: string }) => React.createElement('div', null, greeting),
+      });
+      const firstComponentSpec = setupTestComponentDom('dom-id-root-unmount-throws');
+      const secondComponentSpec = setupTestComponentDom('dom-id-root-unmount-continues');
+      addRailsContext();
+
+      await renderOrHydrateComponent(firstComponentSpec);
+      await renderOrHydrateComponent(secondComponentSpec);
+
+      expect(() => unmountAll()).not.toThrow();
+      expect(firstRootUnmount).toHaveBeenCalledTimes(1);
+      expect(secondRootUnmount).toHaveBeenCalledTimes(1);
+      expect(consoleErrorSpy).toHaveBeenCalledWith(
+        'Error calling root.unmount() for dom node "dom-id-root-unmount-throws":',
+        unmountError,
+      );
+
+      await renderOrHydrateComponent(firstComponentSpec);
+
+      expect(mockReactHydrateOrRender).toHaveBeenCalledTimes(3);
+      expect(thirdRootUnmount).not.toHaveBeenCalled();
+    } finally {
+      consoleErrorSpy.mockRestore();
+    }
+  });
+
   // Complements the no-teardown renderer test: a renderer that opts out of cleanup still owns its own
   // mount, so unmount() must take the renderer-owned path and NOT unmount a React root (it never
   // created one). With the default mock, a non-renderer mount would set `this.root`; a renderer mount


justin808:
## Manual verification: reproduced the bug and confirmed the fix ✅

Reproduced via the PR's regression test (`tests/ClientSideRenderer.test.ts`) using the pre-fix-on-one-file tactic. Bug (#3618): on the modern React root path, a throwing `root.unmount()` propagated out of `ClientSideRenderer.unmount()` and aborted bulk cleanup, so one bad root stopped sibling roots from unmounting and blocked later renders.

### Reproduction

Squash-merged at `042d63414`, so pre-fix = `042d63414~1`. Because the file changed in other PRs since #3716 (type/behavior drift against HEAD), I pinned both the source and test to the #3716 era and reverted only the source to pre-fix, run transpile-only so assertions execute (throwaway config, removed after). Restored everything to HEAD afterward.

```
pnpm exec jest tests/ClientSideRenderer.test.ts   # (#3716-era files, transpile-only)
```

### Results

| Scenario | Behavior | Result |
| --- | --- | --- |
| Pre-fix `ClientSideRenderer.ts` | a throwing `root.unmount()` propagates; bulk teardown aborts (sibling roots not unmounted, replacement teardown skipped) | **4 failed / 22** (BROKEN) |
| Post-fix restored | `root.unmount()` wrapped in try/catch/finally; error logged via `console.error`, `this.root` cleared in `finally`, cleanup continues | **22 passed / 22** (FIXED) |

```text
# PRE-FIX (042d63414~1, transpile-only)
✗ logs and continues cleanup when a framework-created React root throws during unmount
    expect(received).not.toThrow()
    Error message: "root unmount boom"          ← the throw escapes unmount()
✗ does not unmount a React root for a renderer mount that returns no teardown
✗ runs the previous renderer teardown when the same dom id node is replaced
✗ unmounts the previous React root when the same dom id node is replaced
Tests: 4 failed, 18 passed, 22 total

# POST-FIX (restored)
Tests: 22 passed, 22 total
# HEAD full suite: 29 passed, 29 total
```

The target regression case fails because the unmount throw is uncaught; the three teardown cases are the same bug surfacing — one throwing root aborting the rest of cleanup.

### Caveat

Mechanism-level against the real `ClientSideRenderer.unmount()` in jsdom with a mocked throwing root. It confirms the try/catch/finally guard keeps cleanup going and clears `this.root`. I did **not** drive this through a live Pro client mounting/unmounting many real React roots in a browser. Type-checking was disabled only so the pre-fix file (which drifts from HEAD's types) could run; the merged code passes the suite with type-checking on (29/29).

PR #3720, opened 2026-06-06
fix(types): narrow server-component render-function inputs

### LINKED ISSUE #3589: Tighten renderer-function return types to remove load-bearing `as` casts (follow-up to #3576)
### Background

[PR #3576](https://github.com/shakacode/react_on_rails/pull/3576) (renderer-function mount cleanup, [#3209](https://github.com/shakacode/react_on_rails/issues/3209)) widened the `RenderFunction` return type so the 3-argument renderer form can return a `RendererTeardown`:

```ts
// packages/react-on-rails/src/types/index.ts
interface RenderFunction {
  (props?: any, railsContext?: RailsContext, domNodeId?: string): RenderFunctionResult | RendererResult;
  renderFunction?: true;
}
```

`RenderFunctionResult` covers the 2-argument server/client render-function form (component / server-render hash / promise); `RendererResult` covers the 3-argument renderer form (`void | RendererTeardown | Promise<…>`). Because both shapes now live in one union, call sites that only ever handle the 2-argument form need load-bearing `as` casts to narrow back, and those casts are **silent** — they suppress the type error rather than enforcing the documented invariant. If a future refactor breaks the invariant, the failure is a runtime crash, not a compile error.

This was raised in PR review (@claude, @greptile-apps) and intentionally deferred from #3576 because tightening it is a type-system change rather than part of the leak fix.

### Cast sites to tighten

1. **`packages/react-on-rails/src/createReactOutput.ts:79`**
   ```ts
   const renderFunctionResult = (component as RenderFunction)(props, railsContext) as RenderFunctionResult;
   ```
   The 3-argument renderer form is intercepted earlier in the client renderers and never reaches `createReactOutput`, so the result is always a `RenderFunctionResult` — but the `as RenderFunctionResult` cast asserts that rather than the types proving it. (Invariant is documented in the comment above the cast.)

2. **`packages/react-on-rails-pro/src/wrapServerComponentRenderer/client.tsx:67`**
   ```ts
   const Component = isRenderFunction(componentOrRenderFunction)
     ? ((await componentOrRenderFunction(props, railsContext, domNodeId)) as ReactComponent)
     : componentOrRenderFunction;
   ```
   A render function registered via `registerServerComponent` is expected to resolve to a component, not a teardown. The `as ReactComponent` cast accepts a `RendererTeardown` without a type error, and the downstream `typeof Component !== 'function'` guard can't catch it because a teardown is also a `function`. (Invariant is documented in the comment above the cast.)

### Proposed direction

Make the 2-argument render-function form and the 3-argument renderer form distinguishable at the type level so these casts become unnecessary — e.g. split into separate `RenderFunction` (2-arg → `RenderFunctionResult`) and `RendererFunction` (3-arg → `RendererResult`) types, or express the relationship with call-signature overloads. Then remove the `as RenderFunctionResult` / `as ReactComponent` casts at the two sites above so the invariants are compile-time enforced.

### Notes

- No runtime behavior change is expected; this is purely about restoring compile-time enforcement.
- Distinct from [#3578](https://github.com/shakacode/react_on_rails/issues/3578) (deferred #3209 checklist: converting in-tree dummy renderers, renderer-function docs, imperative `ReactOnRails.render(...)` leak).
- Source: review threads on PR #3576.

justin808:
## Summary

Fixes #3589.

- Add `ReactComponentRenderFunction` as the component-returning render-function type for APIs that require a React component result.
- Narrow the Pro `wrapServerComponentRenderer` client/server inputs to component-returning render functions and remove the load-bearing `as ReactComponent` casts.
- Keep the existing runtime guards for JavaScript callers and wrong-shaped results.
- Update the Pro dummy server-component registration cast and add a type assertion that renderer teardown functions are rejected at compile time.

## Validation

- `pnpm --filter react-on-rails run type-check`
- `pnpm --filter react-on-rails-pro exec jest tests/wrapServerComponentRenderer.types.test.ts --runInBand`
- `pnpm start format.listDifferent`
- `pnpm run lint`
- `script/ci-changes-detector origin/main fix/3589-renderer-function-types`
- `/Users/justin/.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main --no-web-search`

Labels: none. This is a focused TypeScript type-surface tightening with existing runtime guards preserved; path-based CI should cover the touched package areas.

## Release Notes

* **Refactor**
  * Refined TypeScript type definitions for the server component renderer to provide stricter compile-time type validation.
  * Replaced generic input type with a more specific union type for better type safety during component rendering.

* **Tests**
  * Added TypeScript type-checking tests for server component renderer to ensure type compatibility.

### DIFF
--- a/packages/react-on-rails-pro/src/wrapServerComponentRenderer/client.tsx
+++ b/packages/react-on-rails-pro/src/wrapServerComponentRenderer/client.tsx
@@ -24,7 +24,7 @@
 import 'react-on-rails-rsc/client.browser';
 import * as React from 'react';
 import * as ReactDOMClient from 'react-dom/client';
-import { ReactComponent, ReactComponentOrRenderFunction, RendererFunction } from 'react-on-rails/types';
+import type { ReactComponent, ReactComponentRenderFunction, RendererFunction } from 'react-on-rails/types';
 import isRenderFunction from 'react-on-rails/isRenderFunction';
 import { isRendererTeardownResult } from 'react-on-rails/@internal/rendererTeardown';
 import { ensureReactUseAvailable } from 'react-on-rails/reactApis';
@@ -33,6 +33,8 @@ import getReactServerComponent from '../getReactServerComponent.client.ts';
 import handleRecoverableError from '../handleRecoverableError.client.ts';
 
 ensureReactUseAvailable();
+
+type ServerComponentRendererInput = ReactComponent | ReactComponentRenderFunction;
 
 /**
  * Wraps a client component with the necessary RSC context and handling for client-side operations.
@@ -54,7 +56,7 @@ ensureReactUseAvailable();
  * ```
  */
 const wrapServerComponentRenderer = (
-  componentOrRenderFunction: ReactComponentOrRenderFunction,
+  componentOrRenderFunction: ServerComponentRendererInput,
   componentName: string = 'Unknown',
 ) => {
   if (typeof componentOrRenderFunction !== 'function') {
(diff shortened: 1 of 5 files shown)


justin808:
Current status: approved with no failing checks. The only remaining blocker is the GitHub Actions queued check `examples (4.0, latest)` in the Generator tests workflow: https://github.com/shakacode/react_on_rails/actions/runs/27071999965/job/79906592988. It has been queued since 2026-06-06T20:45:37Z. I will squash-merge once that check completes green and the merge state clears.

PR #4350, opened 2026-07-02
Treat nil hydrate_on as immediate

### LINKED ISSUE #4342: [Core] hydrate_on: nil raises ArgumentError instead of falling back to :immediate
## Summary

`RenderOptions` reads the new `hydrate_on` option (introduced this window in #4037) with `options.fetch(:hydrate_on, :immediate)`. `Hash#fetch` with a default only falls back when the **key is absent** — an explicit `hydrate_on: nil` returns `nil`, which `normalize_hydrate_on` then rejects with `ArgumentError: Invalid hydrate_on option: nil`. Passing `nil` for "use the default" is the common pattern when plumbing an optional value (e.g. `hydrate_on: some_conditional`), and every other render option treats an explicit `nil` as falsy/default rather than raising. Since `hydrate_on` is brand new, this papercut ships with the API's debut release.

## Affected code

- The `fetch` + normalize call: https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/react_on_rails/lib/react_on_rails/react_component/render_options.rb#L74
- `normalize_hydrate_on` (raises on `nil`): https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/react_on_rails/lib/react_on_rails/react_component/render_options.rb#L7-L22
- Contrast — config-backed options use `fetch` with a block and never raise on explicit `nil`: https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/react_on_rails/lib/react_on_rails/react_component/render_options.rb#L214-L218

## Introduced in

PR #4037 ("Add OSS hydrate_on scheduling"), merged 2026-06-18 (commit `1263119cd`). Confirmed via `git log --oneline -S "normalize_hydrate_on" 43ed0e138..origin/main -- react_on_rails/lib` — single hit.

## Mechanism

1. `RenderOptions#initialize` normalizes eagerly at construction time:

```ruby
@hydrate_on = ReactComponent.normalize_hydrate_on(options.fetch(:hydrate_on, :immediate))
```

2. `options.fetch(:hydrate_on, :immediate)` returns `nil` when the caller wrote `hydrate_on: nil` (key present, value nil) — the `:immediate` default applies only when the key is missing entirely.
3. `normalize_hydrate_on(nil)`: `nil` is not a String, so `normalized_value = nil`; `HYDRATE_ON_MODES` (`%i[immediate visible idle]`) does not include `nil`:

```ruby
unless HYDRATE_ON_MODES.include?(normalized_value)
  raise ArgumentError,
        "Invalid hydrate_on option: #{value.inspect}. " \
        "Supported OSS modes are :immediate, :visible, and :idle."
end
```

4. The `ArgumentError` propagates out of `react_component` at render time (a 500), not at boot.

## Failure scenario

```erb
<%# app/views/products/show.html.erb %>
<%= react_component("ProductGallery",
      props: @props,
      hydrate_on: below_the_fold? ? :visible : nil) %>
```

The author's intent — ":visible below the fold, otherwise the default" — raises `ArgumentError: Invalid hydrate_on option: nil` on every above-the-fold render. Same for any helper/wrapper that forwards an optional `hydrate_on:` keyword defaulting to `nil`. Every other option behaves fine under this pattern (e.g. `prerender: nil` acts as falsy; config-backed options simply skip their config fallback), so users have no reason to expect `hydrate_on` to be the one exception.

## Impact

- P1: runtime 500 for a natural usage pattern of a **new public API**, shipping in its first release (17.0.0 window). Fixing after release would technically change behavior (raise → default), so it is cheapest to fix before the API debuts.
- Inconsistent option semantics within the same options hash makes the API harder to wrap and to document.

## Suggested fix

Treat `nil` as "unset" so it falls back to the default, while still rejecting genuinely invalid values:

```ruby
# Option A (in normalize_hydrate_on):
def self.normalize_hydrate_on(value)
  return :immediate if value.nil?
  ...
end
```

or equivalently at the call site:

```ruby
# Option B (in RenderOptions#initialize):
@hydrate_on = ReactComponent.normalize_hydrate_on(options[:hydrate_on] || :immediate)
```

Option A is preferable: it fixes every caller of `normalize_hydrate_on`, keeps the strict error for typos (`hydrate_on: :visable` still raises), and `false` (also invalid) still raises under A whereas B would silently coerce it — choose deliberately and document.

## Test plan

- Spec: `RenderOptions.new(react_component_name: "App", options: { hydrate_on: nil })` → `hydrate_on == :immediate`, no raise.
- Spec: `hydrate_on: :visible` / `"idle"` unchanged; `hydrate_on: :bogus` still raises `ArgumentError`.
- Spec: omitted key still defaults to `:immediate` (existing coverage).
- Helper-level spec: `react_component("App", hydrate_on: nil)` renders without a `data-hydrate-on` attribute (same as omitting it), via `hydrate_on_data_attribute_value`.
- Run: `cd react_on_rails && bundle exec rspec spec/react_on_rails/react_component/render_options_spec.rb spec/dummy/spec/helpers/react_on_rails_helper_spec.rb`.

## Notes

- Verified against origin/main `5264192cc` by reading `render_options.rb:7-22,74` and tracing `Hash#fetch` semantics; contrasted with `retrieve_configuration_value_for` (`fetch` with block — explicit `nil` short-circuits to `nil`, never raises). Confidence: high.
- Pro note: `normalize_hydrate_on` also raises for non-`:immediate` values under Pro (`render_options.rb:17-21`); with the fix, `hydrate_on: nil` under Pro should resolve to `:immediate` and not raise, which is the desirable behavior for shared view code.
- Related finding (same review): "[Core] Boot-time RSC/Rspack validation spawns node on every Rails boot and hard-fails on undeterminable @rspack/core versions" (both are new-in-window API/validation hardening targeted at 17.0.0).

_Found in a post-2026-06-12 code review of merged changes (review base 43ed0e138). Filed by Claude Code on behalf of Justin Gordon._

justin808:
## Why

Fixes #4342. `hydrate_on: nil` should behave like the omitted option and use the default `:immediate` hydration mode, while invalid explicit values should still fail fast.

## What changed

- Normalize nil `hydrate_on` values to `:immediate`.
- Add render-options coverage for nil handling.
- Add helper-level coverage that nil does not emit a `data-hydrate-on` attribute.
- Add a user-visible `CHANGELOG.md` entry under `[Unreleased]` / `Fixed`.

## Verification

- RED: targeted render-options and helper examples failed before the implementation.
- `(cd react_on_rails && bundle exec rspec spec/react_on_rails/react_component/render_options_spec.rb spec/dummy/spec/helpers/react_on_rails_helper_spec.rb)`
- Targeted RuboCop for changed Ruby files.
- `.agents/bin/ci-detect`
- `script/ci-changes-detector origin/main`
- `pnpm exec prettier --check CHANGELOG.md` through the dependency-ready worktree
- `lychee --offline --include-fragments=none CHANGELOG.md` through the dependency-ready worktree
- `codex review --base origin/main`
- Pre-push hook before changelog: branch Ruby RuboCop and markdown link check.

## QA notes

- Release tracker #3823 is currently in Agent Release Mode: development, so standard main/beta PR gates apply.
- Local commit hooks for the changelog-only commit were bypassed after explicit validation because the local hook environment has a known `lychee`/`.lychee.toml` version mismatch and two sibling worktrees lacked a local `prettier` binary.

## Summary by CodeRabbit

* **Bug Fixes**
  * Passing `hydrate_on: nil` now works as expected and falls back to the default immediate hydration behavior instead of causing an error.
  * Components configured this way no longer include a hydration attribute when no mode is explicitly set.

* **Documentation**
  * Updated the changelog with the fix for `hydrate_on: nil` behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,12 @@ After a release, run `/update-changelog` in Claude Code to analyze commits, writ
 
 #### Fixed
 
+- **`hydrate_on: nil` falls back to immediate hydration**: Passing `hydrate_on: nil` now behaves
+  like the default `:immediate` mode instead of raising, while invalid explicit values still fail
+  fast. Fixes [Issue 4342](https://github.com/shakacode/react_on_rails/issues/4342).
+  [PR 4350](https://github.com/shakacode/react_on_rails/pull/4350) by
+  [justin808](https://github.com/justin808).
+
 - **[Pro]** **Gemfile loader source encodings are honored under C/POSIX locales**:
   The Pro Gemfile now loads its shared dependency fragments in binary mode, applies Ruby
   source-encoding magic comments or a UTF-8 default, and validates content before override-gem
--- a/react_on_rails/lib/react_on_rails/react_component/render_options.rb
+++ b/react_on_rails/lib/react_on_rails/react_component/render_options.rb
@@ -7,6 +7,8 @@ module ReactOnRails
     HYDRATE_ON_MODES = %i[immediate visible idle].freeze
 
     def self.normalize_hydrate_on(value)
+      return :immediate if value.nil?
+
       normalized_value = value.is_a?(String) ? value.to_sym : value
       unless HYDRATE_ON_MODES.include?(normalized_value)
         raise ArgumentError,
--- a/react_on_rails/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
+++ b/react_on_rails/spec/dummy/spec/helpers/react_on_rails_helper_spec.rb
@@ -702,6 +702,12 @@ describe ReactOnRailsHelper do
         expect(result).not_to match(/data-hydrate-on=/)
       end
 
+      it "does not add data-hydrate-on when hydrate_on is nil" do
+        result = react_component("App", hydrate_on: nil)
+
+        expect(result).not_to match(/data-hydrate-on=/)
+      end
+
       it "adds data-hydrate-on for explicit immediate mode" do
         result = react_component("App", hydrate_on: :immediate)
 
--- a/react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb
+++ b/react_on_rails/spec/react_on_rails/react_component/render_options_spec.rb
@@ -158,6 +158,12 @@ describe ReactOnRails::ReactComponent::RenderOptions do
       expect(opts.hydrate_on).to eq(:immediate)
     end
 
+    it "treats nil as the default immediate mode" do
+      opts = described_class.new(**the_attrs(options: { hydrate_on: nil }))
+
+      expect(opts.hydrate_on).to eq(:immediate)
+    end
+
     context "without React on Rails Pro" do
       before do
         allow(ReactOnRails::Utils).to receive(:react_on_rails_pro?).and_return(false)


justin808:
+ci-run-hosted

PR #4378, opened 2026-07-02
Handle fire-and-forget RSCRoute retry failures

### LINKED ISSUE #4330: [Pro npm] RSCRoute retry()/refetch() always rethrows in production recoverOnError mode, causing unhandled promise rejections in natural error-UI usage
## Summary

`RSCRoute`'s `refetch()` (and its `retry()` alias exposed for error-recovery UI) **always rethrows** the `ServerComponentFetchError` — including in production `recoverOnError` mode, where the failure has already been fully surfaced through `refetchError` state and the `onRefetchError` callback. The natural error-UI wiring, `onClick={() => retry()}`, therefore produces an **unhandled promise rejection on every failed retry** of an already-handled failure — noise that feeds `unhandledrejection` listeners and error trackers (Sentry, etc.). The JSDoc example does show `.catch(console.error)`, so this is a sharp-edge/API-ergonomics bug rather than a correctness bug — hence P3.

## Affected code

- The rejection handler that sets recovery state and then unconditionally rethrows:
  https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/packages/react-on-rails-pro/src/RSCRoute.tsx#L223-L248 (state set at L244, unconditional `throw` at L246)
- `retry` is a straight alias of `refetch` on the handle consumed by error UIs:
  https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/packages/react-on-rails-pro/src/RSCRoute.tsx#L256-L259
- The documented dual error surface that makes the rethrow redundant in recover mode (`refetchError` + `onRefetchError` effect):
  https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/packages/react-on-rails-pro/src/RSCRoute.tsx#L261-L265
- The JSDoc example that quietly relies on callers remembering `.catch`:
  https://github.com/shakacode/react_on_rails/blob/5264192ccf8897e4340380183d891feae908b94f/packages/react-on-rails-pro/src/RSCRoute.tsx#L124-L131

## Introduced in

The always-rethrow behavior predates the review window: it shipped with **PR #3786** ("Make RSCRoute refetch failures recoverable in production"), merged **2026-06-09** (commit 7810778cd) — three days before the review base (43ed0e138, 2026-06-12). Within the review window, **PR #4097** ("Bound the RSCProvider RSC payload cache with an LRU", merged 2026-06-21, commit 80dbe0378) reworked `RSCRoute.tsx` around this exact code path (retain/eviction integration) and carried the rethrow forward unchanged, which is when the review flagged it.

`git log --oneline 43ed0e138..origin/main -- packages/react-on-rails-pro/src/RSCRoute.tsx`:

```
80dbe0378 Bound the RSCProvider RSC payload cache with an LRU (#3564) (#4097)
```

## Mechanism

1. In production, `refetch()` runs with `recoverOnError = true`. On failure it records the error in state (which re-renders the error UI and fires `onRefetchError`) — and then rethrows anyway:

```tsx
// RSCRoute.tsx L235-L247
return rejectErrorPayload(refetchPromise).catch((error: unknown) => {
  const serverComponentFetchError = toServerComponentFetchError(error, n, p);
  if (
    recoverOnError &&
    isMountedRef.current &&
    latestRefetchRequestRef.current === requestId &&
    getRefetchVersion(n, p) === sharedRefetchVersion &&
    createRSCPayloadKey(...latestPropsRef.current) === requestKey
  ) {
    setRefetchErrorState([requestKey, serverComponentFetchError]);
  }
  throw serverComponentFetchError;   // <-- also thrown when the state path above fully handled it
});
```

2. `retry` is defined as an alias: `{ refetch, retry: refetch, refetchError, clearRefetchError }` (L257). The handle is explicitly documented for building error UIs around `refetchError`.

3. The idiomatic error-UI consumer treats `retry` as fire-and-forget, because the outcome is already observable via `refetchError`/`onRefetchError`:

```tsx
{refetchError && <button onClick={() => retry()}>Try again</button>}
```

4. When the retry fails again, the returned promise rejects with no attached handler → the browser dispatches `unhandledrejection` → error trackers record a duplicate report for a failure the component already displayed and already reported through `onRefetchError`.

## Failure scenario

Production Pro app with an RSCRoute error state (renderer briefly down, deploy in progress):

1. A refetch fails; `refetchError` renders the error UI; `onRefetchError` reports once (intended path).
2. The user clicks "Try again" (`onClick={() => retry()}`) while the backend is still down.
3. Every click produces: one intended `onRefetchError` report + one unhandled promise rejection for the same `ServerComponentFetchError`.
4. Error-tracking dashboards show doubled/noisy events (`Unhandled Promise Rejection: ServerComponentFetchError: ...`), and frameworks that install strict `unhandledrejection` handlers (or tests with `process.on('unhandledRejection')`) fail.

## Impact

- Noisy `unhandledrejection` events feeding error trackers for an already-handled failure; duplicate error counts; potential CI/test flakiness in consumer apps that fail on unhandled rejections.
- No data loss or broken recovery — the visible recovery UX works; this is API sharpness. P3.

## Suggested fix

Pick one (in order of preference):

1. **Resolve instead of reject in recover mode.** When `recoverOnError` is true and the error was recorded (the state path executed), resolve the promise with a sentinel result instead of rethrowing, and type the handle accordingly:

```ts
// RSCRouteHandle
refetch: () => Promise<RSCRefetchResult>; // { status: 'success'; node: ReactNode } | { status: 'error'; error: ServerComponentFetchError }
```

```ts
.catch((error: unknown) => {
  const serverComponentFetchError = toServerComponentFetchError(error, n, p);
  if (recoverOnError && /* existing staleness guards */) {
    setRefetchErrorState([requestKey, serverComponentFetchError]);
    return { status: 'error', error: serverComponentFetchError } as const;
  }
  throw serverComponentFetchError; // dev mode stays loud
});
```

   This is a breaking change to the handle's resolved type, so it may need a major/minor gate or a parallel method (`retrySafe()`).

2. **Non-breaking:** keep the rejection but pre-attach an internal no-op handler in recover mode so the rejection is never "unhandled" while still propagating to callers that chain `.catch`:

```ts
const result = rejectErrorPayload(refetchPromise).catch(/* existing handler */);
if (recoverOnError) {
  result.catch(() => {}); // mark handled; callers that await/catch still see the rejection
}
return result;
```

   Note: attaching a `.catch` to the *returned* promise marks the rejection handled globally while `await result` in caller code still throws — this preserves the documented contract with zero API change.

3. **At minimum:** document loudly on `RSCRouteHandle.retry`/`refetch` (and in `useCurrentRSCRoute` docs + the RSC docs page) that in error UIs the call must be `void retry().catch(() => {})` or awaited, because it rejects even when `refetchError` handles the failure.

## Test plan

- Unit test in `packages/react-on-rails-pro/tests/RSCRoute.test.tsx` (production-mode block, `process.env.NODE_ENV = 'production'`):
  - Failing refetch via the handle: assert `refetchError` is set AND (fix 2) that no `unhandledRejection` is emitted when the returned promise is not consumed — e.g. `process.once('unhandledRejection', spy)` + flush macrotasks, assert spy not called; or (fix 1) assert the promise resolves with `{ status: 'error' }`.
  - Assert dev-mode behavior unchanged: promise still rejects and no `refetchError` state is set.
  - Assert a caller that awaits `retry()` still observes the error (fix 2) or the sentinel (fix 1).
- Run: `cd packages/react-on-rails-pro && pnpm test RSCRoute`.

## Notes

- Verification: read current `RSCRoute.tsx` at origin/main (5264192cc); confirmed the unconditional `throw` at L246 after the recover-mode state set at L244, the `retry: refetch` alias at L257, and that the only in-repo usage guidance is the JSDoc `.catch(console.error)` example at L128. Confirmed via `git show 43ed0e138:.../RSCRoute.tsx` that the identical catch block predates the review base (#3786).
- Confidence: high on mechanism; classified P3 because the JSDoc example demonstrates the correct call pattern and recovery itself is unaffected.
- Related findings by title: "[Pro npm] Preloaded RSC payload arrays can only be streamed once (createRSCStreamFromArray push override)" — a failed preloaded-path load feeds this same error surface via `RSCRouteErrorBoundary`.

_Found in a post-2026-06-12 code review of merged changes (review base 43ed0e138). Filed by Claude Code on behalf of Justin Gordon._

justin808:
## Summary

Fixes #4330.

Production `RSCRoute` recover-on-error retries preserve the previous route content, but a fire-and-forget caller could still leave the returned retry promise unhandled after the route recorded the recoverable error. This keeps the caller-visible promise rejected while attaching an internal no-op rejection handler in production recover-on-error mode.

## Changes

- Preserve the existing `refetch`/`retry` rejected promise contract for callers that await or catch it.
- Add a production recover-on-error handler so fire-and-forget retry failures do not surface as unhandled rejections.
- Add a regression test that retries from inline controls without catching the retry promise and asserts no `unhandledRejection` is emitted.

## Validation

- `pnpm --filter react-on-rails-pro exec jest tests/imperativeRefetch.client.test.tsx --runInBand`
- `pnpm exec prettier --check packages/react-on-rails-pro/src/RSCRoute.tsx packages/react-on-rails-pro/tests/imperativeRefetch.client.test.tsx`
- `pnpm exec eslint packages/react-on-rails-pro/src/RSCRoute.tsx packages/react-on-rails-pro/tests/imperativeRefetch.client.test.tsx --no-warn-ignored`
- `script/check-pro-license-headers`
- `codex review --base origin/main`

No changelog entry: internal Pro client robustness fix with no public API change.

### DIFF
--- a/packages/react-on-rails-pro/src/RSCRoute.tsx
+++ b/packages/react-on-rails-pro/src/RSCRoute.tsx
@@ -232,7 +232,7 @@ const RSCRouteContent = forwardRef<RSCRouteHandle, Omit<RSCRouteProps, 'ssr'>>(
       // visible while the new promise streams in.
       const refetchPromise = refetchComponent(n, p, recoverOnError);
       const sharedRefetchVersion = getRefetchVersion(n, p);
-      return rejectErrorPayload(refetchPromise).catch((error: unknown) => {
+      const handledRefetchPromise = rejectErrorPayload(refetchPromise).catch((error: unknown) => {
         const serverComponentFetchError = toServerComponentFetchError(error, n, p);
         if (
           recoverOnError &&
@@ -245,6 +245,10 @@ const RSCRouteContent = forwardRef<RSCRouteHandle, Omit<RSCRouteProps, 'ssr'>>(
         }
         throw serverComponentFetchError;
       });
+      if (recoverOnError) {
+        void handledRefetchPromise.catch(() => undefined);
+      }
+      return handledRefetchPromise;
     }, [getRefetchVersion, refetchComponent]);
 
     const clearRefetchError = useCallback(() => {
--- a/packages/react-on-rails-pro/tests/imperativeRefetch.client.test.tsx
+++ b/packages/react-on-rails-pro/tests/imperativeRefetch.client.test.tsx
@@ -242,6 +242,30 @@ class CapturingErrorBoundary extends React.Component<
     );
   };
 
+  const FireAndForgetRetryControls: React.FC = () => {
+    const { refetch, refetchError, retry } = useCurrentRSCRoute();
+
+    return (
+      <div>
+        <button
+          type="button"
+          data-testid="inline-refetch"
+          onClick={() => void refetch().catch(() => undefined)}
+        >
+          inline-refetch
+        </button>
+        {refetchError ? (
+          <div data-testid="recoverable-error">
+            {refetchError.message}
+            <button type="button" data-testid="inline-retry" onClick={() => void retry()}>
+              retry
+            </button>
+          </div>
+        ) : null}
+      </div>
+    );
+  };
+
   beforeEach(() => {
     // Start each test with a sensible default; tests that need different
     // payloads call setupSequencedFetcher again.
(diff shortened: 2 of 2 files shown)


justin808:
+ci-status

justin808:
+ci-run-hosted

justin808:
Rebased onto `origin/main` `b194bb67f` after main advanced.

Post-rebase validation:
- `pnpm --filter react-on-rails-pro exec jest tests/imperativeRefetch.client.test.tsx --runInBand`

PR #4447, opened 2026-07-03
Fix Pro RSC loadable stats retry visibility

justin808:
## Why

Follow-up to #4401 review feedback. The Pro RSC stylesheet stats loader could permanently suppress repeated malformed/unreadable stats warnings, and native ESM stack resolution could choose `src/` when source maps rewrote stack frames away from the compiled `lib/` runtime directory.

## What changed

- Time-bound duplicate malformed-stats warning suppression using the existing max retry delay.
- Normalize source-map-rewritten `src/` stack-derived module directories back to sibling `lib/`.
- Remove dead `jest.dontMock` cleanup calls from the focused test.
- Add regression coverage for both review findings.

## Test plan

- `pnpm --dir packages/react-on-rails-pro exec jest tests/loadClientChunkStylesheetHrefs.test.ts --runInBand`
- `pnpm exec prettier --check packages/react-on-rails-pro/src/injectRSCPayload.ts packages/react-on-rails-pro/tests/loadClientChunkStylesheetHrefs.test.ts`
- `pnpm --dir packages/react-on-rails-pro run type-check`
- `script/check-pro-license-headers`
- `git diff --check`
- `.agents/bin/validate --changed` reached green for Docs Sidebar, RuboCop, ESLint, Prettier, TypeScript, Pro RuboCop, Pro RBS, and Pro TypeScript, then failed at the existing local Pro JS job because `bin/ci-local` runs `cd react_on_rails_pro && pnpm run nps test`, but current `react_on_rails_pro/package-scripts.yml` has no `test` script.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced repeated warning spam when stylesheet metadata can’t be read, so the same warning won’t appear too frequently.
  * Improved how server stack traces are interpreted, helping path lookups work correctly for different path formats.
  * Better handling of source paths in stack traces so module directories resolve more reliably.

* **Tests**
  * Expanded coverage for warning retry timing and stack-path normalization behavior.

### DIFF
--- a/packages/react-on-rails-pro/src/injectRSCPayload.ts
+++ b/packages/react-on-rails-pro/src/injectRSCPayload.ts
@@ -107,6 +107,7 @@ const REACT_SUSPENSE_REVEAL_SCRIPT = /\$RC\(/;
 const LOADABLE_STATS_FILE_NAME = 'loadable-stats.json';
 const LOADABLE_STATS_INITIAL_READ_RETRY_DELAY_MS = 100;
 const LOADABLE_STATS_MAX_READ_RETRY_DELAY_MS = 30_000;
+const LOADABLE_STATS_UNEXPECTED_WARNING_INTERVAL_MS = LOADABLE_STATS_MAX_READ_RETRY_DELAY_MS;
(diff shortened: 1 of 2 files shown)


justin808:
+ci-run-hosted

PR #4554, opened 2026-07-09
Docs: RSC CSS/FOUC findings register, Rspack FOUC verification procedure, vocabulary + ADR 0002

justin808:
## Summary

Stage 3 of the RSC CSS/FOUC research engagement. **Analysis and documentation only — no source or implementation changes** (boundary: nothing behavioral before the 17.0 final release).

Four deliverables:

1. **`internal/contributor-info/rsc-css-fouc-findings.md`** — corrected model of the FOUC pipeline (it is four layers, and the primary one is manifest-css → `preinit`-with-precedence hints, not the loadable-stats scraping the docs describe) plus a ranked risk register with `file:line` evidence.
2. **`internal/contributor-info/rsc-fouc-verification-procedure.md`** — a step-by-step procedure a human can run to confirm or deny the FOUC conditions on a real build, including the build-classification checks (manifest `css` arrays, chunk-id shape, loadable-stats sidecar) and a deterministic browser repro recipe.
3. **`packages/react-on-rails-pro/CONTEXT.md`** — the "CSS & FOUC" vocabulary finalized (provisional marker dropped): corrects the earlier "framework calls no `preinit`" claim (it calls `preinit` *with precedence*, which is the render-blocking mechanism) and resolves the July review's "batches"/"loader" ambiguity to a precise referent (**FOUC fallback layers**).
4. **`packages/react-on-rails-pro/docs/adr/0002-rsc-css-data-precedence-over-preinit.md`** — the CSS strategy met the ADR bar (hard to reverse: cross-package manifest contract; surprising without context: proven by the maintainer review; genuine alternatives: all-CSS-in-head, non-blocking preinit).

## Headline findings (detail + evidence in the register)

- **F1 (HIGH, both bundlers):** the loadable-stats inference and Suspense-reveal deferral layers are **silently inactive in every production-mode build** — the Flight regex requires quoted `"clientN"` chunk ids, but production `chunkIds: 'deterministic'` emits unquoted numbers. Every CI gate (including the Rspack FOUC e2e job) builds `NODE_ENV=test` (named ids), so the production posture is never exercised.
- **F2 (HIGH, Rspack):** manifest `css` arrays — the input of the primary mechanism — are silently omitted under Rspack on the 19.2.0 line entirely, and on 19.2.1-rc.0 when `output.publicPath` is `auto`/a function or CSS is extracted by anything other than `css/mini-extract`.
- **F3 (MEDIUM):** the 100 ms/first-stylesheet deferral window is the only mechanism found consistent with a "reproduced exactly once" flicker in dev/test builds; in production builds F1+F2 instead predict a **deterministic cold-cache flash** that the browser disk cache hides on every later attempt — which equally explains a one-time sighting.
- **F5 (docs):** `css-and-styling.md`'s claim that layout `stylesheet_pack_tag` still covers client CSS when the map is empty is false for async-chunk CSS, and `rspack-compatibility.md` (which it points to "for details") has no CSS content.

## Simplification priority (for a separate post-17.0 engagement)

1. Make the manifest-css hint layer unconditionally healthy (fix Rspack omission conditions; add a production-mode FOUC gate to CI).
2. Then delete the loadable-stats inference + reveal deferral (the only layers with silent env-dependent no-op behavior and `$RC` React-internals coupling — most of the 2,160-line `injectRSCPayload.ts`).
3. Reassess preload promotion; 4. drop the `loadable-stats.json` sidecar; 5. fix the two doc claims (docs-only, can be done now).

## Verification

- `pnpm exec prettier` clean; lefthook pre-commit (trailing-newlines, prettier, pro-license-headers, markdown-links/Lychee) passed on commit.
- No source files touched: diff is 2 new internal docs, 1 new ADR, 1 CONTEXT.md section.

## Summary by CodeRabbit

- **Documentation**
  - Added a new “RSC CSS / FOUC Pipeline” stage-3 documentation artifact, outlining a corrected four-layer reveal model (L1–L4) and the associated silent failure risks across production builds.
  - Updated the CSS & FOUC verification procedure with tighter stream-byte gating rules, revised expectations, and a recorded observed production failure.
  - Expanded core “CSS & FOUC” reference guidance and clarified the `data-precedence="rsc-css"` streamed boundary blocking approach.
  - Updated ADR coverage for per-boundary streamed RSC CSS ordering and fallback behavior when manifest hints are missing.


justin808:
Addressed the review in b178d896d — with one important refutation-turned-confirmation:

**On "L2 preload promotion may remain active in production" (CodeRabbit F1 comment, Greptile CONTEXT.md comment):** verified against Shakapacker 10.3.0 — it sets `chunkFilename: css/[id][hash].css` for both bundlers (`plugins/webpack.js:40`, `plugins/rspack.js:51`), and a production Rspack build of the dummy observably emits `css/4092-98880bc1.css`. L2's href filter only matches the dev/test `css/clientN-*.css` shape, so **L2 is also production-dead by default** (unless an app customizes `chunkFilename` to `[name]`). F1 is now restated with the two independent causes and the stronger conclusion: in a default production build, only the manifest-hint layer (L1) is live. The docs now say this instead of the ambiguous "all four layers inactive" phrasing you flagged.

Other fixes, all in b178d896d:
- **Dummy rsc version** (Greptile): correct — repo pins `19.2.1-rc.0`; the `19.2.0-rc.1` observation came from a stale install in a long-lived checkout. Reworded; the 19.2.0-line claim is now attributed to the published tarball.
- **Manifest glob / sidecar path** (Greptile P1/P2): replaced `public/*/` with concrete env-specific paths and documented where the staged sidecar lives.
- **B1 not guaranteed on 19.2.1+** : caveats for native-CSS extraction and split-chunks grouping added to Step 1.
- **Links + `$RC` ≠ proof of L3**: added an attribution subsection (shape heuristics + the definitive move-the-sidecar discriminator); also noted L3 links are impossible in production builds.
- **Blocked-CSS expectations**: now conditional on the active layer (React `$RR` gating vs parser script-blocking behind a preceding link — both keep content hidden, different contracts).
- **F3 race test prerequisite**: non-empty `loadable-stats.json` with `clientN` CSS is now required alongside B1=false/B2=named.
- **Buffer-count wording**: "three-buffer" → the five-buffer ordering (init → stylesheets → HTML → payload scripts → payload marks).

Also appended executed verification results (Step 1, 2026-07-09, Rspack 2.0.5 + rsc 19.2.1-rc.0 + production): **B1 css arrays present on the supported config; B2 numeric; CSS id-named.**

@coderabbitai resolve

justin808:
+ci-run-hosted

justin808:
Both matrix comments addressed in f025af4e2: rows now name the bundler (Rspack 2.x / 2.0.5 observed) separately from the react-on-rails-rsc package line, and the dev/test F3 row carries the Step 4 prerequisites (B1=false, named ids, CSS-bearing sidecar) with the all-layers-dead classification otherwise.

@coderabbitai resolve

PR #4591, opened 2026-07-12
Guard store re-initialization to prevent divergent state in deferred hydration (#4572)

### LINKED ISSUE #4572: hydrate_on: deferred hydration can bind a shared Redux store to a re-created instance → divergent state (#4037)
## Summary

With OSS `hydrate_on: :visible` / `:idle` scheduling (#4037), a component's mount is deferred to an arbitrary later time. Its late `getStore()` can bind to a Redux store instance that a *later* `forEachStore` pass unconditionally re-created, while an already-hydrated `:immediate` sibling still holds the earlier instance. Two components that share a store then dispatch to **two different store instances**, and their shared state silently diverges — no error, just wrong state.

## Where

`packages/react-on-rails/src/ClientRenderer.ts`

- `initializeStore` (`:115-121`) builds a fresh store from the generator and calls `StoreRegistry.setStore(name, store)` **unconditionally — there is no "already hydrated" guard**.
- `forEachStore` (`:123-127`) re-runs `initializeStore` for **every** `[data-js-react-on-rails-store]` element, and it is invoked on **every** `renderComponent` (`:507`), `renderAllComponents` (`:525`), and `reactOnRailsComponentLoaded` call.
- Deferred mount path: `scheduleHydration` (`:211`) → `runScheduledRender` (`:468`) → `mountReactRoot` (`:476`), whose `getStore` resolves the store lazily at mount time.
- The `sameNode` guard (`:395-397`) skips re-rendering an already-mounted component, so an already-hydrated component is *not* rebound when the store is later recreated.

## Root cause

Before `hydrate_on`, all components on a page hydrated in one synchronous pass, so every component read the same store instance created in that pass — the per-render store re-creation in `forEachStore` was effectively idempotent from the component's point of view. Deferred hydration breaks that invariant: a deferred component's `getStore` can now run on the **far side** of a store re-creation triggered by a later render, while an already-hydrated sibling keeps the original instance. `initializeStore`'s unconditional `setStore` is the pre-existing enabler; #4037's deferral is what makes two live instances observable.

## Failure scenario

OSS with Pro installed (non-`:immediate` modes require it):

1. A page has component **A** (`hydrate_on: :immediate`) and component **B** (`hydrate_on: :visible`) that share Redux store `S`.
2. Initial `renderAllComponents()` builds instance `S1`; **A** hydrates and dispatches against `S1`. **B** is scheduled (not yet mounted).
3. A Turbo Frame / async fragment injects another RoR component and calls `ReactOnRails.reactOnRailsComponentLoaded(...)`, which re-runs `forEachStore` and overwrites the registry with a **brand-new `S2`**. **A** is not re-rendered (skipped by the `sameNode` guard), so it keeps `S1`.
4. **B** scrolls into view, calls `getStore('S')`, and binds to `S2`.

Result: **A** dispatches to `S1`, **B** dispatches to `S2` — updates from one never reach the other. Repro sketch: OSS dummy app, one `redux_store` shared by an `:immediate` and a `:visible` `react_component`; after load, trigger a second `reactOnRailsComponentLoaded(...)` (or a Turbo-frame load) before scrolling B into view; assert `A.store !== B.store`.

## Impact

A real data-integrity bug: silently divergent shared state between components with no error surfaced. The trigger is narrow — it needs mixed `:immediate` + deferred components sharing a store **and** a second render / store re-creation between the two hydrations — but `hydrate_on` is a new feature that makes exactly this timing reachable, and when it hits, debugging is hard because nothing throws.

## Suggested fix

- Guard `initializeStore` to skip re-initializing a store whose name already has a hydrated instance (only create if absent, or if props/railsContext actually changed), OR
- Initialize stores once per page-load rather than per `renderComponent` / `forEachStore` call.

Either way the invariant to restore is: a given store name resolves to one instance for the lifetime of a page render, regardless of when a component hydrates.

## Attribution / verification

- Introduced in-window by #4037 (`1263119cd`, "Add OSS hydrate_on scheduling"); the unconditional `initializeStore` (`:119-120`) is the pre-window enabler, the two-live-instances divergence is newly reachable in-window.
- Verified by static audit + refutation pass against `main` @ `b6674a56b` (confirmed `forEachStore` unconditionally recreates every store, no guard, and the deferred `mountReactRoot` binds the store late; `sameNode` skips rebinding the immediate sibling). Not runtime-reproduced.
- Related: #4037, #4374; ISSUE-26 (a latent hydrate-fallback issue in the same #4037 scheduling code).

AbanoubGhadban:
## Summary

- Adds a guard in `initializeStore` to skip re-creating a store that already exists in `hydratedStores`
- Restores the one-instance-per-name-per-page-lifetime invariant broken by deferred hydration

## Problem

Before `hydrate_on` scheduling (#4037), all components hydrated in one synchronous pass, so every component read the same store instance. Deferred hydration (`visible`/`idle`) breaks this invariant:

1. Initial `renderAllComponents()` creates store S1
2. Component A (`:immediate`) hydrates and gets S1
3. Component B (`:visible`) is scheduled but doesn't mount yet
4. A Turbo Frame or async fragment calls `reactOnRailsComponentLoaded()`, which re-runs `forEachStore` 
5. `initializeStore` unconditionally creates a new store S2, overwriting S1 in the registry
6. Component A is skipped by the `sameNode` guard, so it keeps S1
7. Component B scrolls into view and mounts with S2

Result: A dispatches to S1, B dispatches to S2 — divergent shared state with no error.

## Fix

Check if the store already exists in `hydratedStores` before creating a new one. This matches the Pro package's behavior which already guards via the `storeRenderers` map.

## Test plan

- [x] Added unit tests covering:
  - `forEachStore` running multiple times doesn't re-create stores
  - Store instance remains stable between immediate and deferred component hydration
- [x] All existing ClientRenderer tests pass (363 tests)
- [x] Type check passes
- [x] Lint passes

Fixes #4572

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate Redux store creation during repeated rendering passes within the same page lifecycle.
  * Ensured immediately hydrated and deferred/visible components consistently reuse the same store instance.
  * Added page-unload cleanup so hydrated stores are cleared on navigation, allowing correct re-initialization on the next page.
* **Tests**
  * Added coverage for the store re-initialization guard, deferred hydration timing, and page-unload store clearing.

### DIFF
--- a/packages/react-on-rails/src/ClientRenderer.ts
+++ b/packages/react-on-rails/src/ClientRenderer.ts
@@ -114,6 +114,19 @@ function teardownErrorLabel(entry: RenderedEntry, domNodeId: string): string {
 
 function initializeStore(el: Element, railsContext: RailsContext): void {
   const name = el.getAttribute(REACT_ON_RAILS_STORE_ATTRIBUTE) || '';
+
+  // Guard: don't re-create an already-hydrated store.
+  // Before hydrate_on scheduling (#4037), all components hydrated in one synchronous pass, so every
+  // component read the same store instance created in that pass. Deferred hydration (visible/idle)
+  // breaks that invariant: forEachStore can run again (e.g. via reactOnRailsComponentLoaded for a
+  // Turbo Frame), re-creating the store while an already-hydrated :immediate sibling keeps the old
+  // instance and a not-yet-mounted :visible sibling will get the new one — divergent shared state.
+  // Skipping re-initialization restores the one-instance-per-name-per-page-lifetime invariant.
+  // See: https://github.com/shakacode/react_on_rails/issues/4572
+  if (StoreRegistry.getStore(name, false) !== undefined) {
+    return;
+  }
+
   const props = el.textContent !== null ? (JSON.parse(el.textContent) as Record<string, unknown>) : {};
   const storeGenerator = StoreRegistry.getStoreGenerator(name);
   const store = storeGenerator(props, railsContext);
--- a/packages/react-on-rails/src/ClientRenderer.ts
+++ b/packages/react-on-rails/src/ClientRenderer.ts
@@ -570,5 +570,16 @@ function unmountAllComponents(): void {
   renderedRoots.clear();
 }
 
+/**
+ * Clear all hydrated stores on page unload so navigation to a new page re-initializes stores
+ * with the new page's props. Without this, the initializeStore guard (which prevents duplicate
+ * store creation within a single page lifecycle) would skip re-initialization on navigation,
+ * leaving components bound to stale state from the previous page.
+ */
+function clearAllStores(): void {
+  StoreRegistry.clearHydratedStores();
+}
+
 // Register cleanup on page unload
 onPageUnloaded(unmountAllComponents);
+onPageUnloaded(clearAllStores);
--- a/packages/react-on-rails/tests/ClientRenderer.test.ts
+++ b/packages/react-on-rails/tests/ClientRenderer.test.ts
@@ -1453,5 +1453,68 @@ describe('ClientRenderer', () => {
       const observerGlobal = globalThis as unknown as { IntersectionObserver?: unknown };
       delete observerGlobal.IntersectionObserver;
     });
+
+    it('clears stores on page unload so navigation re-initializes with new props', () => {
+      const storeGenerator = jest.fn((props: Record<string, unknown>) => ({
+        getState: () => ({ count: (props as { initialCount: number }).initialCount }),
+        dispatch: jest.fn(),
+        subscribe: jest.fn(),
+      }));
+      StoreRegistry.register({ NavigationStore: storeGenerator });
+      setupStoreElement('NavigationStore', { initialCount: 1 });
+
+      const TestComponent: React.FC = () => React.createElement('div', null, 'Test');
+      ComponentRegistry.register({ TestComponent });
+
+      const componentElement = document.createElement('div');
+      componentElement.className = 'js-react-on-rails-component';
+      componentElement.setAttribute('data-component-name', 'TestComponent');
+      componentElement.setAttribute('data-dom-id', 'nav-store-test');
+      componentElement.textContent = JSON.stringify({});
+      document.body.appendChild(componentElement);
+
+      const targetNode = document.createElement('div');
+      targetNode.id = 'nav-store-test';
+      document.body.appendChild(targetNode);
+
+      // First page load: creates the store
+      renderComponent('nav-store-test');
+      expect(storeGenerator).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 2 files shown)


AbanoubGhadban:
+ci-run-hosted

AbanoubGhadban:
## Addressed review feedback

@greptile-apps raised a valid concern about stale stores surviving navigation. Fixed in d0015dbb5:

Added `clearAllStores()` registered with `onPageUnloaded()` to clear hydrated stores on Turbo/Turbolinks navigation. This ensures:

1. **Within a page**: Stores are stable (the guard prevents re-creation by `forEachStore`)
2. **Across navigation**: Stores are fresh (cleared on unload, re-created with new page's props)

This matches the Pro package behavior where `unmountAllStores()` clears the `storeRenderers` map on page unload.

Added test: `clears stores on page unload so navigation re-initializes with new props`

AbanoubGhadban:
## Addressing review comments

All automated reviewers (Greptile, Claude, Codex) flagged the same issue on the first commit: stores surviving Turbo navigations. This is now fixed in the second commit (d0015dbb5) which adds `clearAllStores()` on page unload.

### Summary of behavior:
- **Within a page**: Stores are stable (guard prevents re-creation by `forEachStore`)
- **Across Turbo navigations**: Stores are fresh (cleared on unload, re-created with new page's props)
- **Turbo Frames**: If a frame replaces a store element with new props *without* a full navigation, the store keeps its existing state. This is intentional — it maintains the one-instance invariant within a page lifecycle. Apps needing mid-page store refresh should use Redux actions/dispatch rather than relying on store re-initialization.

### Re: documentation suggestion
Good point about documenting this behavior. This is primarily relevant to the new `hydrate_on` feature. I'll add a note to the hydration scheduling docs in a follow-up if maintainers agree it's needed.

AbanoubGhadban:
### Re: Codex P2 (Turbo Frame partial replacement)

The P2 comment raises a valid edge case: when a Turbo Frame replaces content (without a full page unload), the store keeps its existing state rather than re-initializing with new props.

This is **intentional** and correct for the bug being fixed:

1. **The bug (#4572)**: Deferred components (`hydrate_on: :visible`) can bind to a *different* store instance than immediate siblings if `forEachStore` runs between their hydration times.

2. **The fix**: Ensure all components on a page see the *same* store instance throughout the page's lifecycle.

3. **Turbo Frame behavior**: A Turbo Frame replacement is *not* a page navigation — it's a partial DOM update within the same page. The store should remain stable so that:
   - Already-hydrated components keep their state
   - Newly-inserted components bind to the same store
   - No state divergence occurs

Apps that need to update store state when a Turbo Frame loads should:
- Dispatch Redux actions to update the existing store
- Or trigger a full Turbo Drive navigation if they need fresh server-rendered store props

This matches the Pro package behavior where stores are stable within a page and cleared only on full navigation.

PR #4718, opened 2026-07-17
Fix knip duplicate-export failure breaking lint on main

justin808:
## Why

Main is red: the `Lint JS and Ruby` workflow fails on `pnpm exec knip` since #4711 merged ([failing run](https://github.com/shakacode/react_on_rails/actions/runs/29615362951/job/87999199163)). knip's duplicate-exports rule flags `MAX_EXTERNAL_SOURCE_MAP_BYTES = MAX_INLINE_SOURCE_MAP_BYTES` in `vmSourceMapSupport.ts` as one value exported under two names. A red lint gate on main fails every PR branch, so this blocks the whole open-PR queue.

## What

Route both source-map size caps through a private shared constant (`MAX_SOURCE_MAP_BYTES`) so the two exports stay distinct symbols with the same value. No behavior change; the intentional unit difference (string length vs bytes-on-disk) documented in the surrounding comments is preserved.

## Test plan

- `pnpm exec knip` → exit 0 (was exit 1 with `Duplicate exports (1)`)
- `pnpm exec knip --production` → exit 0
- `pnpm exec jest tests/vmSourceMapSupport.test.ts` in `packages/react-on-rails-pro-node-renderer` → 57 passed
- Prettier + ESLint clean on the changed file

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized the maximum supported size for inline and external source maps.
  * Preserved separate limits for each source map type while ensuring they use the same size ceiling.

### DIFF
--- a/packages/react-on-rails-pro-node-renderer/src/worker/vmSourceMapSupport.ts
+++ b/packages/react-on-rails-pro-node-renderer/src/worker/vmSourceMapSupport.ts
@@ -122,8 +122,12 @@ const sourceMapLookupAttempts = new WeakSet<SourceMapLookupAttempt>();
 let warnedMissingSourceMapConstructor = false;
 
 const MAX_MISSING_SOURCE_MAP_RETRIES = 5;
+// Shared ceiling for both source-map size gates below. Kept private so the two
+// exported caps stay distinct symbols (knip flags re-exporting one value under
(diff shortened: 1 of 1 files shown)


justin808:
+ci-run-hosted

Issue #4385, opened 2026-07-02
Investigate selective/independent hydration for traditional (non-RSC) React components


AbanoubGhadban:
## Summary

Investigate whether React on Rails can support **selective (independent) hydration** for **traditional (non-RSC) React components** — allowing different parts of a page to hydrate independently, without waiting for the full page HTML to download and without downloading the entire JS bundle up front, and **without relying on the RSC payload**.

## Motivation

Today, hydration of traditional (client + SSR) React components generally requires:

1. The full page (or at least the relevant HTML) to be available, and
2. The relevant JS bundle to be loaded before the component becomes interactive.

RSC (React Server Components) already gives us streaming + progressive/selective hydration semantics via the RSC payload. But many apps use classic `react_component` / SSR components and cannot (or do not want to) adopt RSC. We'd like to explore bringing "hydrate this island as soon as its HTML and its code are available" behavior to those traditional components as well.

The goal is **island-style / progressive hydration** for classic components:

- Each registered component (or a subset marked as such) hydrates independently as soon as:
  - its server-rendered HTML has streamed in, and
  - the code chunk it needs is available,
- so the browser does not need to wait for the whole page download or a single monolithic bundle before any island becomes interactive.

## Scope of Investigation

- [ ] Determine whether React (18/19) primitives (`hydrateRoot`, streaming SSR, `Suspense`, selective hydration) can be leveraged for **independent per-component hydration** without RSC.
- [ ] Assess how this interacts with React on Rails' current hydration flow (`ReactOnRails.reactOnRailsPageLoaded`, the client startup that scans the DOM for rendered components, store hydration, etc.).
- [ ] Investigate code-splitting integration: can each island load only its own chunk (e.g. via dynamic `import()` / async component registration) so we avoid downloading the whole bundle?
- [ ] Investigate streaming HTML: can islands hydrate as their HTML arrives (streamed responses) rather than waiting for full-page load?
- [ ] Identify interactions/conflicts with existing features: shared Redux stores, `railsContext`, deferred/immediate hydration, `prerender`/SSR vs client-only, and the Pro streaming/RSC code paths.
- [ ] Document the constraints, limitations, and browser/React-version requirements.

## Deliverable / Decision

- If it turns out to be **feasible and reasonably easy** to enable within React on Rails, implement it as part of this issue.
- Otherwise, it is sufficient to **document the findings and constraints here** and open a **follow-up issue** for the actual implementation.

## Notes

[...]

PR #4769, opened 2026-07-22
Scroll-priority streaming: architecture investigation and recommendation (#4385 follow-on)

justin808:
## Summary

Design doc for **zero-waste scroll-priority delivery of cached streamed HTML sections** — the production-architecture question left open by the `selective_hydration_cached` prototype (#4740, issue #4385): when a user scrolls toward a not-yet-arrived Suspense section, deliver it ASAP with **zero duplicate bytes**, selective hydration + component state intact, on CDN-cached chunks, without taxing the normal streamed path.

One file: `docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md`.

## What's in it

- **Verified fact base**: the prototype's empirical facts, re-validated; new findings from shipped `react-dom` builds (18.3.1 / 19.2.7 / 19.2.8 tarballs) and `facebook/react` history; an adversarially-verified research pass (3-vote verification; 18 surviving primary-source claims, 7 refuted) over Cloudflare DO pricing/placement, Service-Worker lifetime rules, and HTTP/2/3 abort semantics.
- **Comparison matrix** of six candidates (DO edge-state, SW stream-stitching, gated `window.stop()`, no-stop pull baseline, official `prerender`/`resume`, pull-only "server islands" mode).
- **Recommendation**: layered — manifest + byte-stable per-section assets (external Fizz runtime for strict CSP) → SW stream-stitching with heartbeat + truncation-recovery as the Scenario-A transport → React's stable `prerender`/`resume` as the Scenario-B pipeline, making the skip/priority signal a stateless request parameter. DO kept as the Cloudflare-native alternative; `window.stop()` rejected.
- **Prototype plan P0–P8**, each step with an experiment and pass/fail criteria.
- **Sources**, split into adversarially-verified / first-hand-verified / fetched-but-unverified / explicitly-open.

## Findings that changed the brief's assumptions

1. `$RC`/`completeBoundary` was **rewritten in React 19.2** (batched async reveals via `$RB`/`$RV`, ~300 ms throttle) — the brief's quoted source is the older sync body; duplicate-safety survives (intentionally hardened in facebook/react#33511/#33531) but manual-`$RC` timing assumptions don't.
2. The official partial-prerender family (`prerender`, `resume`, `resumeToPipeableStream`, `resumeAndPrerender`) is **stable in react-dom 19.2.8** — the exact line the Pro dummy pins — so "shell now, tail later on demand" is framework-sanctioned.
3. `unstable_externalRuntimeSrc` makes cached chunks **inline-script-free** (instructions as `data-rci` templates + MutationObserver runtime), dissolving the CSP-nonce-rewriting problem and manual-reveal internals dependence at once (runtime file must be vendored; contract-tested).
4. The "Chromium kills SWs at 5 minutes while streaming" folklore **failed verification** (MV3-extension-scoped); the real, verified budget is Firefox's 30 s + 30 s, mitigated by a production-proven heartbeat. Both bfcache claims also failed verification — measured in the plan, not assumed.

## Decision points for reviewers

- Agree SW stream-stitching (not the Durable Object) is the **default** transport, with DO as a documented Cloudflare-native alternative?
- First-visit stance: plain stream (keeps bytes ×1.0 absolute) vs opt-in no-stop pull accelerator?
- Green-light P0–P3 (manifest + SW prototype in the dummy app) as the first implementation slice?
- Appetite for the external-runtime (`unstable_`-prefixed) path as the strict-CSP answer?

Related: #4385, #4740.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design specification for “zero-waste” scroll-priority delivery of cached streamed HTML sections using React selective hydration.
  * Covers delivery scenarios (edge fully cached vs origin rendering), recommended layered architecture, and manifest-based byte-stable section delivery.
  * Includes production constraints, security/CSP guidance, failure-mode considerations, and a structured prototype plan with validation gates and open research items.

---

## Plain-English explainer (blog)

For a jargon-free walkthrough of what this design is doing and why it's hard — explained entirely with a restaurant, a waiter, and a busser — see the companion blog posts (publishing shortly on shakacode.com/blog):

- **[Streaming a Web Page, Explained at a Restaurant Table](https://www.shakacode.com/blog/streaming-explained-restaurant-analogy/)** — this PR's problem in plain analogies.
- **[Teach Me With Analogies: The AI Prompt That Makes Hard Things Actually Click](https://www.shakacode.com/blog/teach-me-with-analogies-ai-prompt/)** — the reusable AI prompt used to produce that explainer.

### DIFF
--- a/docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md
+++ b/docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md
@@ -19,8 +19,8 @@ an adversarially-verified research pass (18 surviving primary-source claims; 7 c
 claims refuted) plus direct verification against shipped `react-dom` builds and the React
 repository. **Recommendation: a layered architecture** —
 
-1. **Foundation:** section chunks as individually addressable static assets behind a build
-   -time manifest; for strict-CSP deployments, switch chunk generation to React's external
+1. **Foundation:** section chunks as individually addressable static assets behind a
+   build-time manifest; for strict-CSP deployments, switch chunk generation to React's external
    Fizz runtime so cached chunks contain **zero inline scripts** and are byte-identical for
    every user.
 2. **Default transport (Scenario A):** Service-Worker stream-stitching — the SW proxies the
@@ -192,7 +192,8 @@ Verified directly against npm tarballs `react-dom@18.3.1`, `react-dom@19.2.7`,
 
 ### Edge platforms, Service Workers, and wire semantics — adversarially verified research
 
-An 18-claim verification pass (3 independent votes per claim; 2/3 refutes kill) against
+A verification pass over 25 extracted claims (3 independent votes per claim; 2/3 refutes
+kill; 18 confirmed, 7 refuted) against
 primary sources, fetched live 2026-07-21:
 
 13. **Cloudflare Durable Objects (C1) are documented and costed for this pattern.**
@@ -336,7 +337,9 @@ primary sources, fetched live 2026-07-21:
 Three workable strategies, in order of preference:
 
 1. **External runtime (Fact 12) — recommended for strict-CSP deployments.** Chunks carry
-   zero inline scripts; CSP is `script-src 'self'` (or a hash of the single runtime file).
+   zero inline scripts; CSP is `script-src 'self'` for the runtime file, optionally
+   hash-pinned via Subresource Integrity (`integrity=…` on the script tag — CSP
+   `script-src` hashes apply to inline scripts, not external files).
    Chunks become byte-identical for all users → maximal CDN cacheability, and Range
    resumption becomes possible. Cost: vendoring the runtime file + a contract test; the
    option is `unstable_`-prefixed.
@@ -377,8 +380,9 @@ time, alongside the chunk files:
 ```
 
 `concatOffset`/`bytes` enable the Range variant and SW byte-accounting;
-`boundaryId`/`contentId` remove filename-convention coupling; per-chunk `sha256` doubles as
-integrity check and CSP-hash input. Chunk 0 links the manifest via
+`boundaryId`/`contentId` remove filename-convention coupling; per-chunk `sha256` is a
+content-integrity check (and, under the hash-enumeration CSP strategy only, the input for
+computing the chunk's inline-script hashes). Chunk 0 links the manifest via
 `<link rel="preload" as="fetch">` (or inlines it) so the SW and page JS discover it without
 convention.
 
@@ -458,10 +462,22 @@ possible.
 - **Lifetime hardening (Fact 14):** `event.waitUntil()` on the stitch promise +
   StreamSaver-style `postMessage` heartbeat (~29 s) while the upstream is paced; the
   heartbeat stops once stitching completes.
-- **Truncation recovery:** if the SW is killed mid-stream anyway, the response truncates —
-  the page detects unresolved boundaries (Fact 4) and falls back to client pull of the
-  missing sections (C4 mechanism). Bytes remain ×1.0 because the killed stream stopped
-  sending. This same pull path serves browsers without SW support.
+- **Truncation recovery — with a byte cursor, because truncation can land mid-section:**
+  if the SW is killed mid-stream anyway, the response truncates, and the cut can fall
+  _inside_ a section file (sections are larger than one network chunk). The SW therefore
+  persists a per-page byte cursor (total forwarded bytes, mapped through the manifest's
+  `concatOffset`s to "last complete section + intra-section offset") as it forwards. The
+  page detects unresolved boundaries (Fact 4) and recovers: complete sections are fetched
+  whole; a partially-forwarded section is completed with a suffix request
+  (`Range: bytes=<intra-section-offset>-` against that section file, or the equivalent
+  offset into the concatenated asset). Bytes then remain exactly ×1.0. Where Range is
+  unavailable or the cursor was lost, recovery refetches the partial section from zero —
+  the guarantee degrades to "×1.0 for all sections except at most one repeated prefix",
+  and that degradation is measured, not hidden (P2). The same pull path serves browsers
+  without SW support. The skip path has the same mid-section geometry (an abort usually
+  lands inside a section) but no fragility: the SW is alive, its cursor is exact, and the
+  suffix fetch is precise. Only unexpected SW death risks a lost cursor — hence the
+  persistence requirement.
 - **First visit:** no controller → plain streamed page, no acceleration (requirement 5
   makes this acceptable by design). Optionally offer C4 as an opt-in first-visit
   accelerator where product prefers speed over the strict byte guarantee.
@@ -473,12 +489,29 @@ possible.
 Replace timing-window capture with a deterministic build: `prerender` (aborted at the
 shell) → store `prelude` (= chunk 0, cacheable) + `postponed` JSON. The tail is a
 `resume`/`resumeToPipeableStream` response generated on demand — and because the tail is
-now _a request_, the skip/priority signal is **stateless**: the client (or SW) simply
-issues `GET /resume/:page?first=<boundaryId>` and aborts the paced default. Transports
-from Layer 1 (or C1's edge join) carry the bytes. This is the same shape Next.js PPR
-ships, applied to a Rails + custom Node renderer stack, and it is the honest Scenario B
-answer: re-prioritization = which tail request you issue, not a reordering inside one
-stream.
+now _a request_, the skip/priority signal is **stateless**: the client (or SW) issues the
+tail request (e.g. `GET /resume/:page?priority=<boundaryId>`) and aborts the paced
+default. Transports from Layer 1 (or C1's edge join) carry the bytes.
+
+**Be precise about what the priority parameter can do.** React's `resume` API accepts no
+boundary selector or ordering option — within one resume stream, React completes
+boundaries in the order their data resolves. The `priority` parameter is therefore an
+**application contract, not a React one**, honored by one of two app-level mechanisms:
+
+- **Data-resolution ordering (preferred):** the endpoint resolves the target section's
+  data dependencies first (our `CacheSection` demo literally controls this — its async
+  delay is the data dependency), so React completes that boundary first. P6 explicitly
+  tests whether data-resolution order controls emission order in a resume stream.
+- **Per-section resume units:** produce independently resumable artifacts per section
+  (nested prerenders), so priority = which request is issued first. Coarser but
+  unambiguous; higher build complexity.
+
+These APIs are **recently stabilized** (new in the 19.2 line): treat them as a
+version-pinned contract — P6 is the standalone compatibility gate before any renderer
+integration, re-run on every React upgrade. This is the same shape Next.js PPR ships,
+applied to a Rails + custom Node renderer stack, and it is the honest Scenario B answer:
+re-prioritization = which tail request you issue plus data-resolution order inside it,
+not a protocol-level reordering inside one stream.
 
 ### Positioning of the other candidates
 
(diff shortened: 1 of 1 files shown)


justin808 (commented):
Overall, looks great. 

Questions:
1. how fragile and complex will the implementation be? 
2. how much of the solution is in core react code versus being invented

justin808:
@justin808 Both review questions are now answered in the doc itself (new sections under **Recommended architecture**, pushed in eed8c37e3):

**1. How fragile / complex?** → "[Fragility and complexity budget](https://github.com/shakacode/react_on_rails/blob/claude/scroll-priority-streaming-d76b49/docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md#fragility-and-complexity-budget)". Short version:

- Layer 0 (manifest): ~100 lines of rake extension; failure mode is build-time only.
- Layer 1 (SW): a few hundred lines against standard APIs; the **one** genuine fragility is the spec-acknowledged SW-lifetime gap, mitigated by a production-proven heartbeat — and every failure mode degrades to today's behavior (SW dies → truncated stream → pull recovery completes the page, still zero duplicate bytes; no SW → plain streaming). There is no failure state where the page breaks; worst case is "no acceleration".
- Layer 2 (resume): renderer endpoint plumbing, but it *deletes* invented machinery (timing-window capture, interruptible sleeps, marker files) — net reduction of bespoke surface. Gated behind a standalone spike (P6) before touching renderer code.
- Upgrade drift is made loud: contract tests against the installed react-dom fail CI if the wire format or vendored runtime moves.

**2. Core React vs invented?** → "[How much is core React vs invented?](https://github.com/shakacode/react_on_rails/blob/claude/scroll-priority-streaming-d76b49/docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md#how-much-is-core-react-vs-invented)" (provenance table). Short version: everything semantically hard is core React or web platform —

- Chunk wire format, boundary reveal, hydration retry: **React Fizz**, untouched (we only split bytes on boundaries React emits).
- Shell-now/tail-later: **stable React APIs** (`prerender`/`resume` — the same foundation Next.js PPR ships on).
- Strict-CSP chunks: **React's own** `unstable_externalRuntimeSrc` + React's runtime file (vendored, pinned).
- Transport: **web platform** (Service Worker + TransformStream + AbortController; the stitching pattern is shipped by Google's workbox-streams).
- Invented: the manifest (inert build-time JSON) and ~400–600 lines of SW/page glue React never sees — from React's perspective, stitched bytes are indistinguishable from origin-paced bytes. The only candidates that *call into* React internals (manual `$RC`) are exactly the ones the doc rejects.

Scheduling issues (master + children, parallelism marked, CDN work scoped to Cloudflare free tier) coming next — will link them here.

justin808:
Scheduling is up: master tracking issue #4778 with eight child issues (formally linked as GitHub sub-issues), each self-contained with its experiment and pass/fail criteria from the spec's prototype plan.

**Parallelism:** Wave 1 starts now with #4770 (manifest) ∥ #4771 (prerender/resume spike). Wave 2 runs #4772 (SW hardening) ∥ #4773 (CDN abort measurements — **Cloudflare free tier only**, per direction) ∥ #4774 (external runtime + strict CSP) once #4770 lands. Wave 3: #4775 (the skip feature itself) then #4776 (bfcache measurements). Wave 4: #4777 (productization go/no-go). Waves 1–3 are $0 budget.

justin808:
Bot-review triage complete (commits `accbb6847`, `1a39bae75`). All 13 inline comments assessed:

**Fixed — valid findings:**

| Finding | Raised by | Resolution |
| --- | --- | --- |
| Partial-section truncation retransmits the forwarded prefix, breaking ×1.0 | Greptile, Codex, CodeRabbit (converged) | SW now specifies a persisted per-page **byte cursor** + `Range: bytes=<offset>-` suffix recovery; honest degraded case (≤1 prefix repeat if cursor lost and Range unavailable) stated in Summary, Layer 1, fragility budget, and failure-mode table. P2 gains mid-section-kill and cursor-dropped arms; P3 notes aborts land mid-section and uses the cursor. |
| `resume` has no boundary-priority parameter — `?first=` overstated | Greptile, Codex (converged) | Layer 2 rewritten: the priority param is an **application contract**; honored via data-resolution ordering (preferred; `CacheSection`'s async delay is literally the data dependency) or per-section resume units. P6 now explicitly tests whether data-resolution order controls emission order. |
| `section_cache.rake`'s 4 KB time-window fragments can split Fizz units across files | Codex | P0 now requires re-splitting capture on Fizz unit boundaries (each tail file = exactly one hidden div + instruction) with well-formedness + concat-equality checks. |
| CSP `script-src` hashes don't apply to external files | Copilot (×2) | Corrected to `script-src 'self'` + optional SRI `integrity=` pinning, in both the CSP strategy section and P5. |
| "18-claim verification pass" ambiguous vs "18 surviving of 25" | claude-bot | Reworded: "25 extracted claims … 18 confirmed, 7 refuted". |
| "build/-time" markdown line-wrap artifact | Copilot, claude-bot | Rewrapped. |
| Layer 2 should not be treated as a settled stable contract | CodeRabbit | Layer 2 now marked **recently stabilized (19.2), version-pinned contract**, P6 as the compatibility gate re-run per React upgrade. (Kept the verified fact that the exports are on the stable channel — "experimental" would be factually wrong.) |

**Skipped:** none — every distinct finding resulted in a doc change.

Child issues #4770, #4771, #4772, #4775 get amendment comments reflecting the updated pass/fail criteria.

AbanoubGhadban:
## Independent Evaluation Findings (PR #4837, Issue #4835)

The independent evaluation in #4837 pressure-tested every claim in this design doc. Here are the key findings — what we confirmed, what we dispute, and what we discovered beyond the doc's scope.

### What the evaluation confirms

The design doc's analysis is thorough and accurate in most cells:

- **All 18 verified facts hold.** $RC idempotency, manual reveal via DOMParser+adoptNode, late arrival, boundary discovery, window.stop() hazards — all confirmed against react-dom@19.2.7.
- **The React internals cross-check is correct.** $RC batching rewrite ($RB/$RV/$RT), wire format stability (18.3.1→19.2.7), prerender/resume as stable exports, unstable_externalRuntimeSrc present but unstable-prefixed — all verified by reading the installed builds.
- **The comparison matrix is accurate** for every cell we could independently verify.
- **The layering is the right decomposition.** Layer 0 (manifest) → transport → resume is architecturally sound.

### What the evaluation disputes

**1. C2 (Service Worker) should be ruled out, not recommended as Layer 1.**

The doc acknowledges the SW lifetime risk but characterizes it as "a solvable engineering problem" with heartbeat mitigation. Our deep research (9 agents, 489K tokens examining SW stream control) found it is worse than presented:

- **Safari terminates idle SWs in 10 seconds** (WebKit `defaultTerminationDelay = 10_s`, changeset 291467) — not ~30s as the doc's framing implies. The heartbeat must fire every <10s, not ~29s.
- **`respondWith()` settles immediately** when the Response object is handed off. The SW becomes "idle" even though the body stream is still consumed. W3C issue #882's `waitForBody` proposal is still open after 10 years, assigned to "Version 2" milestone.
- **The doc says StreamSaver.js pings every ~29s.** Verified from source: it's actually **10s** (`mitm.html`).
- **First-visit gap is a hard constraint, not a graceful degradation.** 35-75% of traffic (first visits + cache eviction + hard refresh + private browsing) gets zero benefit. The doc frames this as "acceptable by design" under requirement 5, but if the majority of traffic is unaccelerated, the feature's value proposition weakens significantly.
- **Mid-section abort causes HTML parser state corruption.** The doc proposes byte-cursor recovery for mid-section truncation. Our analysis found this is fundamentally unsafe: an unclosed `<div hidden id="S:4">` traps subsequent sections inside it via the HTML parser's stack of open elements. `$RC("B:4","S:4")` then swallows the trapped content. Injecting `</div>` doesn't reliably close S:4 because the parser may have nested `<div>`, `<table>`, `<select>` scope boundaries inside. The only safe approach is clean-cut-only (abort between complete sections).
- **Byte-cursor Range resumption breaks under compression.** The SW counts decompressed bytes, but Range requests operate on compressed wire bytes. The offset spaces are incommensurable. Runtime-injected content (nonces, heartbeats, stream IDs) further invalidates static manifest offsets.

**2. C4 should be promoted from "rejected baseline" to recommended Layer 1.**

The doc rejects C4 because it violates requirement 1 (zero duplicate bytes). Our evaluation found:

- The duplicate is 21-35KB total for 7 sections (5KB/section). On HTTP/2 this is negligible.
- The duplicate is **harmless** — `$RC` idempotency is designed-in (React PRs #33511, #33531 explicitly hardened this).
- C4 passes both hard production constraints (first-visit + bot-safe) that C2 fails.
- C4 has zero blast radius, zero server changes, any-CDN compatibility, and graceful degradation on React upgrades (stream continues as fallback).
- The requirement for zero duplicate bytes, when combined with CDN-served + first-visit + bot-safe, creates a **fundamental impossibility** — no architecture satisfies all four. Relaxing "zero duplicates" (accepting ~25KB of harmless bytes) is the least costly trade-off.

**3. CDN abort behavior is unreliable (closes the doc's `?` on this).**

The doc correctly marks CDN abort behavior as unverified (P4). Our research closed it:

- Cloudflare Workers' `request.signal` has an **active regression** (workerd#6832) — abort events "never fire during SSE streaming."
- CloudFront and Fastly docs do **not** address viewer disconnect propagation to origin.
- This means the C2 abort-and-stitch approach cannot guarantee the CDN stops draining the origin after the client aborts — a silent bandwidth waste in Scenario B.

### What the evaluation adds beyond the doc

**1. The fundamental tension.** Zero duplicates + CDN-served + first-visit + bot-safe are mutually exclusive. This is the most important finding — it changes the decision framework from "which candidate is best?" to "which constraint are you willing to relax?"

**2. 30+ additional approaches evaluated.** WebSocket sideband, SSE, SharedWorker, Navigation API intercept, ReadableStream tee(), Speculation Rules, content-visibility:auto, HTTP/2 Server Push (removed from browsers), gRPC-Web, iframe sections, Turbo Frames, and more. All filtered against the two hard constraints. Full elimination table in the evaluation doc.

**3. Server-side section skipping** (Rank 3 in our evaluation). A new hybrid: client fetches section, confirms receipt via POST, server skips that section in the stream. Zero duplicates when origin holds the stream. Bot-safe via timeout fallback (send after 10s if no confirmation). ~40 lines delta from the existing prototype.

**4. SW mid-section abort parser analysis.** Deep HTML parser spec analysis proving that the byte-cursor approach is unsafe. The HTML parser's stack of open elements means an unclosed `<div hidden>` creates a scope that swallows subsequent content. The correct approach is clean-cut-only (section-boundary buffering).

### Coverage comparison

Every approach and avenue in this design doc is covered in the evaluation. The evaluation additionally covers the items listed above. There are **no gaps** — no approach in the doc was missed or under-investigated.

### The evaluation's recommendation

See #4837 for the full document. In brief:

| Layer | What | Effort | Status |
|-------|------|--------|--------|
| **1** | C4 fetch + $RC (accept 21-35KB dup) | 7-10 days | Already built |
| **2** | Skip-delay POST (complementary) | 2-3 days | Already built |
| **3** | Section skipping (zero dup, origin only) | 2.5 days | Not built |
| **4** | React PPR (long-term) | 21-31 days | Blocked on React |

C2 (SW) dropped. C3 (window.stop) dropped. C6 (server islands) dropped. C1 (DO) documented as Cloudflare-specific recipe only.

justin808:
@AbanoubGhadban — Justin asked if you could review the proposed close-only disposition for this PR. Current assessment: the architecture investigation has been superseded by merged #4841; the security preflight match is the literal React HTML-boundary example `<div hidden id="…S:">` in the design document, not executable code. Please approve or comment if you agree it should be closed without merge, or flag anything that should be retained first.

Issue #4815, opened 2026-07-29
agent_guardrails.rb: atomic-write copy_file, empty hooks arrays, per-edit interpreter cost


justin808:
## Summary

Three minor items were flagged alongside the `install_rsc_agent_guardrails` hook-schema bug in
#4753 (review of #4606) but are out of scope for that fix. Filing them separately so they aren't
lost when #4753 closes.

`react_on_rails/lib/react_on_rails/agent_guardrails.rb`

## Items

1. **`copy_file` doesn't use the atomic-write helper**: `register_hook` writes `.claude/settings.json`
   via the `atomic_write` helper (temp file + fsync + rename) so an interrupted write can never leave
   a half-written file behind. `copy_file` (used for both the skill markdown and the hook script)
   still uses plain `File.write`. A partial write there could leave a corrupt-but-executable
   `.claude/hooks/rsc-app-safety-check.rb` on disk. Consider routing `copy_file` through the same
   `atomic_write` helper for consistency.

2. **`add_hook` can leave empty `"hooks": []` arrays behind**: `add_hook` strips managed hooks from
   *every* `PostToolUse` entry regardless of matcher:
   ```ruby
   post_tool_use.each do |candidate|
     Array(candidate["hooks"]).reject! { |hook| managed_hook?(hook) }
   end
   ```
   If a managed hook was previously registered under a matcher entry that also had no other hooks,
   this can leave a `{ "matcher" => "...", "hooks" => [] }` entry behind after an upgrade. Cosmetic
   (doesn't break anything functionally) but worth cleaning up — e.g. drop entries whose `"hooks"`
   array is empty after stripping.

3. **Per-edit interpreter boot cost**: the hook is registered against the `Edit|Write` matcher, so a
   fresh `ruby` process is spawned on every edit — including JS, CSS, docs, and unrelated spec files —
   even though `rsc-app-safety-check.rb` exits early for paths that aren't Ruby routes/controllers.
   Worth deciding whether the per-edit interpreter-boot tax is acceptable as-is, or whether the
   matcher/script should narrow the fast-path check (e.g. a cheap extension check before requiring
   any Ruby stdlib) to reduce the common-case cost.

## Context

Surfaced during the fix for #4753 (`claude/4753-guardrail-hook-args`), which addressed the primary
bug (unsupported `"args"` array in the hook registration) and intentionally left these three items
unaddressed to keep that PR small and focused.

justin808:
[...]

PR #4816, opened 2026-07-29
Fix install_rsc_agent_guardrails PostToolUse hook to use supported schema (drop unsupported args array)

### LINKED ISSUE #4753: Add execution-level compatibility coverage for the generated Claude Code hook
> **Reframed 2026-07-29.** This issue originally reported that `install_rsc_agent_guardrails`
> registered a `PostToolUse` hook with an "unsupported `args` array" that would silently never run.
> **That premise was wrong** — `args` is a supported Claude Code hook field (exec form). PR #4816
> implemented the proposed fix and was closed unmerged because it would have introduced a regression.
> Full evidence: https://github.com/shakacode/react_on_rails/pull/4816#issuecomment-5118354590
>
> The valuable part of the original report was its last suggestion: unit specs cannot tell a working
> hook from a broken one. That gap is real and is what this issue now tracks.

## Why the original premise was wrong

Claude Code hooks have **two** execution forms, and `args` selects between them. From the installed
binary's own validation strings:

> `Command hooks require 'command'. For exec form (no shell), set 'command' to the executable and 'args' to its arguments: {"type": "command", "command": "echo", "args": ["hi"]}. For shell form, set 'command' to the full shell string: {"type": "command", "command": "echo hi"}.`

The published docs go further: **"Prefer exec form for any hook that references a path placeholder.
Exec form passes each `args` element as one argument with no shell tokenization, so paths with spaces
or special characters need no quoting."**

So the shipped registration is not just valid, it is the *recommended* shape:

```ruby
{ "type" => "command", "command" => "ruby", "args" => ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"] }
```

The corroborating evidence in the original report — "no hook anywhere in the repo uses an `args`
array" — is true but proves only that this repo consistently uses shell form, not that exec form is
unsupported. Ironically, the repo's own shell-form hooks use **unquoted** placeholders, which is the
shape the docs advise against.

## What to actually do

**1. Keep the current `command` + `args` implementation.** Do not convert to shell form. Converting
would break on project paths containing spaces — Greptile flagged exactly that as a P1 on #4816.

**2. Add an execution-level smoke test for the generated application.** This is the real gap. The
existing specs in `react_on_rails/spec/react_on_rails/agent_guardrails_spec.rb` assert the *shape* of
the emitted JSON, which cannot distinguish a hook that runs from one that does not — the defect this
issue was filed about would have been invisible to them either way, in both directions.

The test should generate an application, run the installer, and prove the installed hook **actually
executes**, asserting on an observable side effect rather than on config shape.

**Include a project path containing a space.** That is the case that separates exec form from shell
form, it is the regression #4816 would have introduced, and no current test covers it.

**3. Document the minimum supported Claude Code version** if clients older than May 2026 matter.
I could not determine when exec form was introduced — see UNKNOWN below. If `args` postdates some
supported client version, the installer needs either a documented floor or a fallback, and the
original report may have been accurate when it was filed.

## Verified

- Exec form and the `args` field are supported in the Claude Code build available locally, confirmed
  twice against the binary's own validation strings by two independent agents.
- `react_on_rails/lib/react_on_rails/agent_guardrails.rb` on `main` correctly retains exec form.
- This repo has never had the guardrail installed: no `.claude/hooks/rsc-app-safety-check.rb`, no
  matching `PostToolUse` entry, and no `args` array anywhere in `.claude/settings.json`. So the
  dogfooding path never exercised it either.

## UNKNOWN

- **When exec form was introduced.** Support is verified in a current build; whether `args` was
  supported when #4606 merged is not established. If it was not, the original report was accurate at
  the time and this became stale rather than wrong.
- **Whether the installed hook has ever executed end to end in a generated app.** Nobody has run it.
  That is precisely what item 2 exists to establish, and until it is done, both "it works" and "it
  never runs" remain unproven by test.

## Suggested labels

`P2`, testing/documentation gap rather than a bug.

Original report and its evidence remain in the history below. Related: #4815 (minor items from the
same review, premise corrected), #4816 (closed unmerged), #4606 (source PR).

justin808:
## Summary

`react_on_rails:install_rsc_agent_guardrails` (shipped in #4606) registered its `PostToolUse`
hook using a `"command"` + `"args"` array shape that is not part of Claude Code's hook schema.
Claude Code hook entries take a single `command` string, so the installed RSC advisory guardrail
almost certainly never ran in a host app. This repo's own `.claude/settings.json` is a working
reference: every hook entry it registers (`main-ci-status.sh`, `main-ci-status-on-push.sh`,
`bin/claude-hooks/autofix-file`, `rsc-guardrails-check.sh`) uses a single `command` string with
no `args` key.

Fixes #4753

Labels: `bug` — a shipped-but-unreleased generator (`install_rsc_agent_guardrails`) produced a
hook registration that the Claude Code hook runner silently no-ops, so the advisory guardrail it
installs never actually executes.

## Before / after

Before (buggy — `args` silently ignored, effective command is bare `ruby` with no script):

```json
{ "type": "command", "command": "ruby", "args": ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"] }
```

After (matches the repo's own supported schema):

```json
{ "type": "command", "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb" }
```

The template already starts with `#!/usr/bin/env ruby` and is chmod'd `0o755` by `copy_file`, so
it's directly invocable without a separate `ruby` interpreter argument.

## What changed

`react_on_rails/lib/react_on_rails/agent_guardrails.rb`:
- `HOOK_COMMAND` now holds the full executable path (`${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb`) instead of the bare `"ruby"` interpreter name.
- `HOOK_ARGS` is removed entirely.
- `add_hook` emits `{ "type" => "command", "command" => HOOK_COMMAND }` — no `args` key.
- `registered_hook?` compares only `type`/`command`, no longer references `args`.

`react_on_rails/spec/react_on_rails/agent_guardrails_spec.rb`:
- The `rsc_hooks` test helper and all specs that previously pinned the buggy `args` shape (the
  install/idempotency spec, the legacy-upgrade spec, and the duplicate-managed-hook spec) now
  assert the single-`command`-string shape instead.
- Added a new spec, "registers the project hook as a single command string, matching Claude
  Code's supported PostToolUse schema (no unsupported `args` array)", asserting the emitted hook
  entry `eq`s a plain two-key hash (`type`, `command`) with no `args` key.

## RED -> GREEN

RED (new spec run against the unfixed code):

```
1) ReactOnRails::AgentGuardrails registers the project hook as a single command string, matching Claude Code's supported PostToolUse schema (no unsupported "args" array)
   Failure/Error: expect(hook).to eq("type" => "command", "command" => described_class::HOOK_COMMAND)

     expected: {"command" => "ruby", "type" => "command"}
          got: {"args" => ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"], "command" => "ruby", "type" => "command"}

     Diff:
     @@ -1 +1,2 @@
     +"args" => ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"],
      "command" => "ruby",

1 example, 1 failure
```

GREEN (full suite after the fix, including the updated pinned specs):

```
Finished in 9.42 seconds (files took 1.53 seconds to load)
53 examples, 0 failures
```

`bundle exec rubocop` on both changed files: `2 files inspected, no offenses detected`.

## Dogfooding check

Per the issue, I checked this repo's own `.claude/settings.json` and
`.claude/hooks/rsc-app-safety-check.rb` for the same defect. The guardrail has never been
installed into this repo (no `rsc-app-safety-check.rb` file exists under `.claude/hooks/`, and
`.claude/settings.json`'s `PostToolUse` entry only contains `bin/claude-hooks/autofix-file` and
`rsc-guardrails-check.sh` — no `rsc-app-safety-check` entry at all), so there is nothing to fix
there. Both files are consequently untouched in this PR.

## Changelog decision

No `CHANGELOG.md` entry added. `install_rsc_agent_guardrails` (#4606) is itself still under
`[Unreleased] > Added` and has never shipped in a tagged release — this PR fixes the same
unreleased feature before it ships, so end users will only ever see the corrected behavior. A
"Fixed" entry describing a bug nobody using a released version experienced would be internal
noise rather than user-facing information. Happy to add one if the coordinator disagrees.

## Out-of-scope follow-up

The same review thread flagged three additional minor items in this file (`copy_file` not using
the `atomic_write` helper, `add_hook` potentially leaving empty `"hooks": []` arrays behind, and
per-edit Ruby interpreter boot cost on the `Edit|Write` matcher). Filed as
https://github.com/shakacode/react_on_rails/issues/4815 to keep this PR focused on the schema
bug and avoid losing them once #4753 closes.

## Codex Decision Log

- **HOOK_COMMAND value**: chose the full `${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb`
  path over keeping `"ruby"` plus a rewritten single-string command, since the template is already
  executable and this matches every other hook in this repo's own `.claude/settings.json`
  (non-blocking, low risk — matches issue's proposed fix verbatim).
- **No backward-compat migration for pre-existing broken installs**: `registered_hook?`/`managed_hook?`
  do not special-case the old `command == "ruby"` shape. Confirmed via CHANGELOG.md that
  `install_rsc_agent_guardrails` (#4606) is still under `[Unreleased]` in this repo (never in a
  tagged release), so no host app can have the old broken entry on disk yet. Non-blocking.
- **Changelog**: no entry added; see "Changelog decision" above. Non-blocking, but flagging since
  the lane brief called it an explicit judgment call.

## Confidence note

Confidence note:
- Validated: `bundle exec rspec spec/react_on_rails/agent_guardrails_spec.rb` (RED before fix: 1
  failure with the exact args-array mismatch shown above; GREEN after fix: 53 examples, 0
  failures). `bundle exec rubocop lib/react_on_rails/agent_guardrails.rb
  spec/react_on_rails/agent_guardrails_spec.rb` (0 offenses). Confirmed this repo's own
  `.claude/settings.json` (4 hook entries) and searched the repo for any other
  `HOOK_ARGS`/`HOOK_COMMAND` references — none outside the two owned files.
- Evidence: RED/GREEN rspec output and rubocop output above; issue #4753 body (includes the
  `#4606` PR review thread link corroborating the schema mismatch).
- UNKNOWN: none — the fix and its correctness are directly verifiable from this repo's own
  working hook registrations without needing external Claude Code schema docs.
- Residual risk: none identified. The change is narrowly scoped to the hook-registration shape;
  all other guardrail behavior (skill install, legacy `.sh` hook removal/upgrade, settings.json
  merge/atomic-write safety, the hook script's own detection logic) is untouched and still
  covered by the existing 53 specs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Claude Code hook registration to use the supported single-command format.
  * Improved hook detection during installation and upgrades, preventing duplicate managed hooks.
  * Ensured legacy hook entries are correctly recognized and removed.

### DIFF
--- a/react_on_rails/lib/react_on_rails/agent_guardrails.rb
+++ b/react_on_rails/lib/react_on_rails/agent_guardrails.rb
@@ -21,8 +21,11 @@ module ReactOnRails
       "rsc_app_safety_check.rb" => ".claude/hooks/rsc-app-safety-check.rb"
     }.freeze
 
-    HOOK_COMMAND = "ruby"
-    HOOK_ARGS = ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"].freeze
+    # Claude Code's PostToolUse hook entries take a single "command" string (see this repo's own
+    # working .claude/settings.json for a live reference). There is no "args" array in that schema;
+    # the template is executable (chmod 0o755) and starts with `#!/usr/bin/env ruby`, so it can be
+    # invoked directly without a separate interpreter argument.
+    HOOK_COMMAND = "${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"
     LEGACY_HOOK_COMMAND = "${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.sh"
     LEGACY_HOOK_REL = ".claude/hooks/rsc-app-safety-check.sh"
     HOOK_REL = ".claude/hooks/rsc-app-safety-check.rb"
@@ -208,11 +211,11 @@ module ReactOnRails
           entry = { "matcher" => HOOK_MATCHER, "hooks" => [] }
           post_tool_use << entry
         end
-        (entry["hooks"] ||= []) << { "type" => "command", "command" => HOOK_COMMAND, "args" => HOOK_ARGS }
+        (entry["hooks"] ||= []) << { "type" => "command", "command" => HOOK_COMMAND }
       end
 
       def registered_hook?(hook)
-        hook["type"] == "command" && hook["command"] == HOOK_COMMAND && hook["args"] == HOOK_ARGS
+        hook["type"] == "command" && hook["command"] == HOOK_COMMAND
       end
 
       def managed_hook?(hook)
--- a/react_on_rails/spec/react_on_rails/agent_guardrails_spec.rb
+++ b/react_on_rails/spec/react_on_rails/agent_guardrails_spec.rb
@@ -26,7 +26,7 @@ module ReactOnRails
     def rsc_hooks
       settings.dig("hooks", "PostToolUse").flat_map { |entry| entry["hooks"] }
               .select do |hook|
-        hook["args"] == described_class::HOOK_ARGS || hook["command"] == described_class::LEGACY_HOOK_COMMAND
+        hook["command"] == described_class::HOOK_COMMAND || hook["command"] == described_class::LEGACY_HOOK_COMMAND
       end
     end
 
@@ -657,15 +657,17 @@ module ReactOnRails
       expect(skill).not_to include("Props passed to SSR can reach your error tracker")
     end
 
-    it "registers the project hook in exec form so paths with spaces are not shell-split" do
+    it "registers the project hook as a single command string, matching Claude Code's supported " \
+       "PostToolUse schema (no unsupported \"args\" array)" do
       described_class.install(@app_root)
 
       hook = settings.dig("hooks", "PostToolUse").flat_map { |entry| entry.fetch("hooks") }
-                     .find { |entry| entry["command"] == described_class::HOOK_COMMAND }
+                     .find { |entry| entry["type"] == "command" && entry["command"] == described_class::HOOK_COMMAND }
 
-      expect(hook).to include(
-        "type" => "command", "command" => described_class::HOOK_COMMAND, "args" => described_class::HOOK_ARGS
-      )
+      # Claude Code's PostToolUse hook entries take a single "command" string (see this repo's own
+      # working .claude/settings.json). An "args" array is not part of that schema and is silently
+      # ignored by the hook runner, so a hook shaped that way never actually invokes the script.
+      expect(hook).to eq("type" => "command", "command" => described_class::HOOK_COMMAND)
     end
 
     it "removes the legacy shell hook file and upgrades its settings registration" do
@@ -697,7 +699,7 @@ module ReactOnRails
         "removed    .claude/hooks/rsc-app-safety-check.sh (replaced by .claude/hooks/rsc-app-safety-check.rb)"
       )
       expect(project_hooks).to contain_exactly(
-        "type" => "command", "command" => described_class::HOOK_COMMAND, "args" => described_class::HOOK_ARGS
+        "type" => "command", "command" => described_class::HOOK_COMMAND
       )
     end
 
@@ -712,11 +714,7 @@ module ReactOnRails
               {
                 "matcher" => "Edit|Write",
                 "hooks" => [
-                  {
-                    "type" => "command",
-                    "command" => described_class::HOOK_COMMAND,
-                    "args" => described_class::HOOK_ARGS
-                  },
+                  { "type" => "command", "command" => described_class::HOOK_COMMAND },
                   { "type" => "command", "command" => described_class::LEGACY_HOOK_COMMAND }
                 ]
               }
@@ -728,7 +726,7 @@ module ReactOnRails
       described_class.install(@app_root)
 
       expect(rsc_hooks).to contain_exactly(
-        "type" => "command", "command" => described_class::HOOK_COMMAND, "args" => described_class::HOOK_ARGS
+        "type" => "command", "command" => described_class::HOOK_COMMAND
       )
     end
 


justin808:
## Blocking finding: this PR's premise does not hold — recommending we close it unmerged

Coordinator for batch `ror-a-17-1-wave-a-20260729`. I verified the underlying claim before merging and it does not survive contact with the evidence. **`args` is a supported Claude Code hook field.** The original code was valid, and this PR replaces it with a form that is measurably worse.

### Evidence

Claude Code ships two hook execution forms. From the installed binary (`claude-code/2.1.219`), its own validation string:

> `Command hooks require `command`. For exec form (no shell), set `command` to the executable and `args` to its arguments: {"type": "command", "command": "echo", "args": ["hi"]}. For shell form, set `command` to the full shell string: {"type": "command", "command": "echo hi"}.`

And a second validation string confirming `args` is schema-aware, not ignored:

> `has both "args" and whitespace in "command". Exec form treats "command" as a single executable name; move the rest into "args". Example: { "command": "node", "args": ["script.js"] }.`

The published documentation agrees, and goes further:

> "Exec form (when `args` is present): Claude Code resolves `command` as an executable on PATH and spawns it directly with `args` as the argument vector. There is no shell."
> "**Prefer exec form for any hook that references a path placeholder.** Exec form passes each `args` element as one argument with no shell tokenization, so paths with spaces or special characters need no quoting."

### What that means for this change

The shape this PR removes —

```ruby
{ "type" => "command", "command" => "ruby", "args" => ["${CLAUDE_PROJECT_DIR}/.claude/hooks/rsc-app-safety-check.rb"] }
```

— is structurally identical to the documented exec-form example. It resolves `ruby` on PATH and passes the script path as a single argument. It is valid, it runs, and it is space-safe.

The shape this PR introduces is shell form with an **unquoted** placeholder. That is the form the docs explicitly advise against for path placeholders, and it breaks outright when the project path contains a space. Greptile independently flagged exactly this as a P1 on the new line (https://github.com/shakacode/react_on_rails/pull/4816#discussion_r3674574221) — that finding is correct about the new code, and it is a defect this PR would introduce rather than one it fixes.

The spec changes have the same problem: the new assertion that the emitted entry has no `args` key pins the *less* correct shape.

### Why the original review missed it

The corroborating evidence in #4753 was that no hook in this repo uses `args`. That is true, but it establishes only that we happen to use shell form everywhere — not that exec form is unsupported. Both forms are valid; ours is the one the docs advise against for path placeholders.

### Honest limits of this finding

- I verified against Claude Code **2.1.219**. I did not establish when exec form was introduced, so I cannot rule out that `args` was unsupported when #4606 merged and the report was accurate at the time.
- If an app runs a Claude Code old enough to lack exec form, the original registration would silently no-op there. That is a real scenario, but the remedy would be shell form *with quoted placeholders*, which is not what this PR does.
- I did not execute the installed hook end to end in a generated app. The issue's own "very likely never runs" was a hypothesis; mine is a schema-level refutation, not an execution trace. #4753's suggestion of an end-to-end check that the generated `settings.json` hook actually executes remains the right way to settle it conclusively, and would be worth doing regardless of which form we land on.

### Recommendation

Close this PR unmerged and leave `agent_guardrails.rb` on exec form. I am leaving #4753 open for a maintainer decision rather than closing someone else's issue on my own analysis — but as written, its central claim that `args` is "not part of Claude Code's hook schema" is not correct, and the "Proposed fix" section would be a regression.

The three minor items split out to #4815 (atomic_write, empty `hooks: []`, per-edit interpreter cost) are unaffected by this and still stand on their own.

justin808:
Closing unmerged: the premise does not hold — `args` is supported exec form, and this change would introduce a path-with-spaces regression. Evidence in the comment above; #4753 left open for a maintainer decision.

PR #4837, opened 2026-08-02
Scroll-priority streaming: candidate architecture evaluation (#4835)

### LINKED ISSUE #4835: Evaluate scroll-priority streaming candidate architectures (feasibility, pros/cons) from #4769
## Summary

Evaluate the **candidate architectures proposed in the scroll-priority streaming design doc** (#4769) — feasibility, pros, and cons of each — and produce a recommendation on which (if any) React on Rails should pursue.

The design doc in #4769 already lays out six candidates and an author's recommendation. This issue is the **independent evaluation pass**: pressure-test the analysis, close the open questions the matrix marks as unverified, and decide what we actually build.

Parent: #4385 (selective/independent hydration for traditional, non-RSC components)
Design doc: #4769 — `docs/superpowers/specs/2026-07-21-scroll-priority-streaming-design.md`
Prototype that raised the question: #4740

## The problem the candidates address

When a page is streamed with Suspense boundaries and the chunks are CDN-cached, and the user scrolls toward a section that has not arrived yet: deliver that section ASAP with **zero duplicate bytes**, with selective hydration and already-mounted component state intact, without degrading the normal streamed path.

## Candidates to evaluate

| ID | Approach | One-line mechanism |
| -- | -------- | ------------------ |
| C1 | Edge-held state (Cloudflare Durable Object) | Navigation + skip signal routed to the same DO; DO races its delay loop against the skip and flushes remaining chunks down the same connection |
| C2 | Service Worker stream-stitching | SW proxies the navigation through a `TransformStream`; on scroll intent aborts upstream, fetches only unseen section files, writes them into the same stream |
| C3 | Gated `window.stop()` + client pull | Stop the navigation stream after critical resources, then fetch remaining sections and reveal via manual `$RC` / external-runtime append |
| C4 | No-stop client pull | Stream keeps flowing, client fetches ahead, idempotent `$RC` makes the race safe (duplicates bytes — kept as baseline, as C2's recovery path, and as an optional first-visit accelerator) |
| C5 | Official `prerender`/`resume` split | Chunk 0 as a cacheable `prerender` prelude, tail generated on demand from stored `postponed` state; priority becomes a stateless request parameter |
| C6 | Pull-only tail ("server islands" mode) | Chunk 0 ends the navigation response; a loader fetches section files in viewport-priority order and reveals them (Astro Server Islands pattern + React hydration) |

Also in scope, as evaluated avenues rather than full candidates: HTTP Range resumption against a pre-concatenated static asset, HTTP/2+ multiplexing vs. the single-stream aesthetic, and priority signals without cancellation (`fetchpriority`, RFC 9218 `PRIORITY_UPDATE`, `103 Early Hints`).

## Scope of the investigation

- [ ] **Re-derive the comparison independently.** Confirm or dispute each cell of the doc's matrix (bytes ×1.0, scroll→interactive, CDN portability, hydration + state, strict CSP, Scenario B story, complexity/ops, reliance on undocumented internals, SEO/no-JS). Flag anything asserted rather than verified.
- [ ] **Close the open risks the matrix marks `?`:**
  - [ ] C1 — Durable Object cross-colo placement/forwarding latency (documented qualitatively, never quantified)
  - [ ] C2 — Chromium/Safari mid-stream Service Worker policy; the Firefox 30 s + 30 s budget and whether the heartbeat mitigation holds
  - [ ] C3/C4 — bfcache behavior and `readyState` stuck at `interactive`
  - [ ] C5 — whether `prerender`/`resume` composes on demand **outside Next.js**, across separate process invocations
- [ ] **Assess fit with React on Rails specifically**, not just in the abstract: the streaming render path (`streamServerRenderedReactComponent.ts`), the node renderer, Pro's streaming/RSC code paths, existing hydration/store-hydration flow, and what each candidate would require of an app's deployment (CDN, edge runtime, SW registration, asset pipeline).
- [ ] **Evaluate the layered recommendation as such** — manifest + byte-stable sections → C2 as Scenario A transport → C5 as the Scenario B pipeline, with C1 as a Cloudflare-native alternative and C3 rejected. Is the layering the right decomposition, or does one candidate dominate?
- [ ] **Cost the work.** For each viable candidate: implementation effort, ongoing maintenance/fragility budget, blast radius on the normal (non-scroll-priority) path, and what breaks on a React upgrade.
- [ ] **Cross-check the React-internals claims** against currently shipped `react-dom` — notably the React 19.2 `$RC`/`completeBoundary` rewrite (`$RB`/`$RV`, ~300 ms reveal batching), the stability of the `prerender`/`resume` family, and `unstable_externalRuntimeSrc` as the strict-CSP answer.
- [ ] **Sanity-check the prior art** the doc leans on (Astro Server Islands, Next.js PPR, workbox-streams, Marko/SolidStart, Turbo Frames / htmx lazy fragments) — the doc flags several of these as fetched but not adversarially verified.

## Deliverable

A written recommendation on this issue (or as a doc PR) that states:

1. Which candidates are **viable**, which are **conditionally viable** (and on what condition), and which should be **ruled out** — with the reasoning, not just the verdict.
2. Whether we adopt the doc's layered recommendation, a different composition, or **nothing yet** (i.e. the payoff does not justify the complexity at this point).
3. If we proceed: which prototype steps from the doc's P0–P8 plan are the first slice, and what their pass/fail criteria are.
4. If we do not proceed: what would have to change (React API, CDN capability, measured demand) for this to become worth revisiting.

Implementation is explicitly **out of scope** here — this issue ends at a decision. Any build work gets its own follow-up issue.

AbanoubGhadban:
## Summary

Independent evaluation of the six candidate architectures from the scroll-priority streaming design doc (#4769), plus 30+ additional approaches discovered during research. Addresses issue #4835.

Produced via 4 deep-research workflows (24 agents, ~1.5M subagent tokens, 631 tool invocations) with adversarial verification against primary sources.

**This PR adds the evaluation document only.** No code changes. Implementation is explicitly out of scope per #4835.

## The Fundamental Tension

A key finding: **zero duplicates + CDN-served + first-visit + bot-safe are mutually exclusive.** No architecture satisfies all four. The evaluation makes this tension explicit and ranks approaches by which constraint they relax.

## Candidate Verdicts

| Candidate | Verdict | Constraint violated |
|-----------|---------|---------------------|
| **C1** — DO edge-state | Conditionally viable | Adds hosting cost (~$47-94/mo per 1M views) |
| **C2** — SW stream-stitching | **Ruled out** | Fails first visit (SW doesn't exist) |
| **C3** — `window.stop()` + pull | **Ruled out** | Fails bot safety (kills all network) |
| **C4** — No-stop client pull | **Viable — recommended** | Accepts 21-35KB duplicate bytes |
| **C5** — `prerender`/`resume` | Conditionally viable — defer | Blocked on `unstable_postpone()` |
| **C6** — Pull-only tail | **Ruled out** | Fails bot safety (no JS = skeletons) |

## Key Findings Beyond the Design Doc

1. **C2 (SW) fails the first-visit constraint** — the navigation response goes directly from network to HTML parser; the SW cannot retroactively intercept it, even with `clients.claim()`. 35-75% of traffic is uncontrolled.

2. **SW mid-section abort causes HTML parser state corruption** — an unclosed `<div hidden id="S:4">` traps subsequent sections inside it. `$RC("B:4","S:4")` then swallows the trapped sections. Injecting `</div>` doesn't work because the SW can't know the parser's stack of open elements.

3. **CDN abort propagation is unreliable** — Cloudflare Workers' `request.signal` has an active regression (workerd#6832). CloudFront and Fastly docs don't address viewer disconnect propagation.

4. **New approach: Server-side section skipping** (Rank 3) — client fetches + confirms receipt; server skips that section in the stream. Zero duplicates when origin holds the stream. Bot-safe via timeout fallback. ~40 lines delta from existing prototype.

## Recommended Layered Architecture

| Layer | Approach | When | Effort |
|-------|----------|------|--------|
| **1** | C4 fetch + `$RC` (accept duplicates) | Ship now | 7-10 days |
| **2** | Skip-delay POST (complementary) | Ship in parallel | 2-3 days |
| **3** | Server-side section skipping (zero dup, origin only) | When origin streams | 2.5 days |
| **4** | React PPR | When `postpone()` stable (2027+) | 21-31 days |

## All 10 Open Questions Closed

See section 6 of the document — every `?` cell from the design doc matrix now has a primary-source answer.

## Files Changed

- `internal/analysis/scroll-priority-candidate-evaluation-2026-08-02.md` — the evaluation document

Closes #4835

## Summary by CodeRabbit

- **New Features**
  - Added a selective hydration demo with configurable content sections, delays, repeated content, topics, and interactive controls.
  - Added scroll-aware loading that prioritizes visible sections through streaming or on-demand delivery.
  - Added status and timing feedback for section arrivals, loading progress, errors, duplicate deliveries, and completion.
  - Added controls to release delayed content when requested.
- **Documentation**
  - Added architecture evaluations and implementation guidance for scroll-priority streaming and cached progressive rendering.

### DIFF
--- a/internal/analysis/scroll-priority-candidate-evaluation-2026-08-02.md
+++ b/internal/analysis/scroll-priority-candidate-evaluation-2026-08-02.md
@@ -1,588 +1,432 @@
 # Scroll-Priority Streaming Architecture Evaluation
 
 **Issue:** #4835 (evaluation) → parent #4385, design doc #4769
-**Date:** 2026-08-02
-**Method:** Independent re-derivation with adversarial multi-agent verification (9 agents, 594K tokens, 280 tool invocations)
+**Date:** 2026-08-02 (updated)
+**Method:** Independent re-derivation with adversarial multi-agent verification (24 agents across 4 workflows, ~1.5M subagent tokens, 631 tool invocations)
 **Branch:** selective-hydration-scroll-priority-demo
 
 ---
 
 ## 1. Executive Summary
 
-Six candidate architectures were evaluated for delivering scroll-priority streaming in React on Rails. The evaluation drew on primary source verification of browser specifications, React internals, CDN platform documentation, and integration analysis against the existing `streamServerRenderedReactComponent.ts` pipeline.
+Six candidate architectures from the design doc (#4769) were evaluated, plus 30+ additional approaches discovered during research. Each was tested against the evaluation criteria from the issue and two hard production constraints:
 
-| Candidate                                    | Verdict                                      | Rationale                                                                                                                                                                                                                                                                                                                                                                                                                       |
-| -------------------------------------------- | -------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **C1** Edge-Held State (Cloudflare DO)       | **Conditionally viable**                     | Technically sound but imposes hard vendor lock-in to Cloudflare. No equivalent exists on any other edge platform (verified against Fastly, AWS, Akamai, Deno Deploy, Vercel). Cost is acceptable (~$47-94/mo per 1M views).                                                                                                                                                                                                     |
-| **C2** Service Worker Stream-Stitching       | **Ruled out**                                | Safari terminates idle SWs in 10 seconds (verified: WebKit `defaultTerminationDelay = 10_s`). The W3C `waitForBody` proposal (issue #882) is still open after 10 years, assigned to "Version 2" milestone. 35-75% of traffic is uncontrolled (first visit + eviction). workbox-streams provides no keepalive, no priority reordering, no abort mid-stream.                                                                      |
-| **C3** Gated window.stop() + Client Pull     | **Ruled out**                                | `window.stop()` kills ALL in-flight network activity. `document.readyState` behavior after stop is historically underspecified and browser-inconsistent. Breaks Google Tag Manager, analytics, Selenium, lazy-loading libraries. Incompatible with RSC payload injection. Direct dependency on `$RC` internal API with 3 major rewrites in 2024-2026.                                                                           |
-| **C4** No-Stop Client Pull (fetch delivery)  | **Viable -- recommended as Layer 1**         | Already implemented and working in the demo (`?mode=fetch`). Zero blast radius on normal path. Zero new dependencies. `$RC` idempotency makes the stream+fetch race safe. Graceful degradation: if fetch path breaks on React upgrade, stream delivers sections normally.                                                                                                                                                       |
-| **C5** Official prerender/resume (React PPR) | **Conditionally viable -- defer to Layer 3** | Uses stable React 19.2 APIs (`prerenderToNodeStream`, `resumeToPipeableStream` -- confirmed in `react-dom@19.2.7`). But requires deep changes to `streamServerRenderedReactComponent.ts`, `injectRSCPayload.ts`, the node renderer protocol, and a PostponedState storage layer. `React.postpone()` is still `unstable_postpone()` (not in stable npm). High integration effort, high reward if React PPR becomes the standard. |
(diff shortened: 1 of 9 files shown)


AbanoubGhadban:
Replacing with a clean PR that contains only the evaluation document (no dummy app changes).

PR #4875, opened 2026-08-12
[Pro] Protocol cleanup: drop targetBundles and wire request digest (#4873)

### LINKED ISSUE #4873: [Pro] Protocol cleanup for next major: drop targetBundles back-compat and the wire request digest
## Problem

Two wire-protocol legacies persist only for backward compatibility and should be removed together at the next breaking (protocol-major) version:

1. **`targetBundles` on asset upload** — kept solely for protocol 2.0.0 renderers. `react_on_rails_pro/lib/react_on_rails_pro/request.rb` (`upload_assets`):

   > `targetBundles` is only kept for backward compatibility with older node renderers (protocol 2.0.0) that require it. The new node renderer derives target directories from the `bundle_<hash>` form keys and ignores this field. Remove at the next breaking version.

2. **The `:renderRequestDigest` URL segment** — the renderer never reads its value. `packages/react-on-rails-pro-node-renderer/src/worker.ts` (render route registration):

   > the digest is part of the request URL. Yes, it's not used here, but the server logs might show it to distinguish different requests.

   Both the `/bundles/:bundleTimestamp/render/:renderRequestDigest` and `/bundles/:bundleTimestamp/incremental-render/:renderRequestDigest` routes carry it. On the Rails side the digest is an MD5 over the full `js_code` (including serialized props) computed for every render even when prerender caching is off (`server_rendering_pool/pro_rendering.rb`, `set_request_digest_on_render_options`; TODOs at `pro_rendering.rb` and `node_rendering_pool.rb` referencing archived `react_on_rails_pro#119`).

## Proposal

At the next protocol-major version:

- Remove `targetBundles` from the asset-upload form (`request.rb#upload_assets`).
- Make the render and incremental-render routes digest-free (or make the segment optional), and update the gem-side path builders (`node_rendering_pool.rb#prepare_render_path` / `#prepare_incremental_render_path`).
- Bump `protocolVersion` in the node-renderer package and the gem's expected protocol version together.

## Sequencing

- Blocked by making the request digest cache-only on the Rails side (separate planned perf issue: skip the MD5 on the non-cached render path; not yet filed). That change is non-breaking and lands first; this issue then removes the wire artifact.
- Target: 18.0 breaking batch.

## Validation

- Protocol-version matrix test in the renderer suite (old gem ↔ new renderer rejection paths).
- Gem/renderer contract specs covering upload and render paths without the removed fields.

_Evidence gathered 2026-08-11 from a full Pro-surface scan; file/line references are as of `main` at that date._

AbanoubGhadban:
## What Changed

**BREAKING** (protocol-major): Bumps the gem↔node-renderer wire protocol from 2.0.0 to 3.0.0 by removing two legacy artifacts:

### 1. `targetBundles` removed from asset upload and asset-exists

- **Gem**: `upload_assets` no longer sends `form["targetBundles"]`. The node renderer already derived target directories from `bundle_<hash>` form keys and ignored this field.
- **`/asset-exists`**: Migrated from `targetBundles` to `bundle_<hash>` keys on both the gem and node sides. The gem now sends `bundle_<hash>: true` keys, and the node endpoint extracts bundle hashes from those keys (matching the convention already used by `/upload-assets`).
- **Single-flight key**: Updated `upload_assets_single_flight_key` parameter name from `target_bundles` to `bundle_hashes`.

### 2. `:renderRequestDigest` removed from render URLs

- **Node renderer**: Render route changed from `/bundles/:bundleTimestamp/render/:renderRequestDigest` to `/bundles/:bundleTimestamp/render`. Same for incremental-render.
- **Gem**: `build_render_path` no longer appends `render_options.request_digest` to the URL.
- **MD5 digest now cache-only**: `set_request_digest_on_render_options` is only called when prerender caching is enabled. Previously it computed an MD5 of the full JS code on every render, even when caching was off.

### 3. Protocol version bump

- Gem: `PROTOCOL_VERSION = "3.0.0"` in `version.rb`
- Node: `"protocolVersion": "3.0.0"` in `package.json`
- Existing `checkProtocolVersion` handler returns HTTP 412 on mismatch (no changes needed)

### 4. Cleanup

- Removed all TODO comments referencing `react_on_rails_pro/issues/119`

## How to Verify

- Gem and node renderer must be upgraded together (protocol version gate rejects mismatches)
- All existing specs updated to reflect the new protocol
- Worker tests: 65 passed ✅
- Upload race condition tests: 5 passed ✅
- Health endpoint tests: 14 passed ✅
- Request spec: 53 passed ✅
- Renderer HTTP client spec: 62 passed ✅

## Migration Impact

Users upgrading to 18.0 must upgrade both the gem and node renderer together. The protocol version gate (HTTP 412) ensures mismatched versions fail fast with a clear error message rather than silently breaking.

Fixes #4873

## Summary by CodeRabbit

- **Breaking Changes**
  - Updated the Pro protocol to version 3.0.0.
  - Asset uploads and availability checks now identify bundles using bundle-specific fields instead of the previous bundle list.
  - Render and incremental-render URLs no longer require a request digest.
  - Ruby and Node renderer upgrades must be coordinated to maintain compatibility.

- **Performance**
  - Request digest computation is now limited to renders using prerender caching.
  - Non-cached renders avoid unnecessary digest calculation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,19 @@ After a release, run `/update-changelog` in Claude Code to analyze commits, writ
 
 ### [Unreleased]
 
+#### Changed
+
+- **[Pro] BREAKING: Protocol version bumped to 3.0.0 — removes two wire-protocol legacies**:
+  The `targetBundles` field is no longer sent on `/upload-assets` requests; the node renderer now derives bundle
+  directories from `bundle_<hash>` form keys on that endpoint (the pattern already used by `/render`).
+  `/asset-exists` still uses the `targetBundles` array — it sends bundle hashes, not files, so an explicit array
+  is the right JSON idiom. The `renderRequestDigest` URL segment has been removed from render and
+  incremental-render routes, and the MD5 digest is now computed only when prerender caching is enabled (previously
+  computed on every render). The gem and node renderer must be upgraded together.
+  [Issue 4873](https://github.com/shakacode/react_on_rails/issues/4873).
+  [PR 4875](https://github.com/shakacode/react_on_rails/pull/4875) by
+  [AbanoubGhadban](https://github.com/AbanoubGhadban).
+
 #### Fixed
 
 - **[Pro]** **RSC render-error details are no longer sent to the browser on the fetched payload path**:
--- a/benchmarks/bench-node-renderer.rb
+++ b/benchmarks/bench-node-renderer.rb
@@ -111,7 +111,7 @@ end
 # Returns true/false/nil (nil means couldn't determine)
 # rubocop:disable Style/ReturnNilInPredicateMethodDefinition
 def rsc_bundle?(bundle_timestamp)
-  url = render_url(bundle_timestamp, "rsc_check")
+  url = render_url(bundle_timestamp)
   body = render_body("ReactOnRails.isRSCBundle")
 
   # Use curl with h2c since Net::HTTP doesn't support HTTP/2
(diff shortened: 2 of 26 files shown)


AbanoubGhadban:
+ci-run-hosted

AbanoubGhadban:
+ci-run-hosted

AbanoubGhadban:
+ci-run-hosted

AbanoubGhadban:
+ci-run-hosted

Issue #4903, opened 2026-08-17
Flaky test: HelloWorldRehydratable onChange — missing XHR-completion wait


AbanoubGhadban:
## Summary

The Selenium test `HelloWorldRehydratable onChange should trigger` (`react_on_rails_pro/spec/dummy/spec/system/integration_spec.rb:309`) is flaky due to a missing synchronization point between an async XHR response and Capybara's DOM interaction.

## Root cause

The test clicks a `remote: true` form which submits an async XHR to `/xhr_refresh.js.erb`. The XHR response replaces `container.innerHTML` with fresh SSR'd HTML and dispatches a `'hydrate'` event. However, the test immediately enters `within("#HelloWorldRehydratable-react-component-1") { find("input").set ... }` **without waiting for the XHR to complete**. Capybara grabs a DOM reference from the initial page load; when the XHR response arrives and does `innerHTML =`, that reference becomes stale:

```
Selenium::WebDriver::Error::UnknownError:
  unknown error: unhandled inspector error:
  {"code":-32000,"message":"Node with given id does not belong to the document"}
```

### Why it became more visible on `ppr-integration`

PR #4889 changed the Pro dummy app's webpack React alias from the workspace root (`react-dom ~19.0.4`) to the dummy app's own `node_modules` (`react-dom ~19.2.7`). React 19.2's slightly different microtask scheduling widens the timing window just enough for CI runners to hit the race. The underlying bug (no XHR-completion wait) has always existed.

## Fix (verified by Codex review)

Two files, already proven on the `4891-ppr-v1-postponedstate-versioning` branch (commit `ebfbc3ead`):

**1. `react_on_rails_pro/spec/dummy/app/views/pages/xhr_refresh.js.erb`** — add a `data-refreshed` marker after `innerHTML` + hydration dispatch:

```js
container.setAttribute('data-refreshed', 'true');
```

**2. `react_on_rails_pro/spec/dummy/spec/system/integration_spec.rb:309`** — wait for the marker before interacting:

```ruby
expect(page).to have_css("#component-container[data-refreshed]", wait: 5)
```

This is an unambiguous post-XHR/post-hydration signal. After `data-refreshed` is set, `innerHTML` is done, `forceClientHydration()` has called `ReactOnRails.render()`, and the new `<input>` is stable.

## Additional note

[...]

Issue #232, opened 2016-01-26
 undefined method `react' for #<Rails::Application::Configuration


gemini-google-group:
Hi there.

I've just added the gem to my rails project and bundled. Although, I'm getting this error when trying to run the install generator. (Using ruby 2.20 and Rails 4.2.2)

/home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/railtie/configuration.rb:95:in `method_missing': undefined method `react' for #<Rails::Application::Configuration:0x0000000d7eb7f8> (NoMethodError)
	from /home/boromir/workspace/gemini/config/application.rb:87:in `<class:Application>'
	from /home/boromir/workspace/gemini/config/application.rb:21:in `<module:Gemini>'
	from /home/boromir/workspace/gemini/config/application.rb:20:in `<top (required)>'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands/commands_tasks.rb:141:in `require'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands/commands_tasks.rb:141:in `require_application_and_environment!'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands/commands_tasks.rb:128:in `generate_or_destroy'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands/commands_tasks.rb:50:in `generate'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
	from /home/boromir/.rvm/gems/ruby-2.2.0@gemini/gems/railties-4.2.2/lib/rails/commands.rb:17:in `<top (required)>'
	from script/rails:6:in `require'
	from script/rails:6:in `<main>'

justin808:
@gemini-google-group Did you maybe forget to uninstall the react-rails gem?

We don't have any method called `react`.

obromios:
When I received this message, it was because I had removed react-rails from my Gemfile but not the configuration code in config/development.rb

akashkinwad-define:
When I received this message, it was because I had removed react-rails from my Gemfile but not the configuration code in config/test.rb

PR #4905, opened 2026-08-17
PPR integration: port the HelloWorldRehydratable UJS/XHR rehydration fix (#4904)

AbanoubGhadban:
Port of #4904 to `ppr-integration` (references #4903 — the issue closes via the `main` PR).

## Why

`ppr-integration` currently carries a **latent deterministic failure**: PR #4895 merged commit `ebfbc3ead`'s marker-wait (`expect(page).to have_css("#component-container[data-refreshed]", wait: 5)`), but the marker is set by `xhr_refresh.js.erb` — code that never executes because nothing on the page handles `remote: true` (jquery-ujs was removed in 2021 by PR #232, so the form does a native full-page navigation). The next `rspec-dummy-app-node-renderer` run against this branch fails that spec 100% (CI already showed it on the pre-merge head: run 32003954087). See #4904 for the full root-cause evidence.

## What this PR does

Ports #4904's two commits (`bb2678e8b`, `33b8b4e23`, cherry-picked with `-x`), which make the existing marker-wait valid and fix the rest of the race:

1. Restore UJS (`@rails/ujs` pinned `7.1.3-4`, guarded `Rails.start()`) so the form XHRs and `xhr_refresh.js.erb` executes again.
2. `csp_meta_tag` in the dummy layout — rails-ujs needs the `csp-nonce` meta to execute SJR responses under the strict per-request-nonce CSP.
3. Pre-click hydration gate + ordered `data-refreshed` → `data-hydrated` waits in the spec (the `data-refreshed` wait merged via #4895 is kept, now reachable).
4. Post-hydration commit marker in `HelloWorldRehydratable#componentDidMount`.
5. `forceClientHydration` via `ReactOnRails.reactOnRailsComponentLoaded` (unmounts the innerHTML-orphaned root — fixes the listener leak).

**Reconciliation with this branch** (verified byte-identical to #4904's final files where applicable): `xhr_refresh.js.erb` has no net delta (ebfbc3ead's marker lines already equal the final content); the spec conflict resolved to #4904's exact example body; `client-bundle.js` keeps this branch's `turbo: enableTurbolinks` line; the lockfile was regenerated (pnpm 10.33.4), not cherry-picked — delta is exactly the `@rails/ujs` entries.

No CHANGELOG entry: test-dummy-only change set (and per #4889, ppr-integration changes log once at the final merge to main).

<details>
<summary>Agent details</summary>

### QA evidence (all against final HEAD, local, this branch)

- Target spec: `1 example, 0 failures` (7.26s); fresh `log/test.log` shows `Processing by PagesController#xhr_refresh as JS`.
- Flake harness (mirrors the fixed spec body): idle `100 examples, 0 failures` (67s); loaded under 16 CPU burners `100 examples, 0 failures`, `stale_node_hits=0 stale_elem=0` (212s).
- Full dummy system suite: `72 examples, 0 failures, 2 pending` (both pending are pre-existing intentional skips) — better than the #4895 baseline, whose only failure was this spec.
- CSP diagnostic: `csp-nonce meta present: true`, `0` SEVERE console entries.
- Lint: eslint, prettier, rubocop clean; pro-license-headers 886/886. No RuboCop rule disabled.
- UJS side-effect scan on this branch's views: only hit is the fix target itself; PPR pages green in the full suite.

### Decision log

- Cherry-picked with `-x` from #4904 to preserve provenance; commits retain original messages.
- `pnpm-lock.yaml` excluded from the cherry-pick and regenerated with the repo-pinned pnpm to avoid cross-branch lock drift; resulting delta is the `@rails/ujs` entries only.
- The single conflict (`integration_spec.rb`) resolved by taking #4904's block wholesale; file verified byte-identical to #4904's final version afterward.

### Confidence note

High. Same statistical validation as #4904 (0/200 combined idle+loaded failures on this branch), plus proof the previously-unreachable `data-refreshed` wait now passes against the real XHR path.

</details>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -650,6 +650,9 @@ importers:
       '@loadable/webpack-plugin':
         specifier: ^5.15.2
         version: 5.15.2(webpack@5.104.1)
+      '@rails/ujs':
+        specifier: 7.1.3-4
+        version: 7.1.3-4
       '@rspack/cli':
         specifier: ~2.0.0
         version: 2.0.5(@rspack/core@2.0.5)(@rspack/dev-server@2.1.0(@rspack/core@2.0.5)(selfsigned@5.5.0))
@@ -3135,6 +3138,9 @@ packages:
 
   '@rails/actioncable@8.1.200':
     resolution: {integrity: sha512-on0DSb7AFUkq1ocxivDNQhhGW/RQpY91zvRVyyaEWP4gOOZWy33P/UyxjQk74IENWNrTqs8+zOGHwTjiiFruRw==}
+
+  '@rails/ujs@7.1.3-4':
+    resolution: {integrity: sha512-z0ckI5jrAJfImcObjMT1RBz2IxH6I5q6ZTMFex6AfxSQKZuuL8JxAXvg2CvBuodGCxKvybFVolDyMHXlBLeYAA==}
 
   '@redis/bloom@5.10.0':
     resolution: {integrity: sha512-doIF37ob+l47n0rkpRNgU8n4iacBlKM9xLiP1LtTZTvz8TloJB8qx/MgvhMhKdYG+CvCY2aPBnN2706izFn/4A==}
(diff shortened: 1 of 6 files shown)


AbanoubGhadban:
+ci-run-hosted

PR #5047, opened 2026-09-10
Fix renderer-URL credential redaction leaks

### LINKED ISSUE #5046: [Core] Renderer-URL credential redaction still leaks query-string credentials, file:// userinfo, and slash-bearing passwords
Renderer-URL credential redaction in `react_on_rails/lib/react_on_rails/server_rendering_pool/ruby_embedded_java_script.rb` still passes credentials through in four shapes. These strings reach `ServerBundleLoadError` messages, so they land in application logs and error trackers.

Found by fuzzing `sanitized_renderer_url` / `strip_userinfo` / `strip_malformed_configured_url_userinfo` at `e560acdd`, immediately after #5017 closed the embedded-`@` case.

## Evidence

Each row executed against the helpers as they stand on `main`; `s3cr3t` is the secret.

| input | redacted? | output |
| --- | --- | --- |
| `http://u:s3cr3t@host:3800/b.js` | ok | `http://host:3800/b.js` |
| `http://u:s3cr3t@more@host:3800/b.js` | ok | `http://host:3800/b.js` (the #5017 fix) |
| `https://cdn/bundle.js?X-Amz-Credential=AKIAs3cr3t&X-Amz-Signature=…` | **LEAK** | returned verbatim |
| `https://cdn/bundle.js?token=s3cr3t` | **LEAK** | returned verbatim |
| `file://u:s3cr3t@host/b.js` | **LEAK** | returned verbatim |
| `http://u:pa/s3cr3t@host/b.js` | **LEAK** | returned verbatim |

Mechanics behind each leak:

- **Query string** — `sanitized_renderer_url` only clears `uri.user` / `uri.password`. Nothing inspects the query, so a presigned or tokenised bundle URL is emitted intact. This is the most realistic shape: presigned S3/CloudFront URLs for a server bundle are an ordinary deployment pattern.
- **`file://`** — `URI.parse("file://u:p@host/x")` returns a `URI::File`, whose `userinfo` is always `nil`. The `return url if uri.userinfo.nil?` guard at `:403` therefore short-circuits and returns the raw string. `file_url_to_string` is a live path that interpolates `sanitized_renderer_url(url)` at `:485`.
- **Password containing `/`** — both fallback patterns bound the userinfo run with `[^/…]`, so they stop at the slash.

`strip_userinfo`'s `[^/?#]*@` also cannot span a `?` or `#` in a password; today both call sites pass `configured_url`, which masks it, but the keyword defaults to `nil` so a future third caller reintroduces the leak silently.

## Tension to resolve

#5017 added a spec asserting that query and fragment `@` characters are **preserved** in the diagnostic (`"http://localhost:3035?source=@config#next=@fragment"` must appear in the error message). Closing the query-string leak therefore conflicts with committed, intentional behaviour. That conflict has to be settled deliberately rather than patched around.

## Scope

Choosing the redaction policy is part of this issue; please do not treat a particular approach as given.

- [ ] Decide what the invariant actually is for a configured renderer URL in diagnostics, and reconcile it with the #5017 spec above.
- [ ] Close the four leaks consistently across `sanitized_renderer_url`, `strip_userinfo`, and `strip_malformed_configured_url_userinfo`.
- [ ] Decide whether `strip_userinfo`'s `configured_url:` may keep defaulting to `nil`.
- [ ] Land the fuzz table as a spec so the next bypass fails CI rather than being found by the next audit.

## Context

Prior rounds in this same area, each closing one shape and surfacing the next: #4823, #4825, #5017. #4990 documents the same one-bypass-per-review-round pattern and argues the mechanism does not generalise on its own.

AbanoubGhadban:
## Summary

Closes #5046. Fixes four credential-redaction leak vectors in renderer-URL diagnostics.

## Problem

Credential redaction in `ruby_embedded_java_script.rb` leaks secrets in four shapes that reach `ServerBundleLoadError` messages, application logs, and error trackers:

| Leak | Example | Mechanism |
|---|---|---|
| Query-string credentials | `?X-Amz-Credential=AKIA...` | `sanitized_renderer_url` only clears `uri.user`/`uri.password`; query is untouched |
| `file://` userinfo | `file://u:p@host/b.js` | `URI::File#userinfo` always returns nil; guard short-circuits |
| Slash in password | `http://u:pa/s3cr3t@host/b.js` | Regex `[^/]*@` stops at `/` |
| `configured_url: nil` default | Future callers omitting the keyword | Nil path skips exact-string replacement |

## Solution

**Shared helper:** `ReactOnRails::Utils.sanitize_url_for_display(url)` — one method, called at every output site:

1. **Strips userinfo** (user:password@) from any scheme
2. **Handles `file://`** — raw-string check before trusting `URI::File#userinfo` (which silently returns nil)
3. **Handles malformed URLs** — `rindex('@')` finds the last `@` regardless of password content (no character-class exclusions)
4. **Redacts all query values** — `key=[REDACTED]` preserves key names for diagnostics
5. **Preserves fragments** verbatim

**Key design principle:** The stored `renderer_url` is never modified. Query values may be required for actual HTTP requests. Only the **display copy** shown in logs, errors, and diagnostics is redacted.

## Output sites updated (9 total)

| Location | Sites | Change |
|---|---|---|
| `ruby_embedded_java_script.rb` | 6 (existing) | `sanitized_renderer_url` delegates to shared helper |
| Pro `request.rb` | 2 (new) | Wrap debug log + error raise |
| Pro `assets.rake` | 1 (new) | Wrap `puts` interpolation |

**Not in scope:** `server_manager.rb` warn sites (5 total) — these are dev tooling used by `bin/dev` and adding a `Utils` dependency there risks breaking startup. Can be addressed in a follow-up.

## Fuzz table (landed as 20 specs)

| Input | Output | Status |
|---|---|---|
| `http://u:s3cr3t@host:3800/b.js` | `http://host:3800/b.js` | ✅ |
| `http://u:s3cr3t@more@host:3800/b.js` | `http://host:3800/b.js` | ✅ |
| `https://cdn/b.js?X-Amz-Credential=AKIA&X-Amz-Signature=abc` | `https://cdn/b.js?X-Amz-Credential=[REDACTED]&X-Amz-Signature=[REDACTED]` | ✅ (was ❌) |
| `https://cdn/b.js?token=s3cr3t` | `https://cdn/b.js?token=[REDACTED]` | ✅ (was ❌) |
| `file://u:s3cr3t@host/b.js` | `file://host/b.js` | ✅ (was ❌) |
| `http://u:pa/s3cr3t@host/b.js` | `http://host/b.js` | ✅ (was ❌) |
| `http://localhost:3035?source=@config#next=@fragment` | `http://localhost:3035?source=[REDACTED]#next=@fragment` | ✅ (query redacted, fragment preserved) |

## #5017 spec reconciliation

The PR #5017 spec that asserted verbatim query/fragment preservation is updated: query values are now redacted (deny-by-default policy), fragments are preserved. The original spec's intent (prove `@` in query/fragment isn't misidentified as userinfo) is satisfied by the new behavior — they're handled by query-value redaction, not userinfo stripping.

## Verification

- 24-case standalone fuzz test: all pass
- 20 new RSpec examples in `utils_spec.rb`
- 2 updated RSpec examples in `ruby_embedded_java_script_spec.rb`
- CI will validate the full suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive credentials and query-string values are now redacted from renderer URLs shown in errors, logs, and console output.
  * URL sanitization more reliably handles malformed URLs, file and FTP URLs, fragments, embedded special characters, and URLs embedded in error messages.
  * URL paths, query keys, fragments, and trailing query delimiters remain readable where appropriate.

* **Tests**
  * Expanded coverage for URL sanitization and renderer error messages across standard and malformed URL formats.

### DIFF
--- a/react_on_rails/lib/react_on_rails/utils.rb
+++ b/react_on_rails/lib/react_on_rails/utils.rb
@@ -484,6 +484,22 @@ module ReactOnRails
       end
     end
 
+    # Sanitizes arbitrary error-message text that may contain URLs with
+    # credentials. Finds URL-like substrings (http:// or https://) in the
+    # text and passes each through sanitize_url_for_display. Trailing prose
+    # punctuation is preserved. Use this for exception messages, log text,
+    # or any string that may embed a credentialed URL in prose.
+    def self.sanitize_error_text(text)
+      return text if text.nil? || text.empty?
+
+      text.to_s
+          .gsub(%r{//[^/?#]*@}, "//")
+          .gsub(%r{https?://[^\s]+}) do |match|
+        trimmed = match.sub(/[).,;:'">\]]+\z/, "")
+        sanitize_url_for_display(trimmed) + match[trimmed.length..]
+      end
+    end
+
     # Strips userinfo from the authority section of a well-formed URL whose
     # scheme's URI class doesn't report userinfo (e.g. URI::File). Only looks
     # for @ in the authority — before the first / ? or # after :// — so @ in
--- a/react_on_rails/lib/react_on_rails/utils.rb
+++ b/react_on_rails/lib/react_on_rails/utils.rb
@@ -463,11 +463,12 @@ module ReactOnRails
       begin
         uri = URI.parse(url)
         if uri.userinfo.nil?
-          # URI::HTTP, URI::HTTPS, and URI::FTP report userinfo reliably.
-          # URI::File (and URI::Generic for unknown schemes) silently discard
+          # URI::HTTP (and its subclass URI::HTTPS) reports userinfo reliably.
+          # URI::File (and URI::Generic for unknown schemes) silently discards
           # it — userinfo is always nil regardless of what the raw string
           # contains. For those classes, check the authority section only.
-          unless uri.is_a?(URI::HTTP) || uri.is_a?(URI::FTP)
+          # React on Rails renderer URLs use http://, https://, or file:// only.
+          unless uri.is_a?(URI::HTTP)
             sanitized = strip_authority_userinfo(url)
             return redact_query_values(sanitized)
           end
--- a/react_on_rails/lib/react_on_rails/utils.rb
+++ b/react_on_rails/lib/react_on_rails/utils.rb
@@ -473,12 +473,12 @@ module ReactOnRails
             return redact_query_values(sanitized)
           end
 
-          return redact_query_values_in_uri(uri)
+          return redact_query_values(uri.to_s)
         end
 
         uri.password = nil
         uri.user = nil
-        redact_query_values_in_uri(uri)
+        redact_query_values(uri.to_s)
       rescue URI::InvalidURIError
         sanitized = strip_malformed_url_userinfo(url)
         redact_query_values(sanitized)
@@ -564,17 +564,7 @@ module ReactOnRails
     end
     private_class_method :strip_malformed_url_userinfo
 
-    # Redacts all query-string values in a parsed URI, keeping keys for diagnostics.
-    # Uses regex-based substitution on the query string to avoid URI.encode_www_form
-    # percent-encoding the [REDACTED] placeholder.
-    # Returns the URI as a string.
-    def self.redact_query_values_in_uri(uri)
-      result = uri.to_s
-      redact_query_values(result)
-    end
-    private_class_method :redact_query_values_in_uri
-
-    # Redacts query-string values in a raw URL string using regex substitution.
+    # Redacts query-string values in a URL string using regex substitution.
     # Used when we don't have a parsed URI (malformed URL path).
     #
     # Per RFC 3986, the fragment starts at the first # and the query is between
--- a/react_on_rails/lib/react_on_rails/utils.rb
+++ b/react_on_rails/lib/react_on_rails/utils.rb
@@ -463,11 +463,11 @@ module ReactOnRails
       begin
         uri = URI.parse(url)
         if uri.userinfo.nil?
-          # URI::File (and potentially other schemes) silently discard userinfo.
-          # Check the raw string for a userinfo-like pattern before trusting the
-          # parser's nil — but only in the authority section (before the first
-          # / ? or # after ://), not in the query or fragment where @ is common.
-          if url.match?(%r{://[^/?#]*@})
+          # URI::HTTP, URI::HTTPS, and URI::FTP report userinfo reliably.
+          # URI::File (and URI::Generic for unknown schemes) silently discard
+          # it — userinfo is always nil regardless of what the raw string
+          # contains. For those classes, fall through to regex-based stripping.
+          unless uri.is_a?(URI::HTTP) || uri.is_a?(URI::FTP)
             sanitized = strip_userinfo_by_regex(url)
             return redact_query_values(sanitized)
           end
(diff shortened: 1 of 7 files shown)


react_on_rails/lib/react_on_rails/utils.rb
```diff
@@ -444,6 +445,121 @@ def self.normalize_to_relative_path(path)
...
+        return match[:scheme] + authority[(last_at + 1)..] + suffix
+      end
+
+      # Password contains / so the @ landed after the first path separator.
+      # e.g. http://u:pa/s3cr3t@host/path — pick the @ with host/path after it.
+      best_at = nil
+      pos = 0
+      while (at_idx = rest.index("@", pos))
+        after_at = rest[(at_idx + 1)..]
+        if after_at.include?("/")
```
ihabadham, line 528: I ran this helper with the same URL used by the spec below:

```ruby
url = "http://bad host/webpack/server@bundle.js?source=@config"
ReactOnRails::Utils.sanitize_url_for_display(url)
```

It returns:

```text
http://config
```

For this input, the spec says the bundle path should remain visible and the query value should be redacted:

```text
http://bad host/webpack/server@bundle.js?source=[REDACTED]
```

The problem is the scan starting here. `URI.parse` has already rejected the URL, so the code searches the remaining string for an `@` that looks like the end of a username/password. In this example, the first `@` is part of the filename and the second is the query value. There are no credentials.

The scan chooses the query’s `@`, removes everything before it, and leaves only `http://config`. That loses the host and bundle path and fails the integration spec added in this PR.

We cannot reliably recover the user’s intended URL structure after parsing has failed. Keep credential removal within the normal authority boundary. When malformed input makes the credentials ambiguous, redact the whole displayed URL instead of guessing through the path and query.

react_on_rails/lib/react_on_rails/utils.rb
```diff
@@ -444,6 +445,121 @@ def self.normalize_to_relative_path(path)
...
+    end
+
+    # Scrubs credentials from arbitrary error-message text that may contain URLs.
+    def self.sanitize_error_text(text)
+      return text if text.nil? || text.empty?
+
+      text.to_s
+          .gsub(%r{//[^/?#]*@}, "//")
+          .gsub(%r{https?://[^\s]+}) do |match|
+        trimmed = match.sub(/[).,;:'">\]]+\z/, "")
```
ihabadham, line 481: I reproduced a problem when `sanitize_error_text` receives text that has already been sanitized:

```text
first call:  token=[REDACTED]
second call: token=[REDACTED]]
third call:  token=[REDACTED]]]
```

The regex on this line treats `]` as punctuation at the end of the matched URL. On the second call, it removes the closing bracket from `[REDACTED]`, passes the incomplete value through `sanitize_url_for_display`, and then appends the removed bracket. That adds one extra `]` on every pass.

This happens in the real bundle-loading flow. `file_url_to_string` sanitizes its error, then `read_bundle_js_code` catches that error and sanitizes the nested message again.

Please make this helper idempotent: once text has been sanitized, passing it through again should leave it unchanged. The regression test can express that directly:

```ruby
once = described_class.sanitize_error_text(text)
twice = described_class.sanitize_error_text(once)

expect(twice).to eq(once)
```

react_on_rails_pro/lib/react_on_rails_pro/request.rb
```diff
@@ -691,10 +692,10 @@ def create_connection
           renderer_http_pool_warn_timeout = #{ReactOnRailsPro.configuration.renderer_http_pool_warn_timeout}
           renderer_http_keep_alive_timeout = #{ReactOnRailsPro.configuration.renderer_http_keep_alive_timeout}
           renderer_http_force_http2 = #{ReactOnRailsPro.configuration.renderer_http_force_http2}
-          renderer_url = #{url}
+          renderer_url = #{ReactOnRails::Utils.sanitize_url_for_display(url)}
```
ihabadham, line 695: I don’t see a test that passes a credential-bearing renderer URL through `create_connection`.

The existing connection tests use:

```text
http://node-renderer.com:3800
```

That URL has nothing to redact, so those tests pass whether this code uses the raw URL or the sanitized one. They do not protect either security change made here.

Add a case using something like:

```text
https://user:secret@node-renderer.com:3800?token=abc
```

The test should prove both sides of the behavior:

1. The debug log and connection error do not contain `user`, `secret`, or `abc`.
2. `RendererHttpClient.new` still receives the complete original URL.

That second check matters because the credentials may be required for the real connection. Only the displayed copy should be redacted.

ihabadham (approved):
I don’t have a problem with the policy here. You found some missing edge cases and fixed them. What still makes me uncomfortable is how much we’re guessing about what these URLs might look like. Redacting every query value feels defensive, but I couldn’t find a better general rule. I think we should define the valid and invalid URLs React on Rails Pro can realistically receive and focus on those instead of trying to handle every malformed string.

AbanoubGhadban:
+ci-run-hosted

PR #5065, opened 2026-09-12
Fix authenticityHeaders() mutating its input object

### LINKED ISSUE #5028: [Core] authenticityHeaders() mutates its argument, baking a stale CSRF token into a shared headers object
## Summary

`authenticityHeaders()` writes into the object it is given and returns that same object, rather than returning a copy. Callers that pass a shared or module-level headers object get a CSRF token permanently baked into it, captured at the time of the **first** call. After a Turbo navigation rotates the `csrf-token` meta tag, later requests reusing that object still send the original token — producing intermittent `422 InvalidAuthenticityToken` responses that do not reproduce on a full page reload.

Trivial fix; the value is mostly in how hard the symptom is to diagnose.

## Code

`packages/react-on-rails/src/Authenticity.ts:11-15`
```ts
export const authenticityHeaders = (otherHeaders: Record<string, string> = {}): AuthenticityHeaders =>
  Object.assign(otherHeaders, {
    'X-CSRF-Token': authenticityToken(),
    'X-Requested-With': 'XMLHttpRequest',
  });
```

`Object.assign(otherHeaders, …)` mutates the first argument in place. Exposed publicly as `ReactOnRails.authenticityHeaders` via `packages/react-on-rails/src/capabilities/core.ts:69-70`.

## Concrete failure

```js
// module scope
const BASE_HEADERS = { 'Content-Type': 'application/json' };

// first mutating request, on page load
await fetch('/posts', { method: 'POST', headers: ReactOnRails.authenticityHeaders(BASE_HEADERS), body });
// BASE_HEADERS is now:
// { 'Content-Type': 'application/json', 'X-CSRF-Token': '<token A>', 'X-Requested-With': 'XMLHttpRequest' }

// ... Turbo Drive visit; Rails serves a new csrf-token meta tag (token B) ...

// second request re-reads BASE_HEADERS, which still carries token A
await fetch('/posts', { method: 'POST', headers: ReactOnRails.authenticityHeaders(BASE_HEADERS), body });
// -> 422 InvalidAuthenticityToken
```

The re-call does overwrite `X-CSRF-Token` with the current token, so this only bites when the mutated object is read somewhere the helper is not re-invoked — a `fetch` wrapper that captures `BASE_HEADERS` once, a spread into another literal, an axios `defaults.headers` assignment, or any request built from the object directly. The general problem is simply that a getter-shaped helper has a side effect on its input.

## The documented usage encourages it

`docs/oss/api-reference/javascript-api.md:14` passes a named variable rather than a fresh literal:

```js
// compose Rails specific request header as following { X-CSRF-Token: csrfToken, X-Requested-With: XMLHttpRequest }
header = ReactOnRails.authenticityHeaders(otherHeader);
```

and the API summary at `:162-167` documents it as `@returns {*} header` with no mention that `otherHeaders` is modified. (`docs/oss/building-features/debugging-hydration-mismatches.md:194` passes an inline literal, which is safe.)

## Test gap

`packages/react-on-rails/tests/Authenticity.test.js:26-29` — the only behavioural test calls the helper with **no argument**, so the mutation path is never exercised:

```js
it('returns valid header with CSRF token', () => {
  const realHeader = ReactOnRails.authenticityHeaders();
  expect(realHeader).toEqual({ 'X-CSRF-Token': testToken, 'X-Requested-With': 'XMLHttpRequest' });
});
```

## Fix

```ts
export const authenticityHeaders = (otherHeaders: Record<string, string> = {}): AuthenticityHeaders =>
  Object.assign({}, otherHeaders, {
    'X-CSRF-Token': authenticityToken(),
    'X-Requested-With': 'XMLHttpRequest',
  });
```

Regression test: pass an object, assert the returned value is not reference-equal to the input and that the input is unchanged.

## Rider in the same five lines (worth deciding, separable)

`AuthenticityHeaders` is typed `'X-CSRF-Token': string | null` (`packages/react-on-rails/src/types/index.ts:151-154`), and `authenticityToken()` returns `null` when no `meta[name="csrf-token"]` is present. So an app missing `csrf_meta_tags` gets a literal `null` header value silently sent on every mutating request, with no warning. Both `railsAction.ts:449` and `useRailsForm.ts:489` handle the null case explicitly — `useRailsForm` throws a helpful error naming the missing meta tag — so this helper is the odd one out. Happy to split this into its own issue if the desired behaviour (warn? throw? omit the header?) needs its own discussion.

## What I did not verify

Read-verified on `main` (`17.0.0.rc.6`). I confirmed the mutation semantics and the test gap by reading; I did not build an app to observe the 422.

AbanoubGhadban:
## Summary

`authenticityHeaders()` used `Object.assign(otherHeaders, ...)` which mutated the caller's input object in place. Callers passing a shared or module-level headers object would get a stale CSRF token baked into it, causing intermittent `422 InvalidAuthenticityToken` errors after Turbo navigations.

## Changes

- **`packages/react-on-rails/src/Authenticity.ts`**: Change `Object.assign(otherHeaders, { ... })` to `Object.assign({}, otherHeaders, { ... })` so the input is never modified.
- **`packages/react-on-rails/tests/Authenticity.test.js`**: Add three regression tests:
  - Input object is not mutated
  - Return value is a new object (not the same reference)
  - Merged output contains both input headers and CSRF headers
- **`docs/oss/api-reference/javascript-api.md`**: Update JSDoc and usage comment to clarify non-mutating behavior.
- **`CHANGELOG.md`**: Add entry under Unreleased > Fixed.

## Scope exclusion

The `null` token rider mentioned in the issue (silently sending `null` when `csrf_meta_tags` is missing) is excluded from this PR — it's a separate design decision.

## Validation

- `npx jest tests/Authenticity.test.js` — 7/7 pass (3 new regression tests)
- Full jest suite — 381/381 pass (27 suites)
- `pnpm run build` — clean TypeScript compilation

Fixes #5028

<details>
<summary>Agent details</summary>

### Confidence note
- Validated: jest tests (unit), TypeScript build
- Remaining UNKNOWN: hosted CI (not yet run)
- Residual risk: None — one-line behavioral fix with deterministic test coverage, no runtime or security boundary touched

### Decision log
- Excluded the null-token rider from scope per issue author's suggestion to split it into its own issue
</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * `authenticityHeaders` now returns a new header object without modifying the caller’s input, preventing stale CSRF tokens from persisting across requests.
  * Provided headers are merged with the required CSRF and XMLHttpRequest headers.

* **Documentation**
  * Clarified header merging behavior, input immutability, and parameter and return types.

* **Tests**
  * Added coverage for input immutability, new object creation, and header merging.


AbanoubGhadban:
+ci-run-hosted

PR #5085, opened 2026-09-18
Allow stable release retries after metadata-only commits

justin808:
## Why

`v17.1.0` already correctly identifies the immutable version-bump commit. PR #5084 then added only release-process metadata on `release/17.1.0`; the release retry refuses to continue because it requires the stable tag to equal the later branch head.

## What

- Allow a stable release retry only when the existing tag is an ancestor and every intervening commit is classified as non-runtime.
- Preserve the tag's original immutable candidate SHA through local and remote publication-boundary checks.
- Keep prerelease tags and runtime changes on the existing exact-SHA path.

## Verification

- `bundle exec rubocop rakelib/release.rake react_on_rails/spec/react_on_rails/release_rake_helpers_spec.rb`
- `(cd react_on_rails && bundle exec rspec spec/react_on_rails/release_rake_helpers_spec.rb:23345)`
- `pnpm install --frozen-lockfile`
- `git diff --check`

The normal release preview intentionally refuses this feature branch; its branch guard permits only `main` or the matching protected release branch.


justin808:
+ci-run-hosted

justin808:
Addressed: retry ancestry now accepts only an explicit release-operational path set in addition to the existing non-runtime classification, and the regression test evaluates the actual `v17.1.0` → `dd646bf` history without stubbing the classifier. Runtime paths remain rejected.

